PR #24, opened 2020-01-17
Jsk fix server construction

jskress:
### Description

This change updates server URL handling to treat a URI of the form, `../uri` as if they are `../uri/`.  This provides the behavior most users would intuitively expect.

In addition, error handling on Connect API calls now properly deal with receiving non-JSON response payloads (as will happen under certain error conditions.

- [ ] Server URLs that don't end in `/` will now work.
- [ ] If we get a non-JSON error back from Connect, it will now be correctly reported rather than throw an exception.

### DIFF
--- a/rsconnect/api.py
+++ b/rsconnect/api.py
@@ -4,6 +4,7 @@ import re
 import socket
 import time
 import ssl
+from json import JSONDecodeError
 
 from os.path import join, dirname
 from six.moves import http_client as http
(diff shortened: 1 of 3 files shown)


rsconnect/api.py
```diff
@@ -4,6 +4,7 @@
 import socket
 import time
 import ssl
+from json import JSONDecodeError
```
scottmmjackson: Whoops! This isn't working.

PR #39, opened 2020-02-06
Allow for empty body tagged as JSON.

jskress:
### Description

This change updates the HTTP code to allow for a response that carries both a content type of JSON and an empty body.

- [ ] The `deploy` command should now work.

### DIFF
--- a/rsconnect/http_support.py
+++ b/rsconnect/http_support.py
@@ -99,7 +99,7 @@ class HTTPResponse(object):
             self.content_type = response.getheader('Content-Type')
             self.response_body = response.read().decode('utf-8')
 
-            if self.content_type and self.content_type.startswith('application/json'):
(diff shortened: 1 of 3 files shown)


rsconnect/http_support.py
```diff
@@ -99,7 +99,7 @@ def __init__(self, response=None, exception=None):
             self.content_type = response.getheader('Content-Type')
             self.response_body = response.read().decode('utf-8')
 
-            if self.content_type and self.content_type.startswith('application/json'):
+            if self.content_type and self.content_type.startswith('application/json') and len(self.response_body) > 0:
                 self.json_data = json.loads(self.response_body)
```
scottmmjackson: We might also want to add more context to the `loads` raise.

PR #41, opened 2020-02-06
Always use unicode certificate data

scottmmjackson:
https://bugs.python.org/issue37079

We were failing to use `cacert` properly in python 2.7 because `.read()` doesn't return unicode. Now it does.

This addresses the "nested asn1 error" message when attempting to contact a server with the `cacert` specified on the command line in python 2.7.

## Testing Notes

Deploy to a server which requires a certificate under python 2.7

Alternatively, add a server which requires a certificate under python 2.7

### DIFF
--- a/rsconnect/main.py
+++ b/rsconnect/main.py
@@ -75,7 +75,7 @@ def test(server, api_key, insecure, cacert, verbose):
 
 
 def _test_server_and_api(server, api_key, insecure, ca_cert):
-    ca_data = ca_cert and ca_cert.read()
+    ca_data = ca_cert and unicode(ca_cert.read())
     me = None
 
     with cli_feedback('Checking %s' % server):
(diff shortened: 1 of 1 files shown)


rsconnect/main.py
```diff
@@ -75,7 +75,7 @@ def test(server, api_key, insecure, cacert, verbose):
 
 
 def _test_server_and_api(server, api_key, insecure, ca_cert):
-    ca_data = ca_cert and ca_cert.read()
+    ca_data = ca_cert and unicode(ca_cert.read())
```
jskress: This doesn't work under 3.x. Is there a `six` function that can help here?

jskress: Maybe `six.u(...)`?

scottmmjackson: done

PR #62, opened 2020-02-20
Detect no conda and refuse to deploy conda content when detected

scottmmjackson:
- Checks the server information before deployment of a notebook or manifest and bails if there's a mismatch between what the user or manifest asks for and the server's capabilities.

### DIFF
--- a/rsconnect/main.py
+++ b/rsconnect/main.py
@@ -340,8 +340,7 @@ def deploy_notebook(name, server, api_key, static, new, app_id, title, python, c
 
     if conda:
         with cli_feedback('Ensuring conda is supported'):
-            if not is_conda_supported_on_server(connect_server.url):
-                conda = False
+            if not is_conda_supported_on_server(connect_server):
                 raise api.RSConnectException('Conda is not supported on the target server. Please try deploying '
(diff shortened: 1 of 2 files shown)


rsconnect/main.py
```diff
@@ -337,6 +338,13 @@ def deploy_notebook(name, server, api_key, static, new, app_id, title, python, c
 
     click.secho('    Deploying %s to server "%s"' % (file, connect_server.url), fg='white')
 
+    if conda:
+        with cli_feedback('Ensuring conda is supported'):
+            if not is_conda_supported_on_server(connect_server.url):
```
jskress:
```suggestion
            if not is_conda_supported_on_server(connect_server):
```
Needs to be the whole server info, not just the URL.

rsconnect/main.py
```diff
@@ -337,6 +338,13 @@ def deploy_notebook(name, server, api_key, static, new, app_id, title, python, c
...
+                conda = False
```
jskress: Strictly speaking, this isn't necessary since the following `raise` will end the program.  You don't do this below so maybe remove this line for consistency?

PR #64, opened 2020-02-24
Add support for deploying Python APIs.

jskress:
### Description

This change adds support for deploying Python APIs to Connect.  Some refactoring was done for code reuse and cleanup.  This includes final conversion to an enum style for app modes.  The previous check for Conda support on Connect has also been generalized to a simple mechanism for verifying Connect server capabilities (the new `deploy api` does this to make sure Connect is new enough to support it).

As with the other deploy actions, the `deploy api` ability has been added both to the CLI and as a callable function, including functions for the sub-pieces of that work.

Connected to https://github.com/rstudio/connect/issues/16407

- [ ] No existing functionality should be broken.
- [ ].Running `rsconnect deploy api --help` should produce clear and complete help.
- [ ] `rsconnect deploy api _<server-info>_  _<dir>_` should successfully deploy the Python API contained in the specified directory to Connect.
- [ ] `rsconnect deploy api ...` with extra files should correctly include those files in the generated bundle.
- [ ] `rsconnect deploy api -x ...` with a glob pattern relative to the deploy dir should properly exclude matching files from being included in the bundle.
- [ ] The `--exclude`/`-x` option should be allowed to repeat to specify more than one glob.
- [ ] Explicitly specifying a file as an extra file that is also matched by a `-x` exclusion glob should result in the file being included in the generated bundle.


rsconnect/main.py
```diff
@@ -340,31 +380,15 @@ def deploy_notebook(name, server, api_key, static, new, app_id, title, python, c
 
     if conda:
         with cli_feedback('Ensuring conda is supported'):
-            if not is_conda_supported_on_server(connect_server):
-                raise api.RSConnectException('Conda is not supported on the target server. Please try deploying '
-                                             'again without conda enabled (remove the --conda/-C flag).')
+            check_server_capabilities(connect_server, [is_version_1_8_2_or_higher, is_conda_supported_on_server])
```
scottmmjackson: We shouldn't fail if the system is masking their version number

rsconnect/main.py
```diff
@@ -408,28 +431,118 @@ def deploy_manifest(name, server, api_key, new, app_id, title, insecure, cacert,
...
+        app_id, deployment_name, title, app_mode = \
+            gather_basic_deployment_info_for_api(connect_server, app_store, directory, new, app_id, title)
+
+    click.secho('    Deploying %s to server "%s"' % (directory, connect_server.url), fg='white')
+
+    with cli_feedback('Checking server capabilities'):
+        check = [is_version_1_8_2_or_higher]
+        if conda:
+            check.append(is_conda_supported_on_server)
+        check_server_capabilities(connect_server, check)
```
scottmmjackson: Same. It would also help if we had a server flag advertising the feature rather than checking the version.

PR #66, opened 2020-02-25
Add write-manifest api support

jskress:
### Description

This change adds the `write-manifest api` command to the `rsconnect-python` CLI and API.  A bug where deploying an API created a store file named `.json` has been fixed; the file is now properly named.  Also, error handling around interacting with Connect is now more consistent.

Connected to https://github.com/rstudio/connect/issues/16408

- [ ] Running `rsconnect write-manifest api --help` should produce clear and complete prose.
- [ ] Running `rsconnect write-manifest api` should work correctly.

### DIFF
--- a/rsconnect/actions.py
+++ b/rsconnect/actions.py
@@ -516,6 +516,9 @@ def gather_basic_deployment_info_for_api(connect_server, app_store, directory, n
         # Use the saved app information unless overridden by the user.
         app_id, title, app_mode = app_store.resolve(connect_server.url, app_id, title, app_mode)
 
+    if directory[-1] == '/':
+        directory = directory[:-1]
+
     return app_id, deployment_name, title or default_title(directory), app_mode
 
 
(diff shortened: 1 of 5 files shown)


rsconnect/actions.py
```diff
@@ -377,7 +378,8 @@ def deploy_python_api(connect_server, directory, extra_files, excludes, entry_po
     :return: the ultimate URL where the deployed app may be accessed and the sequence
     of log lines.  The log lines value will be None if a log callback was provided.
     """
-    app_store = AppStore(directory)
+    module_file = join(directory, entry_point.split[':'][0] + '.py')
```
scottmmjackson: Should we be degrading more gracefully if `entry_point` has no `:`? Or are we assuming this is always called after `_validate_entry_point`?

jskress: I thought about this for a while.  For now, I've landed on the side of the latter.  I do think though that we should do a follow-on change where all the input validation moves to the `actions` layer out of `main`.

PR #70, opened 2020-02-28
Fix force generating environment file not working in api deploy

scottmmjackson:
- Trust the environment generator as to which environment file we add to the bundle.

### DIFF
--- a/rsconnect/bundle.py
+++ b/rsconnect/bundle.py
@@ -366,7 +366,7 @@ def make_api_bundle(directory, entry_point, app_mode, environment, extra_files=N
 
                 if keep_manifest_specified_file(rel_path) and \
                         (rel_path in extra_files or abs_path not in excludes) and \
-                        rel_path is not environment['filename']:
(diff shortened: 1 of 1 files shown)


rsconnect/bundle.py
```diff
@@ -357,13 +357,16 @@ def make_api_bundle(directory, entry_point, app_mode, environment, extra_files=N
...
 
         for subdir, dirs, files in os.walk(directory):
             for file in files:
                 abs_path = os.path.join(subdir, file)
                 rel_path = os.path.relpath(abs_path, directory)
 
-                if keep_manifest_specified_file(rel_path) and (rel_path in extra_files or abs_path not in excludes):
+                if keep_manifest_specified_file(rel_path) and \
+                        (rel_path in extra_files or abs_path not in excludes) and \
+                        rel_path is not environment['filename']:
```
jskress: Doesn't this need to be `!=` rather than `is not` so it does an equality rather than identity check?

PR #73, opened 2020-03-02
Correct dir handling.

jskress:
### Description

This change corrects how `deploy api` does its Python environment checking so that the proper directory is scanned.

Connected to n/a

- [ ] Specifying a directory to `rsconnect deploy api` without a trailing `/` will now do correct `requirements.txt` evaluation.

### DIFF
--- a/rsconnect/actions.py
+++ b/rsconnect/actions.py
@@ -462,7 +462,7 @@ def fake_module_file_from_directory(directory):
     :return: the directory plus the (potentially) fake module file.
     """
     app_name = abspath(directory)
-    app_name = dirname(app_name) if app_name.endswith('/') else basename(app_name)
+    app_name = dirname(app_name) if app_name.endswith(os.path.sep) else basename(app_name)
     return join(directory, app_name + '.py')
 
 
--- a/rsconnect/actions.py
+++ b/rsconnect/actions.py
@@ -392,17 +392,12 @@ def validate_manifest_file(file_or_directory):
     return file_or_directory
 
 
-def validate_entry_point(directory, entry_point):
+def validate_entry_point(entry_point):
     """
     Validates the entry point specified by the user, expanding as necessary.  If the
     user specifies nothing, a module of "app" is assumed.  If the user specifies a
     module only, the object is assumed to be the same name as the module.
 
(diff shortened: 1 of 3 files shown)


rsconnect/actions.py
```diff
@@ -463,6 +453,19 @@ def deploy_jupyter_notebook(connect_server, file_name, extra_files, new=False, a
...
+def fake_module_file_from_directory(directory):
+    """
+    Takes a directory and invents a properly named file that though possibly fake,
+    can be used for other name/title derivation.
+
+    :param directory: the directory to start with.
+    :return: the directory plus the (potentially) fake module file.
+    """
+    app_name = abspath(directory)
+    app_name = dirname(app_name) if app_name.endswith('/') else basename(app_name)
```
Lytol: Instead of relying on consistent use of trailing slashes, what about using `isdir`?

rsconnect/main.py
```diff
@@ -460,7 +460,10 @@ def deploy_api(name, server, api_key, insecure, cacert, entrypoint, exclude, new
         check_server_capabilities(connect_server, checks)
 
     with cli_feedback('Inspecting Python environment'):
-        _, environment = get_python_env_info(directory, python, not conda, force_generate)
+        root = directory
+        if not root.endswith('/'):
```
Lytol: Same as above: using trailing slashes is a little less consistent/definitive than stat'ing whether it's a directory (`isdir`) IMO.  Subjective, but just wanted to offer an alternative... your call.

rsconnect/actions.py
jskress: Because if someone actually specifies the trailing `/`, `isdir()` would still return `True`.

rsconnect/main.py
jskress: Fixed differently.

rsconnect/actions.py
Lytol: Which is correct and what you want, no?  Anyway, possibly I'm missing something, you do you.

rsconnect/main.py
Lytol: All good, you do you.

PR #77, opened 2020-03-05
Refactor so we get API file lists right and some warnings.

jskress:
### Description

This change primarily adds warnings to the terminal when:

- `deploy notebook` or `deploy api` see an existing `manifest.json` file (noting it will be ignored).
- `deploy notebook` or `deploy api` with the `--force-generate` option see an existing requirements file (noting it will be ignored).

It also fixes an incongruity between file lists in the `manifest.json` produced by `deploy api` and `write-manifest api` along with fixing a bug where `deploy api` would incorrectly include an on-disk `manifest.json` file into the bundle.

Connected to https://github.com/rstudio/connect/issues/16793

- [ ] `deploy api` will no longer include an on-disk `manifest.json` at the top-level directory in bundles (which effectively overrode the generated one).
- [ ] The list of files in a bundle generated by `deploy api` will now match the list in the generated `manifest.json` (except for the `manifest.json` itself; it will be in the bundle but not listed in itself).
- [ ] If `deploy notebook` sees a `manifest.json` file that is peer to the notebook file, a warning will be displayed that the file will be ignored.
- [ ] If `deploy notebook --force-generate` sees a requirements file that is peer to the notebook file, a warning will be displayed that the file will be ignored.
- [ ] If `deploy api` sees a `manifest.json` file immediately under the given directory, a warning will be displayed that the file will be ignored.
- [ ] If `deploy notebook --force-generate` sees a requirements file immediately under the given directory, a warning will be displayed that the file will be ignored.

### DIFF
--- a/rsconnect/main.py
+++ b/rsconnect/main.py
@@ -297,7 +297,7 @@ def _warn_on_ignored_requirements(directory, requirements_file_name):
     :param requirements_file_name: the name of the requirements file.
     """
     if exists(join(directory, requirements_file_name)):
-        click.secho('    Warning: the existing $s file will not be used or considered.' % requirements_file_name,
+        click.secho('    Warning: the existing %s file will not be used or considered.' % requirements_file_name,
                     fg='yellow')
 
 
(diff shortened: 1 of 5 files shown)


rsconnect/main.py
```diff
@@ -277,6 +277,30 @@ def _validate_deploy_to_args(name, url, api_key, insecure, ca_cert, api_key_is_r
...
+def _warn_on_ignored_requirements(directory, requirements_file_name):
+    """
+    Checks for the existence of a file called manifest.json in the given directory.
+    If it's there, a warning noting that it will be ignored will be printed.
+
+    :param directory: the directory to check in.
+    :param requirements_file_name: the name of the requirements file.
+    """
+    if exists(join(directory, requirements_file_name)):
+        click.secho('    Warning: the existing $s file will not be used or considered.' % requirements_file_name,
```
scottmmjackson:
```suggestion
        click.secho('    Warning: the existing %s file will not be used or considered.' % requirements_file_name,
```

PR #82, opened 2020-03-06
Add the dash app mode.

jskress:
### Description

This change adds the Dash app app mode.

- [ ] Deploying a manifest using that app mode should now work.


jskress:
@mmarchetti it gets used on the `rsconnect info` command.  I've updated it to match.

Issue #107, opened 2020-03-30
HTTP 400 Bad Request  error during deploy log when publishing


koverholt:
Python version: 3.6.10
rsconnect-python version: 1.4.1.1

I see the following 400 error every time when publishing to Colorado, it seems to always happen while the deploy log is streaming. Note that the deployment continues on RSC and deploys without issues, this error just shows up on the `rsconnect-python` CLI/client side.

```
$ rsconnect deploy dash . -n colorado
Checking arguments...                            [OK]
    Deploying . to server "https://colorado.rstudio.com/rsc/"
    Warning: the existing manifest.json file will not be used or considered.
Checking server capabilities...                  [OK]
Inspecting Python environment...                 [OK]
Creating deployment bundle...                    [OK]
Uploading bundle...                              [OK]
Saving deployment data...                        [OK]

Deployment log:
Building Dash application...
Bundle requested Python version 3.6.10; using /opt/python/3.6.7/bin/python3.6 which has version 3.6.7
2020/03/30 05:47:49.174158531 Running with python "3.6.7 (default, Jul 18 2019, 02:52:05)  [GCC 7.4.0]" at /opt/python/3.6.7/bin/python3.6
2020/03/30 05:47:49.174585071 Running as user: rstudio-connect
2020/03/30 05:47:49.184032698 Creating environment: 72MIoQs7s0k-Jw7BMqwmCQ
2020/03/30 05:47:49.772535226 Already using interpreter /opt/python/3.6.7/bin/python3.6
2020/03/30 05:47:49.772567849 Using base prefix '/opt/python/3.6.7'
2020/03/30 05:47:49.772624580 New python executable in /opt/rstudio-connect/mnt/app/python/env/bin/python3.6
2020/03/30 05:47:49.772636318 Also creating executable in /opt/rstudio-connect/mnt/app/python/env/bin/python
2020/03/30 05:47:49.772645469 Installing setuptools, pip, wheel...
2020/03/30 05:47:58.915781965 done.
2020/03/30 05:48:00.145924075 Collecting dash==1.0.2
2020/03/30 05:48:00.176607955   Downloading dash-1.0.2.tar.gz (56 kB)
2020/03/30 05:48:00.815331950 Collecting gunicorn==19.9.0
2020/03/30 05:48:00.825536957   Using cached gunicorn-19.9.0-py2.py3-none-any.whl (112 kB)
2020/03/30 05:48:01.215812868 Collecting pandas==0.24.2
[...]

PR #108, opened 2020-03-30
#107: Keep cookie header from becoming unnecessarily large.

jskress:
### Description

This change replaces the ever-growing list approach to caching cookie values to a dictionary.  It appears with the right combination of clusters/reverse proxies, `rsconnect` could continually grow the value of the `Cookie` header to larger than can be handled.

Connected to #107 

The key place where this error was appearing was during the printing out of the deploy task log output.  The longer the deploy log output, the more likely the error is to appear (prior to the fix).

- [ ] In a non-clustered environment, everything still works.
- [ ] In a clustered environment everything should still work.
- [ ] The presence of a load balancer should not cause this to appear.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@
 *   Added more helpful feedback when a "requested object does not exist" error is
     returned by Connect.
 
+*   Fixed an issue where cookie header size could grow inappropriately (#107).
+
 
 `rsconnect-python` 1.4.1
 --------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


koverholt:
Yep, works great for me: https://github.com/rstudio/rsconnect-python/issues/107#issuecomment-606246166

Thanks!

kgartland-rstudio:
Performed the following in Dogfood and Colorado using `rsconnect-python 1.4.2`
- [x] deployed notebook
- [x] deployed dash
- [x] deployed flask
- [x] write manifest

PR #115, opened 2020-04-10
Fix logging output for Python 2.x and package versions.

jskress:
### Description

This change corrects an issue where the alignment of output was broken, especially under Python 2.  It also pins the minimum versions we require of the `click` and `six` libraries.

Connected to #n/a

- [x] If a version of, say, `click` is currently installed that is less than version 7, installing `rsconnect-python` will now require version 7+.
- [x] The `[OK]` or `[ERROR]` text that ends a status line should now be back on the corresponding line instead of on the following line.


kgartland-rstudio:
i now see this in the logs when `click < 7.0.0` is installed:
```
pkg_resources.DistributionNotFound: The 'click>=7.0.0' distribution was not found and is required by rsconnect-python
```

alignment for deploy logs in python2 looks better now as well.

PR #118, opened 2020-04-17
Give the user better feedback on some I/O errors

jskress:
### Description

This change catches I/O errors (such as file not found) that occur while building a bundle for deploying from a manifest.  Prior to this change, a traceback was displayed.  Now, the user is given proper guidance about next steps.

Connected to #https://github.com/rstudio/connect/issues/17262

See https://github.com/rstudio/connect/issues/17262#issuecomment-615270088

Now, more appropriate error information and next step guidance will be displayed.

- [ ] Proofread new feedback for completeness and correctness.

### DIFF
--- a/rsconnect/main.py
+++ b/rsconnect/main.py
@@ -1,3 +1,4 @@
+import errno
 import logging
 import textwrap
 from os.path import abspath, dirname, exists, isdir, join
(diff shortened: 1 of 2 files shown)


rsconnect/main.py
```diff
@@ -475,7 +476,15 @@ def deploy_manifest(name, server, api_key, insecure, cacert, new, app_id, title,
         check_server_capabilities(connect_server, [is_conda_supported_on_server])
 
     with cli_feedback('Creating deployment bundle'):
-        bundle = make_manifest_bundle(file)
+        try:
+            bundle = make_manifest_bundle(file)
+        except IOError as error:
+            msg = 'Unable to include the file %s in the bundle: %s' % (error.filename, error.args[1])
+            if error.args[0] == errno.ENOENT:
+                msg = '\n'.join([msg, 'Since the file is missing but referenced in the manifest, you will need to\n'
```
scottmmjackson, line 484:
```suggestion
                msg = '\n'.join([msg, 'Because the file is missing but referenced in the manifest, you will need to\n'
```

scottmmjackson: Optional: some are sticklers about [using because for cause and since for time](https://www.writersdigest.com/online-editor/since-vs-because) but it's not clear that "since" is incorrect so you can take this or leave it.

PR #121, opened 2020-04-27
Revert the usage of the --from-history Conda option.

jskress:
### Description

This change removes the `--from-history` flag from the `conda env export` command we use to freeze the environment.  Having the flag allows for less specificity, especially of Python version, than we want to cope with.

Connected to https://github.com/rstudio/connect/issues/17329

- [ ] The issues brought up in the connected issue should no longer occur.


scottmmjackson:
Merging to unblock testing

PR #123, opened 2020-05-05
We need better HTTP timeout handling.

jskress:
### Description

Our HTTP client support was using an often too-short timeout value.  It would especially fire on the deployment of large bundles.  This change makes the HTTP client timeout values easier to specify and, for bundle deploys, specifies a larger value.

* Deploy bundle calls now have a timeout of 2 minutes.
* All other calls now have a timeout of 30 seconds.

Connected to #n/a

- [ ] Timeouts should now obey the above rules.

### DIFF
--- a/rsconnect/api.py
+++ b/rsconnect/api.py
@@ -51,10 +51,12 @@ class RSConnectServer(object):
 
 
 class RSConnect(HTTPServer):
-    def __init__(self, server, cookies=None):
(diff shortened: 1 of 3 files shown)


jskress:
I'm not sure I want to go there.  We already have two different values for timeouts (deploy and everything else).  Making that both usable and extensible (in case we need to provide alternate timeouts in other situations) would get pretty tricky.

PR #125, opened 2020-05-15
Pipenv and linting and formatting and actions

meatballhat:
### Description

I fully realize there's a lot to take in here. The tl;dr:
- add `pipenv` stuff
- add `black` and `flake8`
- commit all of the `black` reformatting
- commit _some of_ the `flake8` fixes
- implement _most of_ the stuff that Jenkins does in GitHub Actions

I would like to propose managing the remainder of the artifacts publishing via GitHub Releases, which I am also happy to set up.

Criticism very much welcome 😅


.flake8
```diff
@@ -0,0 +1,19 @@
+[flake8]
+max_line_length = 88
```
jskress, line 2: Why this number when PEP8 allows for 120?

conftest.py
```diff
@@ -0,0 +1,6 @@
+import os.path
```
jskress, line 1: You should either change this to just `import os` or change the statements below from `os.path....` to just `path...`

meatballhat:
@jskress Thank you for taking the time to provide this feedback. The latest commit switches to line length 120 👍

conftest.py
meatballhat: I like to import the full module name for clarity 😁

Makefile
```diff
@@ -1,49 +1,88 @@
...
-		-w /rsconnect \
-		rsconnect-python:$* \
-		bash -c
+TEST_COMMAND ?= ./runtests
+SHELL_COMMAND ?= pipenv shell
 
-ifneq (${JOB_NAME},)
-	# Jenkins
-	RUNNER=bash -c
+ifneq ($(JOB_NAME),)
```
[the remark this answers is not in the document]
meatballhat: nope! removing 👍

runtests
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+set -o errexit
+set -o pipefail
+set -o xtrace
+
+: "${PYTEST_ARGS:=-vv --cov=rsconnect --cov-report=term --cov-report=html}"
+
+case "${PYTHON_VERSION}" in
```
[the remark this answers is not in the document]
meatballhat: IIUC these are specifically problems with `black` not being available for python 2.7 and 3.5, which I think would be surfacing whether `black` is in `Pipfile` or `requirements.txt` 🤷

meatballhat:
The line length issue was reasonably straightforward to resolve with `black`, but there is no such config bit for quoting style. I view this mostly as a feature, given experiences I've had with `go fmt` (zero config) and `rubocop` (all the things config) 😅 .

Lytol:
> given experiences I've had with `go fmt` (zero config) and `rubocop` (all the things config)  .

_flinches_ I've seen many holy wars waged over a `rubocop.yml`

mmarchetti:
black supports a `--skip-string-normalization` flag which avoids the quote style issue, but if you choose to normalize, it requires double quotes.

meatballhat:
@mmarchetti yes! IIUC `--skip-string-normalization` is meant for easing migration to `black`, but not intended for long-term use. I didn't mean to be so dismissive.

PR #126, opened 2020-05-19
Exclude known virtualenv dirs from bundle

meatballhat:
### Description

...as described in connected issue.

This change explicitly _does not_ attempt to tackle the concept of an `.rscignore` file. I would like to do that work, but I believe it is a good bit more involved given how many functions currently exist with the expectation that a list of excludes are optionally provided.

Connected to rstudio/connect#17267

- in-repo virtualenv directories of the following names are automatically excluded from the bundle:
    - `.env/`
    - `.venv/`
    - `env/`
    - `venv/`


colearendt:
For anyone (like me) who finds this thread in the future, there is now a "--exclude" flag that supports blob matching

PR #128, opened 2020-05-28
Replace environment dict with Environment type

meatballhat:
### Description

...to help with identifying incomplete environment building bugs, and
code-as-documentation, and types are cool :sunglasses:

Connected to rstudio/connect#17426

- automated tests pass
- normal usage works

### DIFF
--- a/rsconnect/tests/test_environment.py
+++ b/rsconnect/tests/test_environment.py
@@ -5,10 +5,11 @@ from unittest import TestCase
 from os.path import dirname, join
 
 from rsconnect.environment import (
-    detect_environment,
+    Environment,
     EnvironmentException,
-    get_python_version,
+    detect_environment,
     get_default_locale,
+    get_python_version,
 )
 from rsconnect.tests.test_data_util import get_dir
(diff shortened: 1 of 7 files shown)


rsconnect/tests/test_environment.py
```diff
@@ -32,55 +35,46 @@ def test_get_default_locale(self):
...
-            },
-        )
+        for key, value in {
+            "package_manager": "pip",
+            "source": "file",
+            "filename": "requirements.txt",
+            "contents": "numpy\npandas\nmatplotlib\n",
+            "python": self.python_version(),
+        }.items():
+            self.assertEqual(getattr(result, key), value)
```
mmarchetti, line 50: nit: you could construct an expected Environment here and compare to the one you got back.

mmarchetti: ok, maybe not, forgot we pop things out before comparing

meatballhat: no I think you're right! changing it...

PR #134, opened 2020-06-17
Better error message on app mode mismatch

mmarchetti:
### Description

This PR adds a check of the app mode being deployed vs. the saved metadata. A mismatch is considered an error, since the deployment will fail at Connect if allowed to proceed. A message is shown advising use of `--new` to create a new deployment on the desired type. 

Connected to #https://github.com/rstudio/connect/issues/17727

* Deploy an app or API.
* Attempt to redeploy the same directory as a different app mode. It should fail showing the message.
* Redeploy the same directory as the original app mode. It should succeed.


kgartland-rstudio:
Verified.

PR #136, opened 2020-06-17
Release to github, PYPI, and S3 via actions

meatballhat:
### Description

(as opposed to using `release.Jenkinsfile`)

including copying the bdist wheel as `latest` on pushes to the `master` branch for use with rsconnect-python integration tests elsewhere.

- kinda sorta `N/A`, unless publishing prereleases or draft releases or some other such thing is desired


Makefile
```diff
@@ -112,10 +114,29 @@ fmt: fmt-3.8
...
+	@echo $(VERSION)
+
 # NOTE: Wheels won't get built if _any_ file it tries to touch has a timestamp
 # before 1980 (system files) so the $(SOURCE_DATE_EPOCH) current timestamp is
 # exported as a point of reference instead.
 .PHONY: dist
 dist:
-	python setup.py sdist bdist_wheel && \
-		rm -vf dist/*.egg
+	pipenv run python setup.py bdist_wheel
```
Lytol, line 126: If we're not creating a [universal wheel](https://packaging.python.org/guides/distributing-packages-using-setuptools/#universal-wheels) and we're not doing an sdist alongside the wheel, doesn't this mean that we aren't supporting Python 2?  (I'm not great at python packaging, so maybe I'm missing something)

meatballhat: @Lytol we are generating a universal wheel 👍 https://github.com/rstudio/rsconnect-python/blob/83d663abb4cf57c3a5f098e35ecd9c017939a7ad/setup.cfg#L1-L3

Lytol: Nice, I knew I was likely just missing something. Thanks for pointing it out!

PR #142, opened 2020-06-23
Address a few nits and a bug

meatballhat:
### Description

found during rsconnect-juptyer testing

Connected to rstudio/connect#17780

- deployment via `rsconnect-jupyter` does not fail when the `conda` path is taken


meatballhat:
fwiw I'm working to add `mypy` in a separate PR so that we can catch bugs like this in the future:

```
rsconnect/environment.py:71: error: Incompatible return value type (got "Dict[str, str]", expected "Environment")
Found 1 error in 1 file (checked 9 source files)
```

PR #144, opened 2020-06-24
Invert compatibility mode for conda mode

meatballhat:
### Description

as intended/expected when used as a library

Connected to rstudio/connect#17604

- behavior matches expected such that ~when receiving `compatibility_mode=True`, `conda_mode=False` will be passed along~ only `conda_mode` is accepted and is `False` by default

### DIFF
--- a/rsconnect/actions.py
+++ b/rsconnect/actions.py
@@ -481,7 +481,9 @@ def deploy_jupyter_notebook(
     (app_id, deployment_name, deployment_title, default_title, app_mode,) = gather_basic_deployment_info_for_notebook(
         connect_server, app_store, file_name, new, app_id, title, static
     )
-    python, environment = get_python_env_info(file_name, python, compatibility_mode, force_generate)
+    python, environment = get_python_env_info(
+        file_name, python, conda_mode=not compatibility_mode, force_generate=force_generate,
+    )
     bundle = create_notebook_deployment_bundle(file_name, extra_files, app_mode, python, environment)
     return _finalize_deploy(
         connect_server,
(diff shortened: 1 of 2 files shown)


rsconnect/tests/test_actions.py
```diff
@@ -303,3 +310,89 @@ def test_inspect_environment(self):
...
+            False,
+            True,
+            "unused",
+            Environment(error="Could not even do things"),
+            id="exploding",
+        ),
+    ],
+)
+def test_get_python_env_info(
+    monkeypatch, file_name, python, conda_mode, force_generate, expected_python, expected_environment,
```
scottmmjackson, line 375: nit: prefer `runtime_patch` to `monkeypatch`

meatballhat: this is a builtin pytest fixture named `monkeypatch` and I don't see one named `runtime_patch` 😕

rsconnect/actions.py
```diff
@@ -481,7 +481,9 @@ def deploy_jupyter_notebook(
     (app_id, deployment_name, deployment_title, default_title, app_mode,) = gather_basic_deployment_info_for_notebook(
         connect_server, app_store, file_name, new, app_id, title, static
     )
-    python, environment = get_python_env_info(file_name, python, compatibility_mode, force_generate)
+    python, environment = get_python_env_info(
+        file_name, python, conda_mode=not compatibility_mode, force_generate=force_generate,
```
scottmmjackson, line 485: I get the rationale behind not altering the arg name in the func signature, although I feel like here you have a workable argument that a mismatching `rsconnect-jupyter` version is absolutely unsupported and crashing on a syntax error may actually count as more graceful degradation than surreptitiously capturing a user's conda environment.

rsconnect/tests/test_actions.py
scottmmjackson: oh whoops, I thought this was our choice. welp.

rsconnect/actions.py
meatballhat: Is your recommendation that `compatibility_mode` be changed to `conda_mode` throughout this project?

scottmmjackson: Yes, although I think doing so could be a separate issue due to the potential fallout on the jupyter plugin from calling the function incorrectly.

Issue #156, opened 2020-09-08
UnicodeDecodeError on python2


mmarchetti:
Rendering a notebook with binary data in an output cell causes a UnicodeDecodeError upon publishing.

For example, create a notebook with an input cell `!cat filename.jpeg` (where `filename.jpeg` is a file in the notebook directory). Execute the input cell, then attempt to deploy the notebook.

PR #157, opened 2020-10-09
Fix python2 encoding issue

mmarchetti:
### Description

This PR fixes `to_bytes` so it doesn't attempt to re-encode bytes objects.

Connected to #156 

Use rsconnect-jupyter to publish a notebook that includes binary data in output cells.

For example, create a notebook with an input cell `!cat filename.jpeg` (where filename.jpeg is a file in the notebook directory). Execute the input cell, then attempt to publish the notebook.


Lytol (approved):
:+1:

Issue #155, opened 2020-08-28
Exceptions raised from actions.verify_server lack detail


mmarchetti:
The except clause here: 
https://github.com/rstudio/rsconnect-python/blob/071378d049873854cd049017da83f3ea925d2019/rsconnect/actions.py#L201

masks the error details provided here:
https://github.com/rstudio/rsconnect-python/blob/071378d049873854cd049017da83f3ea925d2019/rsconnect/api.py#L40-L57

This makes troubleshooting connection issues more difficult, for example, on the rsconnect-jupyter UI is just says "failed to verify as RStudio Connect".

PR #158, opened 2020-10-14
Preserve exception details

mmarchetti:
### Description

This PR preserves the details of exceptions from `verify_server`. Currently, the detailed exception messages are being discarded and they are valuable for troubleshooting.

Connected to #155

Create various error conditions when adding a server via `rsconnect add`:
* a hostname that cannot be resolved
* incorrect port number
* user http when https is required

Verify that the exception details are shown. Also try this in rsconnect-jupyter's Add Server window.


Lytol (approved):
:+1:

Issue #166, opened 2020-12-17
pip freeze in a conda environment produces unusable requirements.txt


christierney:
For a conda client, pip freeze seems to generate a requirements.txt with local file references.

see https://rstudioide.zendesk.com/agent/tickets/55886

`pip list --format=freeze > requirements.txt` seems to do the right thing.

see also https://github.com/pypa/pip/pull/7612

mmarchetti:
`pip list --format=freeze` helps a lot. There are sometimes package versions in conda environments that don't exist for pip, e.g. we saw `setuptools=49.6.0.post` which is not available on PyPI.

ekungurov:
Also do we have any documentation on how to use conda_mode=True ?

PR #167, opened 2020-12-17
Use pip list --format=freeze instead of pip freeze

mmarchetti:
### Description

In conda environments, `pip freeze` lists paths in the requirements.txt file which do not exist locally. This is caused by the PEP 610 metadata present for conda-installed packages, and is exposed by a change in pip 20.1b1.

Connected to #166 

Publish from virtualenvs and conda envs, without an existing requirements.txt file.


joncfoo (approved):

scottmmjackson (approved):
Do we need to note somewhere that versions up to this are broken with pip 20.1?

PR #168, opened 2021-01-22
Add support for HTTPS_PROXY

mmarchetti:
This is a customer-submitted patch to add support for HTTPS_PROXY. Submitting as a PR for review and testing.

### DIFF
--- a/rsconnect/http_support.py
+++ b/rsconnect/http_support.py
@@ -4,6 +4,7 @@ HTTP support wrappers and utility functions
 import json
 import socket
 import ssl
+import os
 
 from rsconnect import VERSION
 from rsconnect.log import logger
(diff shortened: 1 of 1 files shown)


joncfoo (approved):
LGTM. I can see this would solve the problem.

----

What do you think about:
1. should we parse the URL instead of splitting on `:`?
2. de-dupe the proxy code and have `context` be the only thing that is toggled

rsconnect/http_support.py
```diff
@@ -55,11 +56,39 @@ def _create_ssl_connection(host_name, port, disable_tls_check, ca_data, timeout)
         )
     elif disable_tls_check:
         # noinspection PyProtectedMember
-        return http.HTTPSConnection(
-            host_name, port=(port or http.HTTPS_PORT), timeout=timeout, context=ssl._create_unverified_context(),
-        )
+        proxyURL = os.getenv("HTTPS_PROXY")
+        if proxyURL:
+            print("Using custom proxy server - TLS check disabled {}".format(proxyURL))
```
scottmmjackson, line 61: prefer `logger`

rsconnect/http_support.py
```diff
@@ -55,11 +56,39 @@ def _create_ssl_connection(host_name, port, disable_tls_check, ca_data, timeout)
...
+        else:
+            tmp = http.HTTPSConnection(
+                host_name, port=(port or http.HTTPS_PORT), timeout=timeout, context=ssl._create_unverified_context(),
+            )
+        return tmp
     else:
-        return http.HTTPSConnection(host_name, port=(port or http.HTTPS_PORT), timeout=timeout)
+        proxyURL = os.getenv("HTTPS_PROXY")
+        if proxyURL:
+            print("Using custom proxy server {}".format(proxyURL))
```
scottmmjackson, line 80: prefer `logger`

rsconnect/http_support.py
```diff
@@ -55,11 +56,39 @@ def _create_ssl_connection(host_name, port, disable_tls_check, ca_data, timeout)
...
+            tmp.set_tunnel(host_name, (port or http.HTTPS_PORT))
+        else:
+            tmp = http.HTTPSConnection(
+                host_name, port=(port or http.HTTPS_PORT), timeout=timeout, context=ssl._create_unverified_context(),
+            )
+        return tmp
     else:
-        return http.HTTPSConnection(host_name, port=(port or http.HTTPS_PORT), timeout=timeout)
+        proxyURL = os.getenv("HTTPS_PROXY")
+        if proxyURL:
```
scottmmjackson, line 79: I agree with @joncfoo - we should parse the URL. We don't want to test for falsiness, we want to test for URL validity.

Otherwise we're going to get into trouble with trailing slashes and stuff like that. The coercion to `int` below would just barf on a trailing slash and the exception would flow all the way up the stack.

scottmmjackson (approved):
Approve but with three changes + agree with @joncfoo we should urlparse instead of string massaging.

PR #183, opened 2021-06-10
Add option to hide code cell in Jupyter notebooks

bcwu:
Add option to hide code cells when rendering Jupyter notebooks

### Description

This commit adds requested feature to hide code cells when rendering Jupyter notebooks.

### Usage example

After setting up Connect and rsconnect-python, the user can render a Jupyter notebook without its corresponding code cells by passing the '--hide-all-input' flag through the rsconnect cli:

```
rsconnect deploy notebook \
    -n server \
    -k APIKey \
    --hide-all-input \
    hello_world.ipynb
```

To selectively hide the input of cells, the user can add a tag call 'hide_input' to the cell, then pass the ' --hide-tagged-input' flag through the rsconnect cli:

```
rsconnect deploy notebook \
    -n server \
    -k APIKey \
    --hide-tagged-input \
    hello_world.ipynb
```

### DIFF
--- a/rsconnect/actions.py
+++ b/rsconnect/actions.py
@@ -455,8 +455,8 @@ def deploy_jupyter_notebook(
     conda_mode=False,
     force_generate=False,
     log_callback=None,
-    no_input=None,
-    no_tag_input=None,
+    hide_all_input=None,
+    hide_tagged_input=None,
 ):
     """
     A function to deploy a Jupyter notebook to Connect.  Depending on the files involved
@@ -480,8 +480,8 @@ def deploy_jupyter_notebook(
     (the default) the lines from the deployment log will be returned as a sequence.
     If a log callback is provided, then None will be returned for the log lines part
     of the return tuple.
-    :param no_input: if True, will hide notebook code cells when rendering to output
-    :param no_tag_input: If True, hides the input of cells tagged with "remove_input"
+    :param hide_all_input: if True, will hide all input cells when rendering output
+    :param hide_tagged_input: If True, will hide input code cells with the 'hide_input' tag when rendering output
     :return: the ultimate URL where the deployed app may be accessed and the sequence
     of log lines.  The log lines value will be None if a log callback was provided.
     """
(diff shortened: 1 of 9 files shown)


rsconnect/bundle.py
```diff
@@ -270,6 +276,8 @@ def make_notebook_html_bundle(
         "--to=html",
         filename,
     ]
+    if noinput:
+        cmd.append('--no-input')
```
mmarchetti, line 280: I notice that the nbconvert option is named `--no-input` and ours is `--noinput`. Should we match the nbcvonert name exactly?

mmarchetti (commented):
This looks good. We should hold off merging until the corresponding Connect change is ready.

rsconnect/bundle.py
mmarchetti: Also, if we're going to offer the option to hide tagged input cells we would need to include that here as well.

rsconnect/actions.py
```diff
@@ -1172,7 +1180,13 @@ def spool_deployment_log(connect_server, app, log_callback):
...
 
 def create_notebook_manifest_and_environment_file(
-    entry_point_file, environment, app_mode=None, extra_files=None, force=True
+    entry_point_file,
+    environment,
+    app_mode=None,
+    extra_files=None,
+    force=True,
+    no_input=None,
+    no_tag_input=None,
```
mmarchetti, line 1189: should these default to False?

rsconnect/bundle.py
```diff
@@ -204,6 +208,8 @@ def make_notebook_source_bundle(
     file,  # type: str
     environment,  # type: Environment
     extra_files=None,  # type:  typing.Optional[typing.List[str]]
+    no_input=None,
+    no_tag_input=None,
```
mmarchetti, line 212: same here

mmarchetti (commented):
Let a couple comments but the changes look good. Please hold off on merging until we are ready.

mmarchetti (approved):

PR #184, opened 2021-06-10
More flexible default entrypoints

mmarchetti:
### Description

This PR provides a method of choosing an entrypoint if none was specified with the `--entrypoint` parameter. Previously, the default was hardcoded to `app` which will fail if there is no `app.py`.

Deploy typical example apps without specifying an entrypoint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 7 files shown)


bcwu (approved):

kgartland-rstudio:
verified with app.py, application.py, main.py, api.py.  Logs also show which entrypoint is being used which is nice:
`Loading code from "api"`

Even using a nonsense filename works e.g. `blerp.py`.  
When using multiple nonsense `.py` files in the target dir, we fall back to looking for one of the default filenames.

However, when using multiple supported filenames it's hard to know which one we will choose.  
Example:
when a file named `app.py` and a `main.py` are found, we seem to prefer `app.py`.

I think this can be cleared up with documentation.

Issue #188, opened 2021-07-12
Exclude __pycache__ by default


mmarchetti:
The default exclusion list [here](https://github.com/rstudio/rsconnect-python/blob/master/rsconnect/bundle.py#L26) should include the `__pycache__` folder.

PR #193, opened 2021-07-28
Exclude __pycache__ by default

mmarchetti:
### Description

When creating a bundle for a directory, exclude the `__pycache__` which is created during local runs.

Connected to #188

Run an app locally with python3, then deploy it. Download the bundle from Connect and verify that the `__pycache__` directory was not included in the bundle.


kgartland-rstudio:
Verified, __pycache__ is omitted when deploying from this branch of rsconnect-python.

kgartland-rstudio (approved):

PR #196, opened 2021-07-30
Warn if no requirements.txt file

mmarchetti:
### Description

Add a warning during deployment to encourage users to create a requirements.txt file, if one is not found.  

Many users don't have a requirements.txt file in their projects. This leads to longer deploy times and bloat on the server-side as we replicate the user's entire local environment. 

Deploy both notebooks and APIs/apps, with and without an existing requirements.txt file.

### DIFF
--- a/rsconnect/main.py
+++ b/rsconnect/main.py
@@ -383,7 +383,23 @@ def _warn_on_ignored_manifest(directory):
     """
     if exists(join(directory, "manifest.json")):
         click.secho(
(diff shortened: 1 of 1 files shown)


mmarchetti:
Warning looks like this:

Screen Shot 2021-08-02 at 9 10 13 AM

bcwu (approved):

kgartland-rstudio:
Looks good, tested against API and Notebooks with and without requirements.txt file.

PR #198, opened 2021-08-06
Exclude environment directories

mmarchetti:
### Description

This PR adds a check for virtualenvs when bundling, to prevent accidentally uploading a virtualenv.

It also adds `renv` to the default exclusion list for R/reticulate users.

Create a virtualenv in the project directory; make sure its name is not in the default exclusion list. It doesn't need to be active. Deploy and then download the bundle from Connect to verify its contents.

### DIFF
--- a/rsconnect/bundle.py
+++ b/rsconnect/bundle.py
@@ -402,14 +402,19 @@ def create_glob_set(directory, excludes):
     return GlobSet(work)
 
 
+def is_environment_dir(directory):
+    python_path = join(directory, "bin", "python")
+    return exists(python_path)
+
+
 def list_environment_dirs(directory):
     # type: (...) -> typing.List[str]
     """Returns a list of subdirectories in `directory` that appear to contain virtual environments."""
     envs = []
 
     for name in os.listdir(directory):
-        python_path = join(directory, name, "bin", "python")
-        if exists(python_path):
+        path = join(directory, name)
+        if is_environment_dir(path):
             envs.append(name)
     return envs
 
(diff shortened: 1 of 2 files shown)


colearendt (approved):
LGTM! Very elegant!

rstub:
Does this also take care of the case where the virtual environment is not in a subdirectory but in the app directory itself?, i.e. a situation like
```
flask-api/bin/activate
flask-api/bin/python
[...]
flask-api/lib/...
[...]
flask-api/api.py
flask-ap/requirements.txt
```

mmarchetti:
No, it only handles the subdirectory case, where it can exclude the entire directory. This implementation doesn't know what's in a virtualenv, other than `bin/python`. To handle the case you've described, it would need to partition the files and directories in the deployment directory into "app files" vs. "python environment files".

bcwu (approved):

colearendt:
One thought I had - do virtualenvs ever have `bin/python3` instead of `bin/python`?

mmarchetti:
Virtualenv creates `python`, `python3`, and (e.g.) `python3.8` symlinks.

rstub:
Ok. Maybe you could add a warning if you find `bin/python` in the app's directory?

kgartland-rstudio:
The warning message only seems to display when `bin` and `lib` folders are in the top-level directory but does not display when they are in a subdirectory.  Do we want to warn in either case?
top-level:
```
.
├── app.py
├── bin
├── bologna
├── include
├── lib
├── pyvenv.cfg
├── requirements.txt
├── rsconnect-python
└── data
```
Warning is shown: 
```
Warning: The deployment directory appears to be a python virtual environment.
             Excluding the 'bin' and 'lib' directories.
```

sub-directory (where the env folders are under `bologna`):
```
.
├── app.py
├── bologna
├── pyvenv.cfg
├── requirements.txt
├── rsconnect-python
└── data
```

No warning is produced.


Additionally in the warning we only mention `bin` and `lib` directories, but we also exclude the `include` directory when using virtualenv.

mmarchetti:
We don't warn when detecting an environment in a subdirectory because we can safely exclude the whole directory. When it's the root directory (like you're deploying from inside the virtual environment directory itself) we don't know what to exclude, so we do what we can and warn that it might not be complete.

PR #209, opened 2021-11-29
Add content sub commands for searching and building content

dbkegley:
### Description

This PR contains an initial proposal of an `rsconnect-admin` CLI tool targeting Connect administrators.  

See `README_ADMIN.md` for an overview of the existing `search` and `rebuild` functionality.  The interfaces in this tool are meant to feel consistent with the existing CLI which is more geared toward publishers. The main exception to this is the default output format for most of the admin CLI commands is `json` so that it can be easily used by other scripts.

Getting this out here now so that we can start reviewing the initial implementation while I'm finishing up the integration tests.

edit:

Ive also modified this branch to use the new `v1/content/<guid>/build` endpoint. This means that the `rsconnect-admin build` subcommands have a dependency on `Connect >= 2021.11.1` and I renamed `rebuild` -> `build` everywhere.


rsconnect/metadata.py
```diff
@@ -420,3 +430,203 @@ def resolve(self, server, app_id, app_mode):
...
+            t = datetime.fromtimestamp(os.path.getctime(f), tz=timezone.utc).strftime('%Y-%m-%d %H:%M:%S%z')
+            history.append({'time': t, 'task_id': task_id})
+        history.sort(key=lambda x: x['time'])
+        return history
+
+    def get_rebuild_running(self, server):
+        return self._data.get(server.url, {}).get('rsconnect_rebuild_running')
+
+    def set_rebuild_running(self, server, is_running):
+        self._lock.acquire()
```
mmarchetti, line 518: The lock/unlock sequences look correct. Since they are held locally, you have the option to simplify them using a `with` clause as described here: https://docs.python.org/3/library/threading.html#using-locks-conditions-and-semaphores-in-the-with-statement

```
with self._lock:
    self._data[server.url]['rsconnect_rebuild_running'] = is_running
    self.save()
```

setup.cfg
```diff
@@ -19,6 +19,7 @@ install_requires =
     six>=1.14.0
     click>=7.0.0
     pip>=10.0.0
+    semver
```
mmarchetti, line 22: Should this have a version spec? The `semver` package has started 3.0.0dev builds. Assuming they are following semver, they will likely release a breaking change soon. We haven't pinned major versions on the other dependencies, but this seems like one we should consider.

rsconnect/metadata.py
```diff
@@ -420,3 +430,203 @@ def resolve(self, server, app_id, app_mode):
...
+    def get_content_items(self, server, status=None):
+        """
+        Get all the content items that are tracked for rebuild on the given server.
+        :param status: Filter results by rebuild status
+        :return: A list of content items
+        """
+        if status:
+            return [item for item in self._data.get(server.url, {}).get('content', {}).values() if item['rsconnect_rebuild_status'] == status]
+        else:
+            return list(self._data.get(server.url, {}).get('content', {}).values())
```
mmarchetti, line 632: nit: could factor out `self._data.get(server.url, {}).get('content', {}).values()` to make it clearer that the difference in the 2 cases is filtering vs. not filtering the same list.

rsconnect/api.py
```diff
@@ -9,6 +9,19 @@
 from rsconnect.log import logger
 from rsconnect.models import AppModes
 
+_ok_http_status = [
```
mmarchetti, line 12: This seems pretty specific - would it be fair to characterize all 2xx codes as OK?

rsconnect/admin_main.py
```diff
@@ -0,0 +1,671 @@
...
+_version_search_pattern = r"(^[=><]{1,2})(.*)"
+
+def _verify_rebuild_rm_args(guid, all, purge):
+    if guid and all:
+        raise api.RSConnectException("You must specify only one of -g/--guid or --all, not both.")
+    if not guid and not all:
+        raise api.RSConnectException("You must specify one of -g/--guid or --all.")
+
+@click.group(no_args_is_help=True)
+@click.option("--future", "-u", is_flag=True, hidden=True, help="Enables future functionality.")
```
mmarchetti, line 50: I like the concept of the `--future` option - what functionality does it expose?

rsconnect/admin_actions.py
```diff
@@ -0,0 +1,338 @@
+"""
+Public API for administering content.
+"""
+import contextlib
+import json
+import time
+import sys
+
+# This probably breaks python2, can we remove python2.7 support from setup and/or can we require >3 for only the admin tool?
```
mmarchetti, line 9: This is a good question for @kellobri. Most python tooling is requiring 3.6 or later at this point. Publishing a new version that requires python3 leaves the old version available to python2 users (but strands them on that version).

rsconnect/admin_actions.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+
+def emit_rebuild_log(connect_server, guid, format, task_id=None):
+    log_file = content_rebuild_store.get_rebuild_log(connect_server, guid, task_id)
+    if log_file:
+        with open(log_file, 'r') as f:
+            for line in f.readlines():
+                if format == "json":
+                    yield json.dumps({"message": line}) + "\n"
```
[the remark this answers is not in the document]
dbkegley: I considered this but opted against it because _sometimes_ the output from the task logs includes a timestamp and it felt wrong to add another one on the client side.  And then that also begs the question: do we mutate the incoming log lines by prefixing them with the timestamp when they are written to disk?

rsconnect/admin_actions.py
```diff
@@ -0,0 +1,338 @@
...
+    return list(result)
+
+
+def _apply_content_filters(content_list, published, unpublished, content_type, r_version, py_version, title_search):
+    def content_is_published(item):
+        return 'bundle_id' in item and item['bundle_id'] != None
+
+    def content_is_unpublished(item):
+        if 'bundle_id' not in item:
+            return False
```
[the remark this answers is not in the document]
dbkegley: That's right, a null `bundle_id` indicates unpublished content. I think both of these conditionals could be simplified:

```python
def content_is_published(item):
    return item.get('bundle_id') != None

def content_is_unpublished(item):
    return item.get('bundle_id') == None
```

rsconnect/admin_main.py
```diff
@@ -0,0 +1,671 @@
...
+    help="Order content results.",
+)
+@click.option("--verbose", "-v", is_flag=True, help="Print detailed messages.")
+# todo: --format option (json, text)
+def content_search(name, server, api_key, insecure, cacert, published, unpublished, content_type, r_version, py_version, title_contains, order_by, verbose):
+    set_verbosity(verbose)
+    with cli_feedback("", stderr=True):
+        connect_server = _validate_deploy_to_args(name, server, api_key, insecure, cacert)
+
+        with open_file_or_stdout("-") as f:
```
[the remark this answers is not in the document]
dbkegley: Ah yes, this is a holdover from when I was playing with different output flags. It was useful when I had an output flag that accepted a file path. This way I could default to stdout when no file path was provided.

https://github.com/rstudio/rsconnect-python/pull/209/commits/5d252c2ff03ce90a22aa255d5b76a6a043b40577#diff-bb8011cd3b75f7a586d773ca22e4a45434fc8fbf268fab85c2e5ed5ffbef4530L152-L158

Anyway, this isn't used anymore so we can remove that helper, good call

rsconnect/admin_main.py
dbkegley: None yet. I shamelessly stole this from our existing main.py. Not needed by admin_main.py though, so it can be removed

rsconnect/api.py
```diff
@@ -152,7 +178,22 @@ def deploy(self, app_id, app_name, app_title, title_is_default, tarball):
...
+        results = self.bundle_download(content_guid, bundle_id)
+        self._server.handle_bad_response(results)
+        return results
+
+    def search_content(self):
+        results = self.content_search()
+        self._server.handle_bad_response(results)
+        return results
+
+    def get_content(self, content_guid):
```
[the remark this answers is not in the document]
dbkegley: How about this: https://github.com/rstudio/rsconnect-python/pull/209/commits/a2fcedf07d671cb258a7c0dc2deda64c2d1d4055

I removed all the new `api.do_*` methods and we're now handling the HTTP response in the client methods. Then we can use the client directly from `actions`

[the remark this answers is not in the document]
mmarchetti: That seems reasonable

dbkegley:
@mmarchetti @kellobri One thing that has come up during our discussions a couple times is that there is some inconsistency with this tool because of the introduction of the new `rsconnect-admin` binary. @aronatkins pointed out that it's awkward to do `rsconnect add` to add a new server alias, and then expect to use that same alias from `rsconnect-admin` commands. Another thing to consider is that the `content` sub-commands are not necessarily admin-specific either.  What are your thoughts on continuing to use a single binary `rsconnect` and introducing a new `admin` sub-command? This way it's clear to publishers that they can use the `content` subcommands as well

e.g. 

```
$ rsconnect admin build ...

$ rsconnect content search ...
$ rsconnect content download-bundle ...
```

aronatkins:
One other minor option: Shift "build" to also be a sub-command of "content", giving:

```
rsconnect content build ...
```

README_ADMIN.md
```diff
@@ -0,0 +1,218 @@
...
+RStudio Connect caches R and Python packages in the configured
+[`Server.DataDir`](https://docs.rstudio.com/connect/admin/appendix/configuration/#Server.DataDir).
+Sometimes, these package caches become stale and need to be rebuilt. This refresh
+occurs automatically when an RStudio Connect user visits the content via the
+RStudio Connect dashboard. Some content, however, is high priority or not accessed
+frequently via the dashboard (API content, emailed reports). In these cases, administrators
+can preemptively build specific content items using the `rsconnect-admin build` subcommands.
+This way the user does not have to pay the build cost when the content is accessed next.
+
+The following are some common scenarios where performing a content build might be necessary:
```
sagerb, line 64: It feels like there is an overlap between the paragraph above and these reasons to rebuild. Perhaps you can narrow the paragraph above down a little? I've wondered if the order should switch, if that would be clearer to have what the common scenarios might be before you explain the technical details of the caching / automatic building.

README_ADMIN.md
```diff
@@ -0,0 +1,218 @@
...
+
+To build a specific content item, first `add` it to the list of content that is
+"tracked" for building using its GUID.
+
+```bash
+# add a single content item by guid
+$ rsconnect-admin build add --guid 4ffc819c-065c-420c-88eb-332db1133317
+```
+
+Administrators can view currently "tracked" content items using the
```
sagerb, line 83: It feels like this is a sequence.. but maybe it isn't? If you intended it to be, then you might consider a numeric list for the steps. If it isn't, then perhaps find a way to describe each subcommand in isolation? (I think I'm being thrown off by the `first add it...` words above.

If it is a sequence, I'd re-order, putting run right after add, which would move ls downward to use after building has been initiated. Unless you want to ls twice.. once before starting and then again after, with different options.

README_ADMIN.md
```diff
@@ -0,0 +1,218 @@
...
+`rsconnect-admin build ls` subcommand.
+
+```bash
+# list all tracked content
+$ rsconnect-admin build ls
+
+# list tracked content that still needs to be built
+$ rsconnect-admin build ls --status NEEDS_BUILD
```
sagerb, line 91: If you go with describing each subcommand, then that will give you a good place to describe the parameters and their usage (like status).

README_ADMIN.md
```diff
@@ -0,0 +1,218 @@
...
+
+# list tracked content that has already been built successfully
+$ rsconnect-admin build ls --status COMPLETE
+
+# list tracked content that failed to build
+$ rsconnect-admin build ls --status=ERROR
```
sagerb, line 97: This one uses `status=ERROR`, but the one above used `status COMPLETE`. Which is correct since it seems like these should be consistent. (Aborted below is the same way)

README_ADMIN.md
```diff
@@ -0,0 +1,218 @@
...
+By default the `rsconnect-admin build logs` subcommand prints the logs of the
+most recent build. To view the build logs for a specific build, provide a
+`--task-id` to the `logs` subcommand.
+
+```bash
+$ rsconnect-admin build logs --guid 4ffc819c-065c-420c-88eb-332db1133317 --task-id GoTVLYxWkbvCo2bN
+```
+
+
+### Common Usage Examples
```
sagerb, line 143: If the upper part is just reference, then the first common usage example here could be the general add, build, ls, debug failed ones, you had described above in some ways.

README_ADMIN.md
```diff
@@ -0,0 +1,218 @@
...
+
+Many of the `rsconnect-admin` CLI subcommands are intended to be easily scriptable.
```
sagerb, line 145: Make this a "Note:"...

sagerb (commented):
Hopefully some README feedback which will help organize the commands a bit better. I know if you end up pulling these commands out of the `rsconnect-admin`, then their documentation will probably need to fit better within the many commands which `rsconnect` has. Hopefully, my suggestions will work there as well. 

But.. take them or leave them.. its a README after all.. :-)

PR #219, opened 2022-01-31
update proxy parsing to better parse urls with fields separated by colon

### LINKED ISSUE #203: HTTPS_PROXY parsing breaks when supplying username and password
In our environment, a proxy username and password are required. This is in the format:

`user:password@proxy:port`

The logic used at https://github.com/rstudio/rsconnect-python/blob/master/rsconnect/http_support.py#L41

will split this into:

`['user', 'password@proxy', 'port']`

When trying to unpack that here:

https://github.com/rstudio/rsconnect-python/blob/master/rsconnect/http_support.py#L43

It will fail and error out.

```
  File "<stdin>", line 2, in <module>
ValueError: too many values to unpack (expected 2)
```

Ideally, it should be able to accurately parse the string whether someone has a username:password or not.

bcwu:
fixes #203

### DIFF
--- a/rsconnect/http_support.py
+++ b/rsconnect/http_support.py
@@ -38,11 +38,11 @@ def _get_proxy():
     if proxyURL:
         logger.info("Using custom proxy server {}".format(proxyURL))
         parsed = urlparse(proxyURL)
-        netloc = parsed.netloc.split(":")
-        if len(netloc) > 1:
-            proxyHost, proxyPort = netloc
+        if parsed.port:
+            netloc = parsed.netloc.split(":")
+            proxyHost, proxyPort = ':'.join(netloc[:-1]), parsed.port
         else:
-            proxyHost = netloc
+            proxyHost = parsed.hostname
             proxyPort = 8080
 
         return proxyHost, int(proxyPort)


rsconnect/http_support.py
```diff
@@ -39,10 +39,10 @@ def _get_proxy():
         logger.info("Using custom proxy server {}".format(proxyURL))
         parsed = urlparse(proxyURL)
         netloc = parsed.netloc.split(":")
-        if len(netloc) > 1:
-            proxyHost, proxyPort = netloc
+        if parsed.port:
+            proxyHost, proxyPort = ':'.join(netloc[:-1]), parsed.port
         else:
-            proxyHost = netloc
+            proxyHost = ':'.join(netloc[:-1])
```
mmarchetti, line 45: #203 talks about the case where the proxy requires a username and password to access.

`urlparse` returns these in the `username` and `password` fields of the parsed URL object, while also returning them as part of the `netloc`.

This function should return the username and password as well (or perhaps the parsed URL?). Since the username and password are already stored separately, we can drop them from the netloc (perhaps with something like `netloc = netloc.split("@")[0]`).

mmarchetti: We also need to pass the proxy username and password in the `Proxy-Authorization` header to `set_tunnel` (for https) or with each request (for http). For example, see https://willkeeling.wordpress.com/2019/03/16/proxy-authentication-with-httpconnection-and-httpsconnection/.

mmarchetti (changes requested):

mmarchetti:
We should hold off merging this until 1.7.0 ships

mmarchetti (approved):

mmarchetti (approved):

rsconnect/http_support.py
```diff
@@ -37,8 +37,10 @@ def _get_proxy():
     proxyURL = os.getenv("HTTPS_PROXY")
     if not proxyURL:
         return None, None, None, None
-    logger.info("Using custom proxy server {}".format(proxyURL))
     parsed = urlparse(proxyURL)
+    if parsed.scheme not in ["https"]:
+        raise Exception("HTTPS_PROXY scheme must be https://")
+    logger.info("Using custom proxy server {}".format(proxyURL.replace(parsed.password or "", "REDACTED")))
```
mmarchetti, line 43: replacing `""` with anything will do a replacement at every point in the string

bcwu: Oops...

mmarchetti (approved):

kgartland-rstudio:
Sorry for the delay on this @bcwu. This looks good now!

PR #221, opened 2022-02-07
Support deploying manifests with Quarto app modes.

toph-allen:
#### Problem

`rsconnect-python` refuses to deploy Quarto manifests because it does not recognize the `quarto-static` and `quarto-shiny` appmodes.

#### Solution

- Added the new appmodes to the `AppModes` class in `models.py`.

- I searched the rest of the codebase for any other places that might need appmode-specific code, but couldn't find any. There don't seem to be any appmode-specific tests, either.
- I successfully deployed Quarto projects via `rsconnect deploy manifest` that had previously failed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -616,7 +616,7 @@ Options:
   -c, --cacert FILENAME           The path to trusted TLS CA certificates.
   --published                     Search only published content.
   --unpublished                   Search only unpublished content.
-  --content-type [unknown|shiny|rmd-static|rmd-shiny|static|api|tensorflow-saved-model|jupyter-static|python-api|python-dash|python-streamlit|python-bokeh|python-fastapi]
+  --content-type [unknown|shiny|rmd-static|rmd-shiny|static|api|tensorflow-saved-model|jupyter-static|python-api|python-dash|python-streamlit|python-bokeh|python-fastapi|quarto-shiny|quarto-static]
                                   Filter content results by content type.
   --r-version VERSIONSEARCHFILTER
                                   Filter content results by R version.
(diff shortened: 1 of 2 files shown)


toph-allen:
I don't think the README needs any additional updating, as it doesn't go into any detail about non-Python content anyway.

The one place I added them to the README is in the `--content-type` option under the Content Search heading. I checked, and that option does use `AppModes._modes`, so it makes sense to add them there.

mmarchetti (approved):
Is there a Python equivalent of quarto-shiny mode? (that is, a quarto document with an interactive python app embedded)?

kgartland-rstudio:
how do we `write-manifest `for a quarto app?  @toph-allen did you create the manifest manually?  

> I successfully deployed Quarto projects via rsconnect deploy manifest that had previously failed.

```
-> rsconnect version
1.7.1
```
```
> rsconnect write-manifest quarto-static quarto-jupyter.qmd
Usage: rsconnect write-manifest [OPTIONS] COMMAND [ARGS]...
Try 'rsconnect write-manifest --help' for help.

Error: No such command 'quarto-static'
```

toph-allen:
`rsconnect-python` can't write a manifest for Quarto+Python content yet (see https://github.com/rstudio/connect/issues/20729#issue-1130908463). `rsconnect::writeManifest()` is in the same position for Quarto+R content (see https://github.com/rstudio/connect/projects/26#card-66728381). And I don't know if we understand how Quarto content with no runtimes will get a manifest.

So right now, `rsconnect-python` just supports manifest-only deployment for Quarto appmodes, like it does for R content.

I've been getting manifests for Quarto content by using the IDE's push-button publishing to deploy them, then downloading the bundle and extracting the manifest from there.

PR #229, opened 2022-02-15
Allow setting environment variables during deployment

mmarchetti:
### Description

This is a little side project I did around the holidays.

You can specify environment variables with the new `-E` or `--environment` options to the `rsconnect deploy` commands, for example `rsconnect deploy api -n myserver -E NAME=VALUE .`. The environment variables will be set on the app right after it's created, before the bundle is uploaded and deployed, so they are present during the first render/run of the content on Connect.

Deploy a notebook and an app that requires an environment variable to be set in order for the code to run. You can do this by using something like `print("NAME", os.environ["NAME"])` in the code.
* Deploy without `-E` and see that it fails (at render time for a notebook, or at run time for an app).
* Redeploy with `-E NAME=me` (don't use `--new` so it reuses the app). See that the variable is set in Connect, and the app/notebook runs correctly.
* Deploy again with `-E NAME=you` and see that the variable in Connect has been updated with the new value.
* Try using multiple `-E` options on the command line. All of them should take effect, if you specify unique names. (Duplicate names will cause Connect to return an error).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- You can now set environment variables for the deployed content with the `-E` option.
+  These will be passed to RStudio Connect during the deployment process,
+  so they are available in your code whenever it runs within RStudio Connect.
+
 ## [1.7.0] - 2022-01-10
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - You can now set environment variables for the deployed content with the `-E` option.
   These will be passed to RStudio Connect during the deployment process,
   so they are available in your code whenever it runs within RStudio Connect.
+  Requires RStudio Connect version 1.8.6 or later.
 
 ## [1.7.0] - 2022-01-10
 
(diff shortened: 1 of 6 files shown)


mmarchetti:
> The environment variable value is visible to other users on the same host, as they can inspect running processes.
> 
>     * This has the same visibility as the API key; may need adjustment for multi-user environments.

I considered adding an option to pass a `.env` file, which would prevent the values being visible on the command line. There's a little more complexity which is fine. One concern I had is that the natural place to put that file is in the deployment directory, which of course puts it in the bundle (not great).

aronatkins:
The "leave it out of the bundle" challenge is real. I'm not sure the best path. `.rscignore`? We probably don't need to adjust before moving forward with this change.

Total brainstorm, but if the variables are part of the current environment, we could create an "inherit" option that enumerates environment variables that are read:

```bash
export IMPORTANT="apples every day"
rsconnect deploy -I IMPORTANT ...
```

This would be an alternative to the `-E` that you're proposing; they could coexist.

bcwu (approved):

kgartland-rstudio:
Validated.

Tested all scenarios above and it works well. Deployed bokeh, dash and notebook.

`rsconnect deploy bokeh ./ -n yeti-connect -E NAME=Kevin`
Displays Kevin wherever `os.environ["NAME"]` is defined in the app

`rsconnect deploy bokeh ./ -n yeti-connect -E NAME=Greta`
Overrides the NAME variable on a re-deploy

`rsconnect deploy bokeh ./ -n yeti-connect -E LOCAL_VAR`
Displays whatever value is saved as LOCAL_VAR

multiple variables can be included
`rsconnect deploy bokeh ./ -n yeti-connect -E NAME=Kevin -E SANDWICH=Bologna`

last variable in list wins
`rsconnect deploy bokeh ./ -n yeti-connect -E NAME=Kevin -E NAME=Anna -E NAME=Greta`

Variables are saved in the UI and can be edited there as well.

PR #236, opened 2022-03-15
deploy_html

bcwu:
### Description

Better support for deployment of static html files and bundles.

### QA/Testing

User should be able to deploy html content by passing a path (file or directory) to the new `deploy html` command

`rsconnect deploy html -n server index.html extra_file`

`rsconnect deploy html -n server folder`

### DIFF
--- a/rsconnect/bundle.py
+++ b/rsconnect/bundle.py
@@ -15,7 +15,7 @@ try:
 except ImportError:
     typing = None
 
-from os.path import basename, dirname, exists, isdir, join, relpath, splitext
+from os.path import basename, dirname, exists, isdir, join, relpath, splitext, isfile
 
 from .log import logger
 from .models import AppMode, AppModes, GlobSet
@@ -297,6 +297,7 @@ def make_notebook_source_bundle(
     return bundle_file
 
 
+#
 def make_quarto_source_bundle(
     directory,  # type: str
     inspect,  # type: typing.Dict[str, typing.Any]
@@ -539,58 +540,6 @@ def _create_api_file_list(
     return sorted(file_list)
 
 
-def _create_file_list(
-    path,  # type: str
-    extra_files=None,  # type: typing.Optional[typing.List[str]]
-    excludes=None,  # type: typing.Optional[typing.List[str]]
-):
-    # type: (...) -> typing.List[str]
-    """
-    Builds a full list of files under the given path that should be included
-    in a manifest or bundle.  Extra files and excludes are relative to the given
-    path and work as you'd expect.
-
-    :param path: the path to walk for files.
-    :param extra_files: a sequence of any extra files to include in the bundle.
-    :param excludes: a sequence of glob patterns that will exclude matched files.
-    :return: the list of relevant files, relative to the given path.
-    """
-    # Don't let these top-level files be added via the extra files list.
-    extra_files = extra_files or []
-    skip = ["manifest.json"]
-    extra_files = sorted(list(set(extra_files) - set(skip)))
-
-    # Don't include these top-level files.
-    excludes = list(excludes) if excludes else []
-    excludes.append("manifest.json")
-    if not os.path.isfile(path):
-        excludes.extend(list_environment_dirs(path))
-    glob_set = create_glob_set(path, excludes)
-
-    file_list = []
-
-    for rel_path in extra_files:
-        file_list.append(rel_path)
-
-    if os.path.isfile(path):
-        file_list.append(path)
-    else:
-        for subdir, dirs, files in os.walk(path):
-            for file in files:
-                abs_path = os.path.join(subdir, file)
-                rel_path = os.path.relpath(abs_path, path)
-
-                if keep_manifest_specified_file(rel_path) and (
-                    rel_path in extra_files or not glob_set.matches(abs_path)
-                ):
-                    file_list.append(rel_path)
-                    # Don't add extra files more than once.
-                    if rel_path in extra_files:
-                        extra_files.remove(rel_path)
-
-    return sorted(file_list)
-
-
 def make_api_manifest(
     directory,  # type: str
     entry_point,  # type: str
@@ -625,7 +574,7 @@ def make_api_manifest(
     return manifest, relevant_files
 
 
-def make_html_fileslist(
+def make_html_bundle_content(
     path,  # type: str
     entrypoint,  # type: str
     extra_files=None,  # type: typing.Optional[typing.List[str]]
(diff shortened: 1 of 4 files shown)


rsconnect/bundle.py
```diff
@@ -507,6 +509,92 @@ def make_api_manifest(
...
+        raise ValueError("Entrypoint is not a valid file type or directory.")
+
+    default_mimetype_entrypoints = {"text/html": "index.html"}
+    if mimetype not in default_mimetype_entrypoints:
+        raise ValueError("Not supported mimetype inference.")
+
+    entrypoint_candidates = []
+    mimetype_dict = defaultdict(list)
+
+    for subdir, dirs, files in os.walk(path):
```
mmarchetti, line 554: I'm not sure we need to talk the whole tree (though I may be missing the use case). It would be reasonable to only scan the toplevel directory here.

rsconnect/bundle.py
```diff
@@ -507,6 +509,92 @@ def make_api_manifest(
...
+        for file in files:
```
mmarchetti, line 555: Here, if `file` is the default entrypoint file (`index.html`) then you can return it.

rsconnect/bundle.py
```diff
@@ -507,6 +509,92 @@ def make_api_manifest(
...
+            abs_path = os.path.join(subdir, file)
+            rel_path = os.path.relpath(abs_path, path)
+            mimetype_dict[guess_type(file)[0]].append((file, rel_path))
+
+    for file, rel_path in mimetype_dict[mimetype]:
+        if file in default_mimetype_entrypoints[mimetype]:
+            entrypoint_candidates.append(rel_path)
+
+    return entrypoint_candidates.pop() if len(entrypoint_candidates) == 1 else None
```
mmarchetti, line 564: Otherwise, if I understand correctly, `mimetype_dict[mimetype]` is a list of the HTML files. If that list contains 1 item then it is the default entrypoint, otherwise we cannot choose automatically.

rsconnect/main.py
```diff
@@ -821,6 +827,105 @@ def deploy_manifest(name, server, api_key, insecure, cacert, new, app_id, title,
     )
 
 
+@server_args
+@content_args
+@click.option(
+    "--entrypoint",
+    "-e",
+    help=("The module and executable object which serves as the entry point for the deployment."),
```
mmarchetti, line 835: This help string should be updated.

rsconnect/main.py
```diff
@@ -821,6 +827,105 @@ def deploy_manifest(name, server, api_key, insecure, cacert, new, app_id, title,
...
+    path,
+    env_vars,
+    entrypoint,
+    extra_files,
+    excludes,
+):
+    set_verbosity(verbose)
+
+    entrypoint = entrypoint or infer_entrypoint(path, "text/html")
+    manifest = make_html_manifest(entrypoint)
```
mmarchetti, line 873: These actions are done in `make_html_bundle` so should not be repeated here.

rsconnect/main.py
```diff
@@ -821,6 +827,105 @@ def deploy_manifest(name, server, api_key, insecure, cacert, new, app_id, title,
...
+
+    with cli_feedback("Checking arguments"):
+        connect_server = _validate_deploy_to_args(name, server, api_key, insecure, cacert)
+        manifest = validate_manifest_file(manifest)
```
mmarchetti, line 877: This seems to be a leftover from `deploy manifest`. The manifest file is known to be good because rsconnect-python just created it (user-provided manifests need validation when using `deploy manifest`).

rsconnect/main.py
```diff
@@ -821,6 +827,105 @@ def deploy_manifest(name, server, api_key, insecure, cacert, new, app_id, title,
...
+
+        (
+            app_id,
+            deployment_name,
+            title,
+            default_title,
+            app_mode,
+            package_manager,
+        ) = gather_basic_deployment_info_from_manifest(connect_server, app_store, manifest, new, app_id, title)
+
```
mmarchetti, line 888: We should use a different version of `gather_basic_deployment_info` here.

rsconnect/main.py
```diff
@@ -821,6 +827,105 @@ def deploy_manifest(name, server, api_key, insecure, cacert, new, app_id, title,
...
+    click.secho('    Deploying %s to server "%s"' % (manifest, connect_server.url))
+
+    if package_manager == "conda":
+        with cli_feedback("Ensuring Conda is supported"):
+            check_server_capabilities(connect_server, [is_conda_supported_on_server])
```
mmarchetti, line 893: This should be eliminated since we are not deploying any Python code in this case.

rsconnect/main.py
```diff
@@ -821,6 +827,105 @@ def deploy_manifest(name, server, api_key, insecure, cacert, new, app_id, title,
...
+                error.args[1],
+            )
+            if error.args[0] == errno.ENOENT:
+                msg = "\n".join(
+                    [
+                        msg,
+                        "Since the file is missing but referenced in the manifest, "
+                        "you will need to\nregenerate your manifest.  See the help "
+                        'for the "write-manifest" command or,\nfor non-Python '
+                        'content, run the "deploy other-content" command.',
```
mmarchetti, line 910: This case is also not needed, because this is a generated manifest (so we know the files exist) vs. a user-provided manifest which might be out of date.

mmarchetti (changes requested):

rsconnect/main.py
```diff
@@ -869,28 +867,19 @@ def deploy_html(
...
 
         (
             app_id,
             deployment_name,
             title,
             default_title,
             app_mode,
-            package_manager,
-        ) = gather_basic_deployment_info_from_manifest(connect_server, app_store, manifest, new, app_id, title)
+        ) = gather_basic_deployment_info_for_notebook(connect_server, app_store, path, new, app_id, title, True)
```
mmarchetti, line 880: `gather_basic_deployment_info_for_notebook` will raise an exception if the give path isn't an `.ipynb` file. You might need a variant of that function specific to static deployments.

mmarchetti (approved):

bcwu:
Fixes https://github.com/rstudio/connect/issues/20944

bcwu:
> Summary:
> 
>     * deploying from current working directory with relative path fails.
>       i.e. `rsconnect deploy html ./index.html -n yeti --new`
> 
>     * deploying with the `--new` isn't respected.
> 
>     * the default title either uses the filename or folder name depending on if you use the folder or filename during deploys. Is this what we want?

[4c9ba72](https://github.com/rstudio/rsconnect-python/pull/236/commits/4c9ba72612acc25f98508a94829d259da32826c6) and [3fd9006](https://github.com/rstudio/rsconnect-python/pull/236/commits/3fd90067819ecd569657f8dff2c46ea40e3a15d1) should fix the first two issues.

--title or -t is available for deploying with a user specified name

kgartland-rstudio:
Now we seem to be re-creating the entire local folder structure in the bundle on Connect, I don't think we want to do that either.

For example:  Deploying a single file succeeds now when running:

`rsconnect deploy html ./index.html -n yeti`

Previously that would fail with too many redirects, but now it recreates my entire folder structure in the bundle.  
i.e.:
`/Users/kgartland/Projects/website/v2.vuejs.org/src/v2/examples/vue-20-elastic-header/index.html`

So when I go to deploy individual files that use relative paths (css, js) they fail.  i.e.
`rsconnect deploy html ./index.html ./css/style.css ./scripts/saved.js -n yeti`

My bundle looks like this:
index.html: `/Users/kgartland/Projects/website/v2.vuejs.org/src/v2/examples/vue-20-elastic-header/index.html`
styles.css: `./css/styles.css`
saved.js: `./scripts/saved.js`

Here's what that bundle looks like in my tree:

```
bundle-200]-> tree
.
├── Users
│   └── kgartland
│       └── Projects
│           └── website
│               └── v2.vuejs.org
│                   └── src
│                       └── v2
│                           └── examples
│                               └── vue-20-elastic-header
│                                   └── index.html
├── css
│   └── style.css
├── manifest.json
└── scripts
    └── saved.js

11 directories, 4 files
```

Here's the contents of the manifest I get back in the bundle:

```
{
  "version": 1,
  "metadata": {
    "appmode": "static",
    "primary_html": "/Users/kgartland/Projects/website/v2.vuejs.org/src/v2/examples/vue-20-elastic-header/index.html"
  },
  "files": {
    "./css/style.css": {
      "checksum": "94264a34917fd9077d384c495e99b77d"
    },
    "./scripts/saved.js": {
      "checksum": "3e2664e064c50a0e8d3ba83081826a2c"
    },
    "/Users/kgartland/Projects/website/v2.vuejs.org/src/v2/examples/vue-20-elastic-header/index.html": {
      "checksum": "109ce2aade0ff42b94df0961b03f452b"
    }
  }
}
```

In all cases I think we want to match the local folder structure between files. So I would expect something like:

```
{
  "version": 1,
  "metadata": {
    "appmode": "static",
    "primary_html": "./index.html"
  },
  "files": {
    "./css/style.css": {
      "checksum": "94264a34917fd9077d384c495e99b77d"
    },
    "./scripts/saved.js": {
      "checksum": "3e2664e064c50a0e8d3ba83081826a2c"
    },
    "./index.html": {
      "checksum": "109ce2aade0ff42b94df0961b03f452b"
    }
  }
}
```

PR #237, opened 2022-03-25
conflicts with existing app mode will now raise an error

### LINKED ISSUE #234: App mode should not be overridden with value from Connect
When specifying `--app-id`, rsconnect-python overrides the app mode with the value from the existing deployment specified by the app ID. https://github.com/rstudio/rsconnect-python/blob/216303f40494dea9655b921619c7df9535c43705/rsconnect/actions.py#L1147-L1153

This seems wrong. The next code block handles the case without `--app-id`. It checks the appmode against the value stored in the local deployment metadata, and raises an error if they do not match. We should do the same check in the app-id case: if the deployment mode doesn't match the existing deployment, raise an error.

bcwu:
Fixes #234

### DIFF
--- a/rsconnect/actions.py
+++ b/rsconnect/actions.py
@@ -1014,7 +1014,11 @@ def gather_basic_deployment_info_for_notebook(connect_server, app_store, file_na
     if new and app_id:
         raise api.RSConnectException("Specify either a new deploy or an app ID but not both.")
 
-    app_mode = AppModes.STATIC
+    if static:
+        app_mode = AppModes.STATIC
+    else:
+        app_mode = AppModes.JUPYTER_NOTEBOOK
+
     existing_app_mode = None
     if app_id is None:
         # Possible redeployment - check for saved metadata.
@@ -1033,11 +1037,6 @@ def gather_basic_deployment_info_for_notebook(connect_server, app_store, file_na
             + "Use the --new option to create a new deployment of the desired type."
         ) % (app_mode.desc(), existing_app_mode.desc())
         raise api.RSConnectException(msg)
-
-    if static:
-        app_mode = AppModes.STATIC
-    else:
-        app_mode = AppModes.JUPYTER_NOTEBOOK
 
     default_title = not bool(title)
     title = title or _default_title(file_name)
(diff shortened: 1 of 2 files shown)


rsconnect/actions.py
```diff
@@ -1014,27 +1014,31 @@ def gather_basic_deployment_info_for_notebook(connect_server, app_store, file_na
...
+        existing_app_mode = AppModes.get_by_ordinal(app.get("app_mode", 0), True)
+    if not new and existing_app_mode and app_mode != existing_app_mode:
+        msg = (
+            "Deploying with mode '%s',\n"
+            + "but the existing deployment has mode '%s'.\n"
+            + "Use the --new option to create a new deployment of the desired type."
+        ) % (app_mode.desc(), existing_app_mode.desc())
+        raise api.RSConnectException(msg)
+
+    if static:
```
mmarchetti, line 1037: It seems like this needs to go before the conflict check above, so we're checking correct value for `app_mode`.

rsconnect/actions.py
```diff
@@ -1144,25 +1148,25 @@ def _gather_basic_deployment_info_for_framework(
...
-    if app_id is not None:
-        # Don't read app metadata if app-id is specified. Instead, we need
-        # to get this from Connect.
-        app = api.get_app_info(connect_server, app_id)
-        app_mode = AppModes.get_by_ordinal(app.get("app_mode", 0), True)
-
-        logger.debug("Using app mode from app %s: %s" % (app_id, app_mode))
-
-    if not new and app_id is None:
+    app_mode = AppModes.STATIC
```
mmarchetti, line 1151: The `app_mode` parameter should contain the correct value for the framework... I'm not sure I understand why this is set to `STATIC`.

mmarchetti (approved):

kgartland-rstudio:
Confirmed on `master` that you can indeed overwrite a target app with a different app-mode when deploying a notebook.  

Validated the fix on this branch and can no longer overwrite another target app with a different appmode.

```
Checking arguments...                            [ERROR]
Error: Deploying with mode 'Jupyter Notebook',
but the existing deployment has mode 'Streamlit Application'.
Use the --new option to create a new deployment of the desired type.
```

- [x] error when using a target -app-id  with a different appmode
- [x] error when using an rsconnect-python json file from an app with a different appmode
- [x] error when deploying a manifest with a different app mode

PR #244, opened 2022-05-04
refactor

bcwu:
### Description

Refactor various parts of the code to make the programmatic side easier to understand and more extensible.

#### deploy
- [ ] api
  - [ ] --new
  - [ ] re-deploy
- [ ] bokeh
- [ ] dash
- [ ] fastapi
- [ ] html
- [ ] manifest
- [ ] notebook
- [ ] other-content
- [ ] quarto
- [ ] streamlit
#### errors
- [ ] redeploy with different mode should raise error


rsconnect/api.py
```diff
@@ -268,6 +268,312 @@ def output_task_log(task_status, last_status, log_callback):
         return new_last_status
 
 
+class RSConnectExecutor:
+    def __init__(self, *args, **kwargs) -> None:
```
mmarchetti, line 272: To make this API more usable and support tooling, we need named arguments with types here. Rather than having lots of arguments as the functions in actions.py currently do, we can group them as we have done in main.py with the `@server_args` and `@content_args` decorators. If those decorators each consumed multiple CLI arguments and bundled them into an object (perhaps a dataclass, which is usable as far back as Python 3.6), then the CLI functions would accept a `server` argument and/or a `content` argument which could be passed through to this constructor.

rsconnect/api.py
```diff
@@ -268,6 +268,312 @@ def output_task_log(task_status, last_status, log_callback):
...
+        print(kwargs)
+        self.reset()
+        self._d = kwargs
```
mmarchetti, line 275: In general, this class should try to avoid pulling values from a dictionary but instead have its own named/typed attributes, again for tooling support. That would enable typechecking and prevent bugs that can creep in through dict key typos.

rsconnect/api.py
```diff
@@ -268,6 +268,312 @@ def output_task_log(task_status, last_status, log_callback):
...
+    def state(self):
+        return self._d
+
+    def get(self, *args, **kwargs):
+        key = args[0]
+        return kwargs.get(key) or self.state.get(key)
+
+    def pipe(self, func, *args, **kwargs):
+        func(*args, **kwargs)
+        return self
```
mmarchetti, line 307: Something feels a little off about this method. If I'm in the middle of a set of chained method calls, and I need to call an unrelated function, I feel like it's better to break the chain and call the function.

If the intent is to allow inspection of the executor state mid-stream, then the called function should probably receive `self` as an argument.

rsconnect/api.py
```diff
@@ -268,6 +268,312 @@ def output_task_log(task_status, last_status, log_callback):
...
+        # If our info came from the command line, make sure the URL really works.
+        if not from_store:
+            connect_server, _ = connect_server.test_server()
+
+        connect_server.api_key = api_key
+
+        if not connect_server.api_key:
+            if api_key_is_required:
+                raise RSConnectException('An API key must be specified for "%s".' % connect_server.url)
+            return connect_server
```
mmarchetti, line 354: Should this return `self`?

rsconnect/bundle.py
```diff
@@ -642,7 +642,9 @@ def make_html_bundle_content(
     return manifest, relevant_files
 
 
-def infer_entrypoint(path, mimetype):
+def infer_entrypoint(*args, **kwargs):
```
mmarchetti, line 645: Where we have specific, known arguments, we should keep those (and add types) rather than using * or **.

rsconnect/bundle.py
```diff
@@ -664,12 +666,7 @@ def infer_entrypoint(path, mimetype):
     return mimetype_filelist[mimetype].pop() if len(mimetype_filelist[mimetype]) == 1 else None
 
 
-def make_html_bundle(
-    path,  # type: str
-    entry_point,  # type: str
-    extra_files=None,  # type: typing.Optional[typing.List[str]]
-    excludes=None,  # type: typing.Optional[typing.List[str]]
-):
+def make_html_bundle(*args, **kwargs):
```
mmarchetti, line 669: same here

rsconnect/main.py
```diff
@@ -980,59 +979,13 @@ def deploy_quarto(
...
-        app_store,
-        path,
-        app_id,
-        app_mode,
-        deployment_name,
-        title,
-        default_title,
-        bundle,
-        env_vars,
+def deploy_html(*args, **kwargs):
```
mmarchetti, line 982: If we adopt the pattern mentioned above, this would have a signature like

```
def deploy_html(server: ServerArgs, content: ContentArgs, files: FileArgs):
```
where ServerArgs, ContentArgs, and FileArgs are dataclasses that define the parameters (and their types) passed on the command line. We can then pass them through to the executor.

mmarchetti (changes requested):

rsconnect/api.py
bcwu: This is an interesting design decision to think through. The design so far is intended to push the complexity down. That comes with some advantages and disadvantages. The advantages being that the user can provide a grab bag of keyword arguments and we'll setup the Connect backend target correctly, or warn the user that something is missing. 

Let's say that we switched over to use another `RSConnectServer`, but also have to handle backwards compatibility; instead of having the user (or the `@server_args` in this case) worry about constructing the correct version of `RSConnectServer`, we just seamlessly handle it within the Executor.

This is probably more evident in how we are refactoring to encapsulate `RSConnect(HTTPServer)`. If we need to switch `RSConnect(HTTPServer)` to another backend target like connect-cloud, the user can just pass another keyword argument instead of passing us a fully constructed http server.

rsconnect/log.py
```diff
@@ -120,3 +120,68 @@ def setLevel(self, level):
...
+        return wrapper
+
+    return decorator
+
+
+console_logged = partial(logged, console_logger)
+
+
+# generic logger
+connect_logger = logging.getLogger("connect_logger")
```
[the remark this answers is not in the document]
mmarchetti: Actually, I'm a little confused as to why we have two loggers and are doing this configuration here.

Idiomatically, Python libraries get their named logger (`getLogger('rsconnect')`) and log everything to that without configuring it. The rsconnect CLI (in main.py) can call `getLogger('rsconnect')` and configure it for console logging as you have done above. Other library clients can configure the logger as they prefer. We should only need a single logger.

rsconnect/main.py
```diff
@@ -1038,45 +963,141 @@ def deploy_html(
...
+        environment = create_python_environment(
+            directory,
+            extra_files,
+            force_generate,
+            python,
+            conda,
+        )
+
+        rsce = api.RSConnectExecutor(**kwargs)
+        (
```
mmarchetti, line 1081: The parentheses and indentation aren't needed here since you're including `rsce` on the line below.

rsconnect/main.py
```diff
@@ -1384,7 +1410,7 @@ def write_manifest_notebook(
     verbose,
     file,
     extra_files,
-    image=None,  # type: str
+    image,
```
mmarchetti, line 1413: Maybe need to merge latest master? `image` is now optional.

mmarchetti (changes requested):
I know you're still working on things, but I wanted to take a pass back through the whole PR since there have been a ton of changes since my last review.

rsconnect/main.py
bcwu: The parentheses are for omitting line breaks. Removing them will mean every line will have to add '\'

rsconnect/api.py
```diff
@@ -268,6 +268,403 @@ def output_task_log(task_status, last_status, log_callback):
...
+        insecure: bool = False,
+        cacert: IO = None,
+        cookies=None,
+        timeout: int = 30,
+        **kwargs
+    ) -> None:
+        self.reset()
+        self._d = kwargs
+        self.setup_connect_server(name, url or kwargs.get("server"), api_key, insecure, cacert)
+        self.setup_connect(cookies, timeout)
```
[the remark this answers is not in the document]
bcwu: `connect_server` is the server information, `connect` is the client. I've renamed `connect` to client.

rsconnect/api.py
```diff
@@ -268,6 +268,403 @@ def output_task_log(task_status, last_status, log_callback):
...
+        api_key: str = None,
+        insecure: bool = False,
+        cacert=None,
+    ):
+        if name and url:
+            raise RSConnectException("You must specify only one of -n/--name or -s/--server, not both.")
+
+        url, api_key, insecure, cacert, _ = ServerStore().resolve(name, url, api_key, insecure, cacert)
+        self.connect_server = RSConnectServer(url, api_key, insecure, cacert)
+        self._name = name
```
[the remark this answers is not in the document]
bcwu: Name is resolved at run time so storing it may lead to subtle bugs. Will omit it altogether since it doesn't serve any large purpose.

rsconnect/api.py
```diff
@@ -268,6 +268,403 @@ def output_task_log(task_status, last_status, log_callback):
...
+
+        return self
+
+    @console_logged("Making bundle ...")
+    def make_bundle(self, func, *args, **kwargs):
+        path = (
+            self.get("path", **kwargs)
+            or self.get("file", **kwargs)
+            or self.get("file_name", **kwargs)
+            or self.get("directory", **kwargs)
```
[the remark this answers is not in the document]
bcwu: This is caused by the different upstream bundle functions that take either a file or a folder. Unification needs to happen there first before it can be simplified here.

rsconnect/api.py
```diff
@@ -268,6 +268,403 @@ def output_task_log(task_status, last_status, log_callback):
...
+        url: str = None,
+        api_key: str = None,
+        insecure: bool = False,
+        cacert: IO = None,
+        cookies=None,
+        timeout: int = 30,
+        **kwargs
+    ) -> None:
+        self.reset()
+        self._d = kwargs
```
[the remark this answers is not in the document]
bcwu: Main issues with fixed attributes is 1) that modification will immediately render past versions incompatible and 2) combinatorial explosion from new content. For example, if a new type of content is added that requires a new parameter, then the developer will also need to add that attribute here. The complexity is cascaded down and multiplied. 

What the current design is going for is to store the minimum information to communicate with the backend client, with a flexible state that can accommodation future content growth.

kgartland-rstudio:
## QA Testing Notes (Server + Deploy Testing):

### Server Testing:

- [ ] add server
```
rsconnect add \
    --api-key my-api-key \
    --server https://connect.example.org:3939 \
    --name myserver
```

- [ ] deploy with named server
```
rsconnect deploy notebook --name myserver my-notebook.ipynb
```

- [ ] deploy when only a single server is samed:
```
rsconnect deploy notebook my-notebook.ipynb
```
- [ ] list servers
```
rsconnect list
```

- [ ] remove server
```
rsconnect remove --name myserver
```

- [ ] server details
```
rsconnect details --server https://connect.example.org:3939
```

- [ ] server details w/ api key
```
rsconnect details --server https://connect.example.org:3939 --api-key my-api-key
```


### Notebook Deployment Options:
- [ ] extra files
```
rsconnect deploy notebook my-notebook.ipynb data.csv
```

- [ ] requirements.txt file is honored when present
- [ ] force-generate option overwrites existing requirements.txt file
```
rsconnect deploy notebook my-notebook.ipynb --force-generate
```

- [ ] specify python path
```
rsconnect deploy notebook --python /path/to/python my-notebook.ipynb
```
- [ ] static deploys work
```
rsconnect deploy notebook --static my-notebook.ipynb
```

### API/Application Deployment Options
- [ ] Include extra files
```
rsconnect deploy api flask-api/ data.csv
```

- [ ] Exclude files in directory
```
rsconnect deploy dash --exclude dash-app-venv --exclude TODO.txt dash-app/
```

- [ ] Exclude entire directories
```
rsconnect deploy dash --exclude dash-app-venv --exclude output/ dash-app/
```

- [ ] Exclude multiple times with pattern matching
```
rsconnect deploy dash --exclude dash-app-venv --exclude “*.txt” dash-app/
```

- [ ] Including extra file takes precedence over exclude
```
rsconnect deploy dash --exclude “*.csv” dash-app/ important_data.csv
```

- [ ] Exclude specifically named directories by default
```
.Rproj.user
.env
.git
.svn
.venv
__pycache__
env
packrat
renv
rsconnect-python
rsconnect
venv
```

### Deploying R or Other Content

- [ ] deploy manifest for R (other) content
```
rsconnect deploy manifest /path/to/manifest.json
```

### Options for All Types of Deployments
- [ ] title
```
rsconnect deploy notebook --title "My Notebook" my-notebook.ipynb
```

- [ ] environment variables
```
rsconnect deploy notebook --environment MYVAR='hello world' notebook.ipynb
```
```
export SECRET_KEY=12345

rsconnect deploy notebook --environment SECRET_KEY notebook.ipynb
```

### Network Options
- [ ] provide ca cert
```
rsconnect details \
    --api-key my-api-key \
    --server https://connect.example.org:3939 \
    --cacert /path/to/certificate.pem
```

- [ ] use insecure mode
```
rsconnect add \
    --api-key my-api-key \
    --server https://connect.example.org:3939 \
    --insecure
```

### Updating a Deployment
- [ ] force new
```
rsconnect deploy dash --new my-app/
```

- [ ] update different deployment
-- [ ] as owner
-- [ ] as collaborator
-- [ ] as viewer (should fail)
-- [ ] with no access (should fail)
```
rsconnect deploy notebook --app-id 123456 my-notebook.ipynb
```

### Show Deployment Info
```
rsconnect info my-notebook.ipynb
```

### Hide Jupyter Notebook Input Code Cells
- [ ] hide all
```
rsconnect deploy notebook \
    --server https://connect.example.org:3939 \
    --api-key my-api-key \
    --hide-all-input \
    my-notebook.ipynb
```

- [ ] hide tagged cells
```
rsconnect deploy notebook \
    --server https://connect.example.org:3939 \
    --api-key my-api-key \
    --hide-tagged-input \
    my-notebook.ipynb
```

### Other options
- [ ] entry-point (deploy app with a different entrypoint than `app`)
- [ ] --verbose (-v)
- [ ] image
- [ ] --help
- [ ] version

kgartland-rstudio:
capturing testing issues here:
https://github.com/rstudio/rsconnect-python/issues/256

rsconnect/api.py
```diff
@@ -268,6 +269,452 @@ def output_task_log(task_status, last_status, log_callback):
...
+        )
+        self.connect_server.handle_bad_response(log_lines)
+        app_config = self.client.app_config(app_id)
+        self.connect_server.handle_bad_response(app_config)
+        app_dashboard_url = app_config.get("config_url")
+        log_callback.info("Deployment completed successfully.")
+        log_callback.info("\t Dashboard content URL: ")
+        log_callback.debug(app_dashboard_url)
+        log_callback.info("\t Direct content URL: ")
+        log_callback.debug(self.state["deployed_info"]["app_url"])
```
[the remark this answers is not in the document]
bcwu: This is comes back to color outputs. Log levels having different colors. Info is white whereas debug is green. Error is red.

mmarchetti (approved):
There are some improvements that could be made, but I don't think this PR is the place to make them at this point.

PR #250, opened 2022-05-26
revert parameters to optional for rsc-jupyter backward compatibility

bcwu:
### Description

Revert some bundle function parameters back to optional for rsc-jupyter backward compatibility


mmarchetti (changes requested):
The pre-existing `hide_all_input` and `hide_tagged_input` parameters do not need default values, they can be required arguments. Thaat's more consistent with how we treat the other existing arguments. The new `image` argument does need a default so rsconnect-jupyter can omit it.

PR #260, opened 2022-06-28
Switch docs from AWS id/secret to a role

mmarchetti:
### Description

Switch the docs upload from using an AWS ID and secret (both repo secrets) to using an assumed role (also a repo secret).

This will be exercised when we tag the beta of the next release.


limasilet (approved):
Looks good to me 👍

bcwu (approved):

PR #261, opened 2022-07-01
support for shinyapps.io

mslynch:
### Description

Adds support for adding shinyapps.io credentials and deploying to shinyapps.io.


rsconnect/api.py
```diff
@@ -42,21 +43,63 @@ def handle_bad_response(self, response):
...
+
+
+TargetableServer = typing.Union[ShinyappsServer, RSConnectServer]
+
+
+class S3Server(AbstractRemoteServer):
+    remote_name = "S3"
+
+    def __init__(self, url: str):
+        self.url = url
```
ssinnott, line 95: Is there a reason `remote_name` is handled differently in this case compared to the above two. I would have expected...

```
    def __init__(self, url):
        super().__init__(url, "S3")
```

rsconnect/api.py
```diff
@@ -312,28 +368,57 @@ def drop_context(self):
...
+            token = server_data.token
+            secret = server_data.secret
+        self.is_server_from_store = server_data.from_store
+
+        if api_key:
+            self.remote_server = RSConnectServer(url, api_key, insecure, ca_data)
+        elif token and secret:
+            self.remote_server = ShinyappsServer(url, account, token, secret)
+        else:
+            self.remote_server = ShinyappsServer(url, account, token, secret)
```
ssinnott, line 413: This seems a little odd. If it's possible to pass in no arguments when creating a server (api_key, token, secret) why would it end up create a shinyapps server. 

I would have expected either a validation error - or a `RSConnectServer` for backwards compatibility?

rsconnect/api.py
```diff
@@ -406,14 +509,30 @@ def validate_server(
...
 
-        self.connect_server = connect_server
-        self.client = RSConnect(self.connect_server)
+        self.remote_server = connect_server
+        self.client = RSConnectClient(self.remote_server)
 
         return self
 
+    def validate_shinyapps_server(
+        self, url: str = None, account_name: str = None, token: str = None, secret: str = None, **kwargs
```
ssinnott, line 521: The format for the arguments here is different from the above - is black supposed to reformat these?

rsconnect/api.py
```diff
@@ -713,6 +887,179 @@ def filter_out_server_info(**kwargs):
...
+        counter = 1
+        status = None
+
+        while counter < timeout and status not in ["success", "failed", "error"]:
+            task = self.get_task(task_id)
+            self._server.handle_bad_response(task)
+            status = task.json_data["status"]
+            description = task.json_data["description"]
+
+            print("Waiting: {} - {}".format(status, description))
```
ssinnott, line 998: I would include the task id here - which would make debugging failures easier.

`print("Waiting on task {}:  {} - {}".format(task_id, status, description))`

rsconnect/api.py
```diff
@@ -713,6 +887,179 @@ def filter_out_server_info(**kwargs):
...
+    def deploy_application(self, bundle_id, app_id):
+        return self.post("/v1/applications/{}/deploy".format(app_id), body={"bundle": bundle_id, "rebuild": False})
+
+    def get_task(self, task_id):
+        return self.get("/v1/tasks/{}".format(task_id), query_params={"legacy": "true"})
+
+    def get_current_user(self):
+        return self.get("/v1/users/me")
+
+    def wait_until_task_is_successful(self, task_id, timeout=60):
```
ssinnott, line 988: Builds can sometimes take quite a bit so I would probably up this...

```suggestion
    def wait_until_task_is_successful(self, task_id, timeout=180):
```

rsconnect/api.py
```diff
@@ -485,17 +608,61 @@ def deploy_bundle(
...
+            ) as s3_client:
+                upload_result = s3_client.upload(
+                    "{}?{}".format(parsed_upload_url.path, parsed_upload_url.query),
+                    prepare_deploy_result.presigned_checksum,
+                    bundle_size,
+                    contents,
+                )
+                S3Server(upload_url).handle_bad_response(upload_result)
+
+            self.client.do_deploy(prepare_deploy_result.bundle_id, prepare_deploy_result.app_id)
```
ssinnott, line 655: I don't know the best place to inject this. But the task that does the deploy can fail. In that case it probably doesn't make sense to send the user to the `app_url` (because the app_url would either not work or not be the version of the app they think they just uploaded).

So I might check for this in `wait_until_task_is_successful`... and print and error and exception? I think that's what the connect side does (https://github.com/rstudio/rsconnect-python/blob/master/rsconnect/api.py#L45-L52).

rsconnect/api.py
mslynch: Nope, that was just an oversight.

rsconnect/api.py
mslynch: This is how black formatted it. I'm guessing it tries to fit the parameters all onto a single line, but if that's too long, it would break them up onto separate lines.

rsconnect/api.py
bcwu: per Sean's comment, this should look as follows:
https://github.com/rstudio/rsconnect-python/blob/3e4b0a7ab31ba41cc5b070522be0cc4bfd2efc49/rsconnect/api.py#L410-L416

rsconnect/api.py
```diff
@@ -587,13 +755,17 @@ def validate_app_mode(self, *args, **kwargs):
...
-                # to get this from Connect.
-                app = get_app_info(connect_server, app_id)
-                existing_app_mode = AppModes.get_by_ordinal(app.get("app_mode", 0), True)
+                # to get this from the remote.
+                if isinstance(self.remote_server, RSConnectServer):
+                    app = get_app_info(self.remote_server, app_id)
+                    existing_app_mode = AppModes.get_by_ordinal(app.get("app_mode", 0), True)
+                else:
+                    app = get_shinyapp_info(self.remote_server, app_id)
+                    existing_app_mode = AppModes.get_by_cloud_name(app.json_data["mode"])
```
bcwu, line 768: the pattern we should follow now that we are accepting different backends is to have a default raise exception when we don't get a match, example as follows. This also allows for future expansion without needing to modify the else statement:
https://github.com/rstudio/rsconnect-python/blob/3e4b0a7ab31ba41cc5b070522be0cc4bfd2efc49/rsconnect/api.py#L411-L416

rsconnect/actions.py
```diff
@@ -1514,13 +1523,17 @@ def _gather_basic_deployment_info_for_framework(
...
             # Don't read app metadata if app-id is specified. Instead, we need
             # to get this from Connect.
-            app = api.get_app_info(connect_server, app_id)
-            existing_app_mode = AppModes.get_by_ordinal(app.get("app_mode", 0), True)
+            if isinstance(remote_server, api.RSConnectServer):
+                app = api.get_app_info(remote_server, app_id)
+                existing_app_mode = AppModes.get_by_ordinal(app.get("app_mode", 0), True)
+            else:
+                app = api.get_shinyapp_info(remote_server, app_id)
+                existing_app_mode = AppModes.get_by_cloud_name(app.json_data["mode"])
```
bcwu, line 1536: default raise exception pattern here:
https://github.com/rstudio/rsconnect-python/blob/3e4b0a7ab31ba41cc5b070522be0cc4bfd2efc49/rsconnect/api.py#L411-L416

rsconnect/api.py
mslynch: Updated to handle error/failed status as an exception.

ssinnott (approved):
There's some broken tests after the merge... but the script works for deploying to shinyapps.io locally on my end. So I think it's good to go from my perspective.

rsconnect/main.py
```diff
@@ -724,6 +728,26 @@ def deploy_notebook(
...
 @server_args
 @content_args
+@click.option(
+    "--account",
+    "-a",
+    envvar="SHINYAPPS_ACCOUNT",
+    help="The shinyapps.io account name.",
+)
+@server_args
+@content_args
```
bcwu, line 738: @server_args and @content_args are duplicated here.

rsconnect/main.py
```diff
@@ -276,27 +280,72 @@ def _test_server_and_api(server, api_key, insecure, ca_cert):
...
+    "-T",
+    envvar="SHINYAPPS_TOKEN",
+    help="The shinyapps.io token.",
+)
+@click.option(
+    "--secret",
+    "-S",
+    envvar="SHINYAPPS_SECRET",
+    help="The shinyapps.io token secret.",
+)
```
bcwu, line 300: these new options could be all added under @server_args https://github.com/rstudio/rsconnect-python/blob/72c58f47675c7db84632a0688c49400ca2ea73dc/rsconnect/main.py#L100-L133

mslynch: The only method that will work with shinyapps.io right now is `deploy_manifest`, so I think it would be misleading for rsconnect-python to explicitly accept those options for the other methods.

rsconnect/validation.py
```diff
@@ -0,0 +1,47 @@
...
+
+from rsconnect.exception import RSConnectException
+
+
+def _get_present_options(options: typing.Dict[str, typing.Optional[str]]) -> typing.List[str]:
+    return [k for k, v in options.items() if v]
+
+
+def validate_new_server_options(url, api_key, insecure, cacert, account, token, secret):
+    validate_connection_options(None, url, api_key, insecure, cacert, account, token, secret)
```
bcwu, line 11: In general, pass-through methods create extra indirection that we have to reason through, so simplifying them can help a developer save time. In this case, we can make the underlying method use optional keyword arguments, thereby removing the need of a pass-through method.

bcwu (approved):

PR #275, opened 2022-08-03
Zack jwt

zackverham:
### Description

>
> context.

Connected to #

>


zackverham:
Noted this in the connect-admin channel, but posting it here as well for visibility / to set expectations:

QA for this is sort of by-definition tightly coupled with the Connect changes - so QA isn’t fully possible yet. But I think its ready to begin collecting iterative feedback.

gbrlsnchs:
One comment I have after using the command is that perhaps we might want to output the response terminated by a newline. WDYT?

rsconnect/api.py
```diff
@@ -81,9 +81,10 @@ class RSConnectServer(AbstractRemoteServer):
     instance of the Connect server.
     """
 
-    def __init__(self, url, api_key, insecure=False, ca_data=None):
+    def __init__(self, url, api_key, jwt=None, insecure=False, ca_data=None):
```
bcwu, line 84: A good practice is to always add new kwargs to the end of the parameters, because some users might be passing  args in the original order. Adding the jwt kwarg here will break existing behavior.

rsconnect/main.py
```diff
@@ -278,6 +286,80 @@ def _test_shinyappsio_creds(server: api.ShinyappsServer):
...
+    is_flag=True,
+    help="Disable TLS certification/host validation.",
+)
+@click.option(
+    "--cacert",
+    "-c",
+    envvar="CONNECT_CA_CERTIFICATE",
+    type=click.File(),
+    help="The path to trusted TLS CA certificates.",
+)
```
bcwu, line 312: use server_args instead

zackverham: I decided not to use server_args because I didn't want to expose --api-key or --name: 

--api-key because it isn't a valid parameter, --name because the server storage components rely on api_key for the connection.

dethmasque:
Was pokin around and noticed we tell users to look at `rsconnect --server` which does not exist:

```
MacBook-Pro:rsconnect-python questionsleep$ rsconnect bootstrap -s https://rsc.radixu.com -j ./keytest.pem
{
  "status": 404,
  "api_key": "",
  "message": "Unable to find provisioning endpoint. Please check the 'rsconnect --server' parameter and your Connect configuration."
}
```

```
MacBook-Pro:rsconnect-python questionsleep$ rsconnect --server
Usage: rsconnect [OPTIONS] COMMAND [ARGS]...
Try 'rsconnect --help' for help.

Error: No such option: --server
```

zackverham:
@dethmasque great catch - I'll improve the language there

marcosnav (approved):
LGTM

requirements.txt
```diff
@@ -10,6 +10,8 @@ ipython
 jupyter_client
 mypy; python_version >= '3.6'
 nbconvert
+pyjwt>=2.4.0; python_version >= '3.6'
```
zackverham, line 13: I don't love this conditional dependency here, but if we don't have a version of `pyjwt` installed for older versions of python, we (unsurprisingly) get import errors. 

The feature is only supported for python > 3.6, so that the pyjwt dependency in older versions is really a placeholder. 

If somehow the user was able to run the feature using python 3.5, they would encounter failures because some of the implementation details in pyjwt changed between version ~1.7 (the last supported version for 3.5) and the 2.4.0/2.5.0 pyjwt versions. But this should be a non-issue because we only support this feature for python >= 3.6.

Just providing some context here - if someone has a cleaner solution I'd be happy to implement it.

bcwu (approved):

kgartland-rstudio:
I started looking at the jwt branch this morning and it looks like if you don't have jwt installed locally all commands fail with the following error:

```
> rsconnect deploy --help
  File "/Users/kgartland/.pyenv/versions/rsconnect-tag/bin/rsconnect", line 5, in <module>
    from rsconnect.main import cli
  File "/Users/kgartland/.pyenv/versions/3.8.2/envs/rsconnect-tag/lib/python3.8/site-packages/rsconnect/main.py", line 71, in <module>
    from .json_web_token import (
  File "/Users/kgartland/.pyenv/versions/3.8.2/envs/rsconnect-tag/lib/python3.8/site-packages/rsconnect/json_web_token.py", line 11, in <module>
    import jwt
ModuleNotFoundError: No module named 'jwt'
```

Should the jwt dependency come with rsconnect-python?  I'm not sure how that works when installing via git.

kgartland-rstudio:
- [x] [Testlodge](https://app.testlodge.com/a/18880/projects/32628/test_plans/56366) regression tests pass
- [x] rsconnect-python installed with python 3.5
- [x] rsconnect-python deploy with python 3.5
- [ ] rsconnect-python tests [running...](https://github.com/rstudio/connect/actions/runs/3315975407)

noxferatu:
JWT is generating, bootstrap commands are functioning as expected, running against dev environment with the JWT-enabled endpoint.
Integration test script looks good.

PR #284, opened 2022-09-21
add CLI command for git repo deployment

colearendt:
### Description

Add options to deploy a git repository to Connect. Some caveats:

- We should evaluate this UX to decide if we want a related mechanism for "telling a piece of content to fetch new git commits"
- Need to add testing and ensure all of the appropriate bits are copied over
- It has been a while since I wrote this. I may be missing things
- There are internal API endpoints in use here
- The interface can be evaluated. IIRC it is currently `rsconnect deploy git`
- This was written before the "great refactor" and may need to be tweaked as a result

Connected to #

### DIFF
--- a/rsconnect/api.py
+++ b/rsconnect/api.py
@@ -189,6 +189,29 @@ class RSConnectClient(HTTPServer):
         self._server.handle_bad_response(response)
         return response
 
+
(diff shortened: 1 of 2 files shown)


mmarchetti:
Closing in favor of #501, which includes this change.

PR #293, opened 2022-10-13
option for adding vetiver testing

isabelizimm:
### Description

>
> context.

Wanted to highlight out a few places where it might be convenient to test vetiver <> rsconnect-python. Essentially this code will run a Docker container with Connect configured inside of it, and test deploying a model with vetiver into Connect. This will check against the latest released version of vetiver, which I think is sufficient for testing these two packages' interactions.

If this seems useful to implement, or take part of and implement, feel free! If this doesn't align with how you all are testing now, no worries if this is not a useful approach for this group.

(still working on the Docker container, will update with running version if this interests you!)

CC: @juliasilge 

>


isabelizimm:
Okay, after playing around with this, this should be nearly working (needs an RSC_LICENSE env variable set to actually set up the Connect instance). It is as self contained as I think I can make it. Is this a style of testing that would be optimal for this repo to test with vetiver's deployment?

Issue #285, opened 2022-09-21
rsconnect write-manifest not excluding files on windows


shapiromatron:
When running `rsconnect write-manifest`, the generated manifest.json still includes files in the `venv` and `.git` folder, even though they're supposed to be ignored by default, per the readme documentation.  

I tried excluding and I still cannot get them excluded. I suspect this may be because I'm running this command on Windows 10.  I'm using `rsconnect-python==1.10.0`

Here the command I'm using:

```batch
rsconnect write-manifest dash -o -e app:app .
```

The manifest is very large (5mb) because of these directories being included in the files section.

shapiromatron:
I ended up creating a `clean_manifest.py` script that I can whenever generating an updated manifest:

```python
import json
from pathlib import Path


HERE = Path(__file__).parent


def main():
    """
    Clean the manifest does the following:

    1. Change locale setting to utf8
    2. Replace windows paths \\ with /
    3. Remove files in .git and venv

    An issue has been reported to fix the third item:
    https://github.com/rstudio/rsconnect-python/issues/285

    """
    fn = HERE / 'manifest.json'
    data = json.loads(fn.read_text())

    data['locale'] = 'en_US.utf8'
    new_files = {}
    for key, value in data['files'].items():
        if key.startswith('.git') or key.startswith('venv'):
            continue
        new_files[key.replace('\\', '/')] = value
    data['files'] = new_files

    fn.write_text(json.dumps(data, indent=2))


if __name__ == "__main__":
    main()

```

hcoohb:
None of the folders that should be excluded by default actually are on windows. (`.git`, `__pycache__`,  ...)
And the glob pattern for file exclusion does not seem to work either on windows...
This is very frustrating.

Any recommendation on how to deal with that?

shapiromatron:
The "hotfix" is running the script above after generating the manifest, and before deployment. The correct way would probably be to not check paths in an OS-specific way (fwd vs backwards slash, using pathlib.Path for example) - see https://github.com/rstudio/rsconnect-python/blob/f595f69d0f943949bcc77f9f25710791ddadd454/rsconnect/bundle.py#L36-L49

PR #294, opened 2022-10-18
make exclude environment files work on multiple OS

bcwu:
### Description

Fixes https://github.com/rstudio/rsconnect-python/issues/285

discovered that --exclude glob patterns do not work on Windows, fix is outside of the scope of this PR but issue is logged as follows:
https://github.com/rstudio/rsconnect-python/issues/320


mmarchetti (approved):

kgartland-rstudio:
Verified that we now exclude proper folders like `.git` by default on Windows, but the `--exclude` flag still doesn't work properly.

```
> tree
C:.
|--.git
|--ignore_me
```

`rsconnect write-manifest dash ./ --exclude ignore_me`

yet `ignore_me` is still in the file list in the manifest while `.git` is not.

kgartland-rstudio:
I'm still having some issues with this.

test case:
single sub-folder with a file in it:
```ignore-folder/sub-file.txt```

command: 
```rsconnect write-manifest dash ./ --exclude ignore-folder```

The manifest created contains the `ignore-folder`:
Screen Shot 2022-11-28 at 12 14 33 PM

The following command works in this scenario:
```rsconnect write-manifest dash ./ --exclude ignore-folder/*```

test case:
single sub-folder with an additional sub-folder and file in it:
```ignore-folder/subfile.txt```

command:
```rsconnect write-manifest dash ./ --exclude ignore-folder/*```

errors with:
Screen Shot 2022-11-28 at 12 50 20 PM

**Comment by kgartland-rstudio** [comment]
Confirmed that using the `--exclude` flag works on windows now with both files and folders with and without subdirectories.

It also ignores folders and files that match our exclusion list:

```
.Rproj.user
.env
.git
.svn
.venv
__pycache__
env
packrat
renv
rsconnect-python
rsconnect
venv
```

Issue #269, opened 2022-07-19
Connect not installing python github packages


isabelizimm:
When using `vetiver` alongside `rsconnect-python`, things seem to break when trying to use a github version of a package that includes functions not available in previous versions. In particular, Connect will only install PyPI versions of packages, even if that is not what is in someone's environment. 
It gives some [cryptic errors](https://rstudio.slack.com/archives/C021090K8Q5/p1658242583551209). The only way to resolve this is to do a `vetiver` release with the new or updated feature.

isabelizimm:
It looks like when `rsconnect-python` creates a requirements.txt, it does so with `pip list --format=freeze`

https://github.com/rstudio/rsconnect-python/blob/f595f69d0f943949bcc77f9f25710791ddadd454/rsconnect/environment.py#L221

so a GitHub installation will be saved as `vetiver==0.1.7.post41+g60e590a`, which ca. Using `pip freeze` instead will save the requirements to something similar to `vetiver @ git+https://github.com/rstudio/vetiver-python@60e590abae1ce688f3458f30bafc453256920984`, which should allow a GH installation of a package

isabelizimm:
From what I am seeing on my local Connect, a requirements.txt is generated every time, which will overwrite any file I am passing with the name `requirements.txt`. 

This is not related to the pip list vs. pip freeze discussion above, but is relevant since this is blocking users from supplying their own requirements (ie, passing in github installs)

PR #297, opened 2022-10-21
change pip list to pip freeze

isabelizimm:
### Description

>
> context.

Connected to #269 
Would unblock https://github.com/rstudio/vetiver-python/pull/119

>


mmarchetti:
There is a reason we use pip list --format=freeze instead of pip freeze, but it is subtle and probably not well documented. 

pip freeze preserves some build metadata (IIRC it is specified in PEP 518). When you run pip freeze in a conda environment, it includes build-system paths for all the packages. pip installing with those present does not work because of course the paths don't exist on the target system.  Connect strips them out if present, but we preferred to avoid sending them. pip list excludes those paths, even when format=freeze. That may change eventually.

isabelizimm:
Thank you for the explanation, that all makes sense! 🙌 The way I encounter this the most is when I use a github version of a package (via `rsconnect.actions.deploy_python_fastapi`), and pip list generates a version of a package that is not-downloadable from PyPI, even though it looks like it should be (`vetiver==0.1.7.post41+g60e590a`), so I encounter an error on Connect. When I pass in my own `requirements.txt`, it is overwritten by the pip list one that is generated. I can pass in a requirements file named something else, but Connect will still install the `requirements.txt`, which is a totally reasonable guess, just not helpful in this case.

Is there a way I am missing to accomplish installing GH packages?

mmarchetti:
We discussed at our kickoff this morning. You can create requirements.txt and Connect will use it, but we are also considering switching to "pip freeze" as that will improve the automatic capture.

isabelizimm:
Oh awesome! That would be great (selfishly 😄) if pip freeze was used.

> You can create requirements.txt and Connect will use it

From what I am seeing, a requirements.txt is generated every time, which will overwrite any file I am passing with the name `requirements.txt`, but if any other file name will be ignored in favor of installing from the generated `requirements.txt` file. Maybe this should be a separate issue?

isabelizimm:
duplicate of #309

PR #299, opened 2022-10-27
Use directory in `get_python_env_info`

isabelizimm:
### Description

>
> context.

Connected to #269 

`actions._deploy_by_python_framework` ingests a directory, and the directory is passed into `actions.get_python_env_info` in the `filename` argument. t is assumed to be , causing `actions.get_python_env_info` to look one step too high in the directory structure.

This manifests itself when users pass in a `requirements.txt` file when deploying different python frameworks. Because `actions.get_python_env_info` is looking too high in the directory, it does not recognize the user's requirements file in the working directory and will overwrite it with one generated by rsconnect-python.

>

Tried this on my local Connect, and it allows for me to pass in custom `requirements.txt`. Trying it out on the CI to make sure its not breaking other things 😄

### DIFF
--- a/rsconnect/actions.py
+++ b/rsconnect/actions.py
@@ -1632,7 +1632,7 @@ def get_python_env_info(file_name, python, conda_mode=False, force_generate=Fals
     """
     python = which_python(python)
     logger.debug("Python: %s" % python)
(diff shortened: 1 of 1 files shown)


bcwu:
Fixed by https://github.com/rstudio/rsconnect-python/pull/300

PR #300, opened 2022-10-28
update actions

bcwu:
### Description

Update actions.py to reuse the code in main, minus the CLI components.


rsconnect/actions.py
```diff
@@ -693,91 +700,63 @@ def deploy_jupyter_notebook(
...
+                secret=connect_server.secret,
+            )
+        )
+    else:
+        raise RSConnectException("Unable to infer Connect client.")
 
+    base_dir = dirname(file_name)
+    _warn_on_ignored_manifest(base_dir)
+    _warn_if_no_requirements_file(base_dir)
+    _warn_if_environment_directory(base_dir)
```
mmarchetti, line 732: It seems a little odd for API usage to trigger console warnings. We should ask about whether that's a desired behavior.

rsconnect/actions.py
```diff
@@ -1246,401 +1139,47 @@ def deploy_by_manifest(
...
-        raise RSConnectException("Specify either a new deploy or an app ID but not both.")
-
-    if static:
-        app_mode = AppModes.STATIC
+    kwargs = locals()
+    kwargs["manifest_file_name"] = manifest_file_name = validate_manifest_file(manifest_file_name)
+    app_mode = read_manifest_app_mode(manifest_file_name)
+    kwargs["title"] = title or default_title_from_manifest(manifest_file_name)
+
+    if isinstance(connect_server, api.RSConnectServer):
```
mmarchetti, line 1147: This code block occurs repeatedly - let's factor out a helper that returns a dict we can pass to `kwargs.update`.

mmarchetti (approved):

isabelizimm (commented):
popping in here to say I tested this out with vetiver, and it feels really nice! fixes the requirements bugs I had when using `python_deploy_fastapi`

PR #306, opened 2022-11-13
show help instead of err for more commands

edavidaja:
### Description

a few top level commands error rather than returning help when invoked without arguments; I've applied to the `no_args_is_help` option to those commands.


kgartland-rstudio:
@edavidaja i don't have access to your private repo to test this.  Can you bring this change over to our repo for testing?

kgartland-rstudio (approved):
Looks good. Missing commands now show help page where necessary.

Issue #310, opened 2022-11-28
deploy html does not have an `exclude` option like other content types


kgartland-rstudio:
When deploying html the `--exclude` option is not available.  Instead you must instead use `--excludes`.

Example using `--exclude`:
```
> rsconnect deploy html ./ --exclude .test-env -n dogfood
Usage: rsconnect deploy html [OPTIONS] PATH [EXTRA_FILES]...
Try 'rsconnect deploy html --help' for help.

Error: No such option: --exclude (Possible options: --excludes, --insecure)
```

Example using `--excludes`:
```
> rsconnect deploy html ./ --excludes .test-env -n dogfood
Validating server... 	[OK]
Validating app mode... 	[OK]
Making bundle ... 	[OK]
Deploying bundle ... 	[OK]
Saving deployed information... 	[OK]
Building static content...
Launching static content...
Deployment completed successfully.
	 Dashboard content URL: https://rsc.radixu.com/connect/#/apps/adebe4be-1101-4f75-af04-2d1ec7fea029
	 Direct content URL: https://rsc.radixu.com/content/adebe4be-1101-4f75-af04-2d1ec7fea029/
```

PR #312, opened 2022-11-29
add --exclude to deploy html

bcwu:
### Description

fixes https://github.com/rstudio/rsconnect-python/issues/310


kgartland-rstudio (approved):
`--exclude` now works as expected with html deploys

PR #326, opened 2022-12-16
remove RETICULATE_PYTHON detection

edavidaja:
### Description

Detection of `RETICULATE_PYTHON` leads to surprising behaviors in environment detection in workbench. As per slack discussion, this was initially set in expectation of the CLI being used to deploy content from the RStudio IDE, but this functionality was never realized.


mmarchetti:
This looks good to me. I think we should provide a changelog entry indicating why we made this (breaking) change. The CI failures may be due to the PR branch being in a fork.

mmarchetti (approved):

PR #331, opened 2023-01-11
Vetiver tests

bcwu:
### Description

Adding this as a branch instead of a fork so CI can use secrets.


.github/workflows/main.yml
```diff
@@ -146,3 +146,28 @@ jobs:
...
+    name: "Test vetiver"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
```
isabelizimm, line 159:
```suggestion
          python -m pip install --upgrade pip
          python -m pip install -r requirements.txt
```
I think we'll need to install all the requirements? I originally thought this wouldn't be necessary due to `actions/checkout@v2`, but it looks like the tests might need everything installed anyway. Hope this helps!

mmarchetti (commented):
This looks good. There are some warnings in the Vetiver tests that seem like they should be checked out.

mmarchetti (approved):

Issue #177, opened 2021-04-14
--cacert option does not handle non-ascii data


christierney:
A customer tried to use the `--cacert` option and got this error:

```
Error- Internal error: cadata should be an ASCII string or a bytes-like object
```

From slack discussion (https://rstudio.slack.com/archives/C9H3Y1KGE/p1618407352125500?thread_ts=1617900945.097000&cid=C9H3Y1KGE):

> The python function we're using is https://docs.python.org/3/library/ssl.html#ssl.create_default_context. The data we provide needs to be "either an ASCII string of one or more PEM-encoded certificates or a bytes-like object of DER-encoded certificates".

> https://github.com/pycurl/pycurl/issues/490
> It could be that we're loading the file as a string, it contains non-ascii comments, so we really need to make sure it's loaded as bytes

> If it's a PEM file with non-ascii characters in the comments, we'll need to load it as binary, then decode to ascii with discard so it ends up as an ascii string (or ask the user to remove the comments). If it's a DER certificate, we need to load as bytes and not convert to text at all. To distinguish those cases, we'd need to inspect the file contents.

Our customer-facing documentation does not give any specifics about the expected format of the certificate file.

customer issue: https://rstudioide.zendesk.com/agent/tickets/59936

DevMui:
Hello, we have the exact same issue on our RSConnect instance; is there any workaround?

greenjune-ship-it:
We also have the exact same issue for our RStudio Connect. Do you have any updates about this issue?
Thank you in advance!

PR #336, opened 2023-01-17
GH-177 - Adds support for DER formatted certficates

tdstein:
### Description

Modifies the --cacert option to support binary-encoded certificates,
such as DER (Distinguished Encoding Rules) format. This change is
backwards compatible and has no effect on text-encoded certificate.

Additionally, file-type checks have been added for common certificate
file types. This includes '.cer', '.der', '.ca-bundle', '.crt',
'.key', and '.pem'.

Connected to #177 

The included unit tests validate the suffix routing for each type of certificate.

Additional manual testing is required to validate that requests using the `--cacert` option are successful. To facilate testing of DER certificiate types, I have generate a valid DER file in this pull request: https://github.com/rstudio/connect/pull/22753

To validate DER formatting:

`rsconnect details --api-key $CONNECT_API_KEY --server https://alwayssecure.rsc:3443 --cacert alwayssecure.rsc.der`

The output should be identical to the same call with the existing CRT file.

`rsconnect details --api-key $CONNECT_API_KEY --server https://alwayssecure.rsc:3443 --cacert alwayssecure.rsc.crt`


rsconnect/api.py
```diff
@@ -360,7 +360,7 @@ def __init__(
         url: str = None,
         api_key: str = None,
         insecure: bool = False,
-        cacert: IO = None,
```
bcwu, line 363: cacert should be handled as a file, hence IO; whereas cadata is text, hence str

rsconnect/certificates.py
```diff
@@ -0,0 +1,37 @@
+from pathlib import Path
+
+BINARY_ENCODED_FILETYPES = [".cer", ".der"]
+TEXT_ENCODED_FILETYPES = [".ca-bundle", ".crt", ".key", ".pem"]
+
+
+def read_certificate_file(location: str):
```
bcwu, line 7: If we are changing cacert to location then we need to be internally consistent and change existing cacert to something like cacert_location.

bcwu: Going down that route can potentially break backwards compatibility so that's an important aspect to consider.

tdstein: Hey @bcwu - could you help clarify your comment. Are you suggesting that I rename the variables named `cacert` to `cacert_location` so that the variable type is more obvious?

rsconnect/api.py
tdstein: Correct. Here the cacert variable is a string representing the file path, instead of a file pointer. The call on line 436 opens the file pointer, which was originally handled by Click), and reads the file contents in either text mode or byte mode depending on the file-type.

The existing implementation assumes that the file pointer has been opened in text mode, which doesn't work when using DER files since they are encoded in binary format. L436 onwards, cadata is either a `str` or `bytes`, both of which are accepted when creating an SSLContext. This is the same as the existing implementation.

The Click documentation goes into scenario a little further: https://click.palletsprojects.com/en/8.1.x/arguments/#file-path-arguments

rsconnect/certificates.py
tdstein: I did a little more digging. It looks like `cafile` is the typical nomenclature for file name references. https://docs.python.org/3/library/ssl.html

bcwu: I was thinking about a couple of things:
1) Whether changing the type hints of cacert (from IO to str) will break any existing behavior
- In this case either way (type hint set to IO or str) is probably all right. In general it helps to think through whether changes are backwards compatible.
2) Possibly rename cacert if we do change the type hint
- this will probably cause more confusion than it clarifies. It will certainly break existing behavior; so, on second thought not a good idea, please disregard.

rsconnect/certificates.py
```diff
@@ -0,0 +1,37 @@
...
+
+    Files ending in '.cer' and '.der' are assumed DER (Distinguished
+    Encoding Rules) files encoded in binary format.
+
+    Files ending in '.ca-bundle', '.crt', '.key', and '.pem' are PEM
+    (Privacy Enhanced Mail) files encoded in plain-text format.
+    """
+
+    path = Path(location)
+    suffix = path.suffix
```
bcwu, line 21: Checking the file type by examining its magic number is a safer way of inference than checking the suffix. This especially pertains to security related files like certificates.

tdstein: Good idea. I'll look into this a bit. 

Do you happen to know of any documentation that shows how to accomplish this across each of the certificate types?

bcwu: We use [mimetypes](https://docs.python.org/3/library/mimetypes.html) to infer some file types. It may not have full coverage of certificates, but it's a good starting point.

tdstein: It doesn't appear that certificates contain standard header blocks, but PEM files and DER files both adhere to a standard. https://www.openssl.org/docs/man3.0/man1/openssl-format-options.html

I can add additional validation that inspects the contents to check expecatations based on this documentation. But, I'm not sure if that is necessary. 

@mmarchetti - do you have any additional insights into this?

rsconnect/certificates.py
tdstein: I checked the usage of `cacert` in each of the methods where the type signature is changed. In each case the variable is only used in combination with `read_certificate_file`. So, this shouldn't break any exisiting behavior.

rsconnect/certificates.py
mmarchetti: `mimetypes` determines types by file suffix, so it's probably equivalent to what we're doing. I don't think we need to inspect file contents - if an invalid certificate file is provided, it will fail with an SSL error when we try to connect.

mmarchetti (approved):

bcwu (approved):

kgartland-rstudio:
Verified fix.

Reproduced the issue on 1.14.0:
```
> rsconnect deploy fastapi ./ -s https://alwayssecure:3443 -k {API_KEY} --cacert ../../ssl/alwayssecure.rsc.der
    Warning: the existing manifest.json file will not be used or considered.
  File "/Users/kgartland/.pyenv/versions/3.8.2/lib/python3.8/site-packages/rsconnect/main.py", line 91, in wrapper
    result = func(*args, **kwargs)
  File "/Users/kgartland/.pyenv/versions/3.8.2/lib/python3.8/site-packages/rsconnect/main.py", line 1196, in deploy_app
    ce = RSConnectExecutor(**kwargs)
  File "/Users/kgartland/.pyenv/versions/3.8.2/lib/python3.8/site-packages/rsconnect/api.py", line 376, in __init__
    self.setup_remote_server(
  File "/Users/kgartland/.pyenv/versions/3.8.2/lib/python3.8/site-packages/rsconnect/api.py", line 436, in setup_remote_server
    ca_data = text_type(cacert.read())
  File "/Users/kgartland/.pyenv/versions/3.8.2/lib/python3.8/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x82 in position 1: invalid start byte
Internal error: 'utf-8' codec can't decode byte 0x82 in position 1: invalid start byte
```

On this branch (1.14.1.dev3+gf4f56c1), using the same command, the `der` cert works as expected.
- [x] other certs continue to work
- [x] works on Windows VM
- [x] `--insecure` option still works as expected without supplying `--cacert`

tdstein:
@bcwu / @mmarchetti - Now that @kgartland-rstudio has validated the change, am I safe to merge?

bcwu:
> @bcwu / @mmarchetti - Now that @kgartland-rstudio has validated the change, am I safe to merge?

Yes, go ahead.

Issue #341, opened 2023-01-24
Deploying additional files


MichaelScharfe81:
Has there something changed when deploying multiple (html) files to a Posit server?
For example

rsconnect deploy html -s <PositServer> -k <Key> -i -a <app> start.html a0.html a1.html a2.html a3.html

where start.html contains iframes for a0.html a1.html a2.html a3.html.

The rsconnect deploy output looks okay (no errors) and start.html is rendered. But a0.html a1.html a2.html a3.html are not found the server.

MichaelScharfe81:
I just tried the same command with version 1.13.0 ... and it works!
This looks like a bug in version 1.14.0

PR #343, opened 2023-01-25
fix extra files not being included in deploy html

bcwu:
fixes https://github.com/rstudio/rsconnect-python/issues/341


tdstein (approved):
LGTM

Validated the `set#union` does not modify the original set.

kgartland-rstudio:
Confirmed that extra files are ignored and not present in Connect after deployment when deployed with rsconnect-python 1.14.

Validated multiple files do deploy properly on this branch.

We do have a test in the rsconnect-python tests in Connect but we do _not_ check to see if the files actually made it to Connect.  I'll add an additional check there so we don't hit this again.

PR #350, opened 2023-02-09
Add jupyter intergration testing to CI

### LINKED ISSUE #351: rsconnect-jupyter + rsconnect-python integration tests

kgartland-rstudio:
## Description

`rsconnect-jupyter` uses `rsconnect-python` to add servers and deploy content. Due to this dependency, we've had to manually test `rsconnect-jupyter` for all `rsconnect-python` releases. This can cause delays in the release process or defects in `rsconnect-jupyter` if not properly tested.  Adding tests in CI protects us from making breaking changes to `rsconnect-jupyter`.  This will also set us up to test for voila mode in `rsconnect-jupyter`.

## Motivation and Context

We've had to hold releases or release patches in the past to fix breaking changes in `rsconnect-jupyter`.

## How Has This Been Tested?

The tests are currently passing:
https://github.com/rstudio/rsconnect-python/actions/runs/4135755212

Cypress failures cause CI to fail:
https://github.com/rstudio/rsconnect-python/actions/runs/4135608575

In the `intergration-testing` directory, run:
`just build && just up && just up-cypress`

`just build` build the docker images (`client`, `cypress`)
`just up` will start the docker images (`client`, `connect`) 
`just up-cypress` will start the `cypress` docker image and run the tests.
 
* The `client` container is where `rsconnect-python` and `rsconnect-jupyter` packages are installed.  `rsconnect-jupyter` installs on the `just build` command where the current dev build of `rsonnect-python` gets installed on the `just up` command so rebuilding isn't necessary to get the current `rsconnect-python` dev build.

* The `connect` container is pulled from docker hub with the latest version of Posit Connect.

* The `cypress` container runs cypress version 12.0.0 and tests all defined specs on the `just up-cypress` command.

Fixes #351
## Types of Changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] CI Changes 

## Checklist:

- [ ] I have added tests to cover my changes.


integration-testing/cypress/e2e/jupyter.cy.js
```diff
@@ -0,0 +1,55 @@
...
+    cy.get('li[class="list-group-item"]').first().should('have.text'," quandl-wiki-tsla.json.gz");
+    cy.wait(1000);
+    cy.get('a[class="btn btn-primary"]').last().should('have.text',"Publish").click({ force: true });
+    cy.wait(1000);
+    cy.get('input[name="location"]').first().click();
+    cy.wait(1000);
+    cy.get('a[class="btn btn-primary"]').last().should('have.text',"Next").click();
+    cy.wait(1000);
+    cy.get('a[class="btn btn-primary"]').last().should('have.text',"Publish").click();
+    cy.wait(1000);
```
mmarchetti, line 45: Are the `cy.wait`'s required? I thought that `cy.get` waits until elements appear.

mmarchetti (approved):

.github/workflows/main.yml
```diff
@@ -172,3 +172,45 @@ jobs:
...
+        
+      - name: Run Cypress Tests
+        run: |
+            export ADMIN_API_KEY="${{ secrets.ADMIN_API_KEY }}"
+            just integration-testing/up-cypress
+
+      # Videos are captured whether the suite fails or passes
+      - name: Save videos
+        uses: actions/upload-artifact@v3
+        if: always()
```
tdstein, line 201: This is the first time I've seen the use of `always()`, so I went to the documentation. There is a note about usage: https://docs.github.com/en/actions/learn-github-actions/expressions#always

Should this be changed so that it respects cancels?

```suggestion
        if: success() || failure()
```

.github/workflows/main.yml
```diff
@@ -172,3 +172,45 @@ jobs:
...
+        with: 
+          name: cypress-videos
+          path: integration-testing/cypress/videos
+          if-no-files-found: ignore
+          retention-days: 1
```
tdstein, line 206: Where are these files uploaded? Do I need to request access to cypress.io from IT?

integration-testing/cypress/e2e/jupyter.cy.js
```diff
@@ -0,0 +1,55 @@
+describe('Publishing Jupyter Notebook', () => {
+  beforeEach(() => {
+  });
```
tdstein, line 3: Do you think this block can be deleted?

```suggestion
```

.github/workflows/main.yml
kgartland-rstudio: They're uploaded as artifacts.  You can find them at the bottom of the page:
https://github.com/rstudio/rsconnect-python/actions/runs/4135994847

tdstein: Very cool! Thank you!

integration-testing/cypress/e2e/jupyter.cy.js
tdstein: Is it possible to wait for some action to complete, instead of an arbitrary time? For example, if the click() action triggers a network request.

kgartland-rstudio: I've tried waiting for different elements to load but cypress is simply running too fast, the only way to avoid flakiness here I think is to hardcode `cy.wait()` each time the modal loads. 

Here is a screenshot of what's happening.
Screen Shot 2023-02-09 at 3 23 09 PM

Cypress code (after removing `cy.wait()`):
```
    cy.get('button[id="rsc-add-files"]')
      .should('be.visible')
      .and('contain',"Select Files...")
      .click();
```
This says the Select Files button should be visible and contain "Select Files..." and if so, click it.  But Cypress "clicks" it before the element is ready and then can't find the files from the _next_ page and fails there.

This seems to be Cypress' biggest issue, there are loads of forum posts and blogs about it and most come to the conclusion, just add a wait(). 

I rather like this blog post, it's right in the title, `waiting-in-cypress-and-how-to-avoid-it` but then immediately admits, it's probably best to add the wait().
https://filiphric.com/waiting-in-cypress-and-how-to-avoid-it

PR #358, opened 2023-03-03
add is_python_enabled_on_server

bcwu:
## Description
fixes https://github.com/rstudio/connect/issues/23063


rsconnect/bundle.py
```diff
@@ -1062,6 +1062,15 @@ def fake_module_file_from_directory(directory: str):
     return join(directory, app_name + ".py")
 
 
+def is_python_enabled_on_server(connect_details):
+    """
+    Returns whether or not the Connect server has Python itself enabled.
+
+    :error: The Posit Connect server does not have Python enabled.
+    """
+    return any(connect_details["python"]["versions"])
```
tdstein, line 1071: Here's another way to avoid the KeyError when the "python" key doesn't exist.

```suggestion
    return "python" in connect_details and "versions" in connect_details["python"]
```

bcwu: The reason we use `any` here is because "versions' is a list, so we also need to check whether it is empty.

bcwu: This is what connect_details  looks like:
connect_details = {'connect': '2023.02.0-dev+202-dirty', 'python': {'api_enabled': False, 'versions': []}, 'conda': {'supported': False}}

tdstein: That makes sense. 

The potential issue is that the key `python` may not exist in the object `connect_details`, which will result in a KeyError raise. This same logic is also true for the `versions` key.

The following snippet should avoid situations where the KeyError may be raised. 

```suggestion
    return any(connect_details.get("python", {}).get("versions", []))
```

kgartland-rstudio:
Validated `rsconnect deploy api` now throws the proper error:

```
> rsconnect deploy api . -s http://34.229.160.227:3939 -k ${DEFAULT_ADMIN_API_KEY} -N
Validating server... 	[OK]
Validating app mode... 	[OK]
Error: The Posit Connect server does not have Python enabled.
```

Issue #359, opened 2023-03-08
`deploy html` outside of working directory fails 


kgartland-rstudio:
Deploying html from outside of the directory fails with the following error:

```
> rsconnect deploy html bundles/html-single-file/index.html -n dogfood
Validating server... 	[OK]
Validating app mode... 	[OK]
Making bundle ... 	[ERROR]: Unable to include the file index.html in the bundle: No such file or directory
Error: Unable to include the file index.html in the bundle: No such file or directory
```

Deploying from within the directory succeeds:
```
> rsconnect deploy html index.html -n dogfood
Validating server... 	[OK]
Validating app mode... 	[OK]
Making bundle ... 	[OK]
Deploying bundle ... 	[OK]
Saving deployed information... 	[OK]
Building static content...
Launching static content...
Deployment completed successfully.
```

kgartland-rstudio:
We also no do not allow directory deploys for html content on the voila branch:
```
> rsconnect version
1.15.0b2.dev1+g5566262
> rsconnect deploy html bundles/html-single-file/ -n dogfood
Validating server... 	[OK]
Validating app mode... 	[OK]
Making bundle ... 	[ERROR]: Unable to find a valid html entry point.
Error: Unable to find a valid html entry point.
```

The public build deploys successfully:
```
> rsconnect version
1.14.1
> rsconnect deploy html bundles/html-single-file/ -n dogfood
Validating server... 	[OK]
Validating app mode... 	[OK]
Making bundle ... 	[OK]
Deploying bundle ... 	[OK]
Saving deployed information... 	[OK]
Building static content...
Launching static content...
Deployment completed successfully.
```

bcwu:
> We also no do not allow directory deploys for html content on the voila branch:
> 
> ```
> > rsconnect version
> 1.15.0b2.dev1+g5566262
> > rsconnect deploy html bundles/html-single-file/ -n dogfood
> Validating server... 	[OK]
> Validating app mode... 	[OK]
> Making bundle ... 	[ERROR]: Unable to find a valid html entry point.
> Error: Unable to find a valid html entry point.
> ```
> 
> The public build deploys successfully:
> 
> ```
> > rsconnect version
> 1.14.1
> > rsconnect deploy html bundles/html-single-file/ -n dogfood
> Validating server... 	[OK]
> Validating app mode... 	[OK]
> Making bundle ... 	[OK]
> Deploying bundle ... 	[OK]
> Saving deployed information... 	[OK]
[...]

PR #360, opened 2023-03-08
add absolute path file list and use it for voila

bcwu:
## Description
fixes https://github.com/rstudio/rsconnect-python/issues/359

### DIFF
--- a/rsconnect/bundle.py
+++ b/rsconnect/bundle.py
@@ -890,7 +890,7 @@ def create_file_list(
     file_set = set(extra_files)  # type: typing.Set[str]
 
     if isfile(path):
-        file_set.add(Path(path).name)
+        file_set.add(path)
         return sorted(file_set)
 
     for cur_dir, sub_dirs, files in os.walk(path):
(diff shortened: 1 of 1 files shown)


mmarchetti:
This PR introduces more differences in behavior between Voila and other content types. I would prefer to see us go the other way toward more commonality. We have existing code that handles both file and directory mode deployments; let's use it if possible or emulate its behaviors if not.

bcwu:
> 

The divergence is definitely not ideal, but I believe we have exhausted the compatibility of existing code. 

We can converge them again when we refactor existing code in bundle.py to use the Manifest and Bundle classes (or do the reverse and give up on the idea of Manifest and Bundle), but that refactor will be some effort either way, and seems out of scope.

mmarchetti:
I know we won't be able to converge the implementation right now, but it is important that we converge the behavior. We would benefit from some tests that verify that Voila and other deployment types follow the same inclusion/exclusion rules (except for things like the default entrypoint).

PR #365, opened 2023-03-14
test on windows-latest Python 3.9

### LINKED ISSUE #364: Windows CI for rsconnect-python

isabelizimm:
## Description

Adding Windows testing for Python 3.9 in CI

## Motivation and Context

resolves #364

## How Has This Been Tested?

N/A

## Types of Changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] I have added tests to cover my changes.


isabelizimm:
I was mostly curious what would happen if I threw in windows in CI...and well...curiosity killed the GHA 🐱

tests/test_bundle.py
```diff
@@ -828,6 +828,7 @@ def test_is_directory(self):
 
     def test_is_not_executable(self):
         with tempfile.NamedTemporaryFile() as tmpfile:
+            # assert os.access(tmpfile.name, os.X_OK) == False #for windows this fails, linux this passes
```
isabelizimm, line 831: Not sure what to do with this test: inside the context manager of a temp file in MacOS/Linux , `os.access(tmpfile.name, os.X_OK) == False`, which triggers the RSConnectError. 

However, on Windows, `os.access(tmpfile.name, os.X_OK) == True`, so this test fails.

isabelizimm: Looking at the [comp.lang.python google group](https://groups.google.com/g/comp.lang.python/c/FCiVzouJt5s) , it seems like, on Windows, `os.access()` will generally return True if the path can be accessed at all.

I tried to update with `os.chmod(tmpfilename, 000)` and still could not force the RSConnectException 😩.

NOTE: the discussion is from 2007, but it does seem that the behavior is still the same.

mmarchetti (approved):

PR #384, opened 2023-03-24
Add `system caches` `list` and `delete` subcommands to manage runtime caches

toph-allen:
## Intent

Add new subcommands to call Connect's forthcoming runtime cache management API.

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach

Note: CI is failing because of one of the Connect-in-Docker integration tests I wrote, where I try to use Python's `system()` to run commands in the Docker container. See below for more details. I'll take another shot at it on Monday.

There are three main functional components:
- Low-level functions in `RSConnectClient` that call the endpoints and return the response.
- Higher-level functions in `RSConnectExecutor`.
- Command-line definitions.

The `list` command is very simple: it either serializes the JSON response from the server, or prints the error.

The `delete` command is more complex, because it has a dry-run case, which doesn't result in a server-side task, and a non-dry-run case, which _does_ result in the creation of a task. The logic to handle this is in `RSConnectExecutor.delete_runtime_cache()`.

- Dry-run invocations of `delete` contain `"task_id": null`. In this case, we print "Dry run finished", as the server doesn't log anything. If the command is called with `-v`, the result will also be printed. In the dry run case, if the deletion wouldn't happen for any reason (e.g. the cache doesn't exist) the endpoint will return an error.
- Wet-run invocations of `delete` will have a non-null `task_id`. The endpoint will call `client.wait_for_task()` and log server messages from that task, which include a start and a success message. The task status object is silently returned at the end, but it'll be printed with `-v`.

## Automated Tests

The API is tested on two levels:
- Unit tests in `test_api.py`
  - These use `httpretty` to mock responses to confirm that the endpoints function as desired when given the right inputs.
- Integration tests building on the approach used by the `vetiver` tests, in `-test_main_system_caches.py`
  - These spin up a Connect server and use `click.CliRunner()` to test the end-to-end functionality.

## QA Notes

We'll manually test that we can enumerate and delete caches.

- Spin up a Connect server. 
- Create a fake cache directory, e.g. `mkdir pip/1.2.3` within the python-environments directory.
- Use `rsconnect system caches list` to list the runtime caches.
- Use `rsconnect system caches delete --language Python --version 1.2.3 --image-name Local` to delete the cache directory.

## Checklist

I have not updated `CHANGELOG.md` yet. Will do so before this merges.


rsconnect/main.py
```diff
@@ -2417,6 +2417,119 @@ def start_content_build(
...
+    help="Disable TLS certification/host validation.",
+)
+@click.option(
+    "--cacert",
+    "-c",
+    envvar="CONNECT_CA_CERTIFICATE",
+    type=click.Path(exists=True, file_okay=True, dir_okay=False),
+    help="The path to trusted TLS CA certificates.",
+)
+@click.option("--verbose", "-v", is_flag=True, help="Enable verbose output")
```
bcwu, line 2462:
```suggestion
@server_args
```

rsconnect/main.py
```diff
@@ -2417,6 +2417,119 @@ def start_content_build(
...
+    is_flag=True,
+    help="Disable TLS certification/host validation.",
+)
+@click.option(
+    "--cacert",
+    "-c",
+    envvar="CONNECT_CA_CERTIFICATE",
+    type=click.Path(exists=True, file_okay=True, dir_okay=False),
+    help="The path to trusted TLS CA certificates.",
+)
```
bcwu, line 2502:
```suggestion
@server_args
```

rsconnect/main.py
toph-allen: Oh, thanks! I figured there must be a more concise way to do that.

rsconnect/main.py
```diff
@@ -2417,6 +2417,119 @@ def start_content_build(
...
+    "-I",
+    help='Either "Local" or the image name of the target cache\'s execution environment.',
+)
+@click.option(
+    "--dry-run",
+    "-d",
+    is_flag=True,
+    help="If true, verify that deletion would occur, but do not delete.",
+)
+@click.option("--verbose", "-v", is_flag=True, help="Enable verbose output")
```
bcwu, line 2524:
```suggestion
```

covered by server_args

toph-allen: Thanks. I didn't use `@server_args` because the furthest-down examples in the page I could find restated them explicitly, so I've replaced the redundant args throughout `main.py` as well.

rsconnect/api.py
```diff
@@ -981,6 +991,21 @@ def make_deployment_name(self, title, force_unique):
 
         return name
 
+    def list_runtime_caches(self):
+        return self.client.system_caches_runtime_list()
```
bcwu, line 995:
```suggestion
    @property
    def runtime_caches(self):
        return self.client.system_caches_runtime_list()
```
The design pattern for Connect Executor is to make chainable functions, and to make non-chainable information into properties.

rsconnect/api.py
```diff
@@ -981,6 +991,21 @@ def make_deployment_name(self, title, force_unique):
...
+        result = self.client.system_caches_runtime_delete(target)
+        if result["task_id"] is None:
+            print("Dry run finished")
+            return result
+        else:
+            (log_lines, task_status) = self.client.wait_for_task(
+                result["task_id"], connect_logger.info, raise_on_error=False
+            )
+            return task_status
+
```
bcwu, line 1008: Executor functions should be chainables that ends with `return self`.

rsconnect/main.py
tdstein, line 1: Thank you! This refactor was needed!

.github/workflows/main.yml
```diff
@@ -171,7 +171,9 @@ jobs:
       # NOTE: edited to run checks for python package
       - name: Run tests
         run: |
+          python -m unittest tests.test_main_system_caches
           pytest -m 'vetiver'
```
tdstein, line 175:
```suggestion
          pytest tests/test_main_system_caches.py
          pytest -m 'vetiver'
```

This should also work.

tdstein (commented):
Tests that depend on docker are isolated from unit tests within the project structure. In my opinion, the integration tests should be moved to another directory to separate them from the unit tests. Then a command should be added to the Makefile, which invokes them independently. 

That way, I should be able to run `make integration-tests` from my machine and get the same results as running the same command in GitHub Actions.

Also, this is the first time I've seen setup and teardown Docker commands invoked as part of the unit test lifecycle. I have always utilized shell scripts to achieve this. Are there any pros or cons to running the Docker commands via the Python `os` package?

rsconnect/api.py
toph-allen: Gotcha. I understand that information like the list function above should be a property. With this, I should just return `self` at the end of the function and no other changes? Or does it make sense to surface the final task result somehow? I guess I could pass verbosity down into this function.

toph-allen:
> Tests that depend on docker are isolated from unit tests within the project structure. In my opinion, the integration tests should be moved to another directory to separate them from the unit tests. Then a command should be added to the Makefile, which invokes them independently.
> 
> That way, I should be able to run `make integration-tests` from my machine and get the same results as running the same command in GitHub Actions.
> 
> Also, this is the first time I've seen setup and teardown Docker commands invoked as part of the unit test lifecycle. I have always utilized shell scripts to achieve this. Are there any pros or cons to running the Docker commands via the Python `os` package?

Yeah, having these tests as unit tests that are just skipped if the docker service isn't available seems really hacky. I agree that there should be a single `make` target to run them.

There's already an `integration-tests` directory, which — I'm not sure what it's used for. I was initially thinking about renaming the `vetiver-tests` directory to `integration-tests` till I saw there was already a directory there, because I'm trying to contain the scope of this PR to getting this cache deletion story done and not trying to make infrastructure changes to `rsconnect-python`. Probably makes sense to put it into a follow-up PR? @bcwu thoughts?

rsconnect/api.py
bcwu: Yeah, a `return self` here seems right. `wait_for_task` is printing out the logs so the thing left to do is decide whether you want to add any custom exceptions based on the task_status in the `else` block. Or just pass `raise_on_error=True`

```suggestion
    def delete_runtime_cache(self, language, version, image_name, dry_run):
        target = {"language": language, "version": version, "image_name": image_name, "dry_run": dry_run}
        result = self.client.system_caches_runtime_delete(target)
        if result["task_id"] is None:
            print("Dry run finished")
        else:
            (log_lines, task_status) = self.client.wait_for_task(
                result["task_id"], connect_logger.info, raise_on_error=False
            )
        return self
```

bcwu:
> > Tests that depend on docker are isolated from unit tests within the project structure. In my opinion, the integration tests should be moved to another directory to separate them from the unit tests. Then a command should be added to the Makefile, which invokes them independently.
> > That way, I should be able to run `make integration-tests` from my machine and get the same results as running the same command in GitHub Actions.
> > Also, this is the first time I've seen setup and teardown Docker commands invoked as part of the unit test lifecycle. I have always utilized shell scripts to achieve this. Are there any pros or cons to running the Docker commands via the Python `os` package?
> 
> Yeah, having these tests as unit tests that are just skipped if the docker service isn't available seems really hacky. I agree that there should be a single `make` target to run them.
> 
> There's already an `integration-tests` directory, which — I'm not sure what it's used for. I was initially thinking about renaming the `vetiver-tests` directory to `integration-tests` till I saw there was already a directory there, because I'm trying to contain the scope of this PR to getting this cache deletion story done and not trying to make infrastructure changes to `rsconnect-python`. Probably makes sense to put it into a follow-up PR? @bcwu thoughts?

A separate PR makes sense. For now, feel free to make as much (or as little) change you feel is needed to get the feature ready. That is to say, if we can get the tests pass CI, feel free to leave them in the PR.

toph-allen:
@bcwu @tdstein Right now the Vetiver integration test workflow uses the local `make dev` target that's used for running locally. This uses the `-T` flag, which makes it run as a TTY, which is causing the integration test failures atm.

The `dev` target is currently split across a few sub targets. Roughly excerpted:

```make
RSC_API_KEYS=vetiver-testing/rsconnect_api_keys.json

...

dev: vetiver-testing/rsconnect_api_keys.json

dev-start:
	docker-compose up -d
	docker-compose exec -T rsconnect bash < vetiver-testing/setup-rsconnect/add-users.sh
	# curl fails with error 52 without a short sleep....
	sleep 5
	curl -s --retry 10 --retry-connrefused http://localhost:3939

...

$(RSC_API_KEYS): dev-start
	python vetiver-testing/setup-rsconnect/dump_api_keys.py $@
```

But the GitHub Actions `.yml` _also_ calls `docker-compose up --build -d` just before it calls `make dev`.

I've honestly hardly ever used `make` — Connect switched to `just` right when I joined the team — so it's not immediately obvious what the simplest way to make the `-T` change in CI is without breaking the local dev setup or making the `Makefile` overly complex. I'll look into that now.

tests/test_api.py
```diff
@@ -173,12 +169,7 @@ def test_executor_delete_runtime_cache_wet_run(self):
         sys.stdout = sys.__stdout__
 
         # Print expectations
-        # TODO: *We* don't print anything here anymore. Unsure how to capture log messages from Connect.
-        # output_lines = captured_output.getvalue().splitlines()
-        # self.assertEqual(output_lines[0], "Cache deletion finished")
-
-        # Result expectations
-        self.assertDictEqual(mocked_task_status, result)
+        # TODO: Figure out how to capture the output of `connect_logger` and make assertions.
```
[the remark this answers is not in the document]
toph-allen: Thanks

toph-allen:
It looks like I've got _my_ integration tests working but have now broken the Vetiver integration tests. Will continue poking at this tomorrow.

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+- Added `system caches list` and `system caches delete` commands which allow administrators to enumerate and delete R and Python runtime caches from Connect servers.
```
aronatkins, line 10: We will want to indicate the minimum Connect release having this functionality.

README.md
```diff
@@ -978,3 +978,86 @@ $ rsconnect bootstrap --server https://connect.example.org:3939 --jwt-keypath /p
 
 A full description on how to use `rsconnect bootstrap` in a provisioning workflow is provided in the Connect administrator guide's 
 [programmatic provisioning](https://docs.posit.co/connect/admin/programmatic-provisioning) documentation.
+
+## Server Administration Tasks
+
+Starting with the 2023.04 edition of Posit Connect, `rsconnect-python` can be
+used to perform certain server administration tasks, for instance managing
+runtime caches. For more information on runtime caches in Posit Connect, see the
+Connect Admin Guide's section on [runtime caches](TODO:Link).
```
aronatkins, line 987: https://docs.posit.co/connect/admin/server-management/runtime-caches/

README.md
```diff
@@ -978,3 +978,86 @@ $ rsconnect bootstrap --server https://connect.example.org:3939 --jwt-keypath /p
...
+    },
+    {
+      "language": "Python",
+      "version": "3.8.12",
+      "image_name": "Local"
+    },
+        {
+      "language": "R",
+      "version": "3.6.3",
+      "image_name": "teapot"
```
aronatkins, line 1022: Let's stop using `teapot` in real examples. Maybe use one of the r-base images? https://hub.docker.com/r/rstudio/r-base

README.md
```diff
@@ -978,3 +978,86 @@ $ rsconnect bootstrap --server https://connect.example.org:3939 --jwt-keypath /p
...
+    },
+    {
+      "language": "Python",
+      "version": "3.8.12",
+      "image_name": "teapot"
```
aronatkins, line 1027: Same for Python - maybe use one of the Python images to reference a real one? https://hub.docker.com/_/python/

README.md
```diff
@@ -978,3 +978,86 @@ $ rsconnect bootstrap --server https://connect.example.org:3939 --jwt-keypath /p
...
+
+### Delete Runtime Caches
+
+When Connect's execution environment changes, runtime caches may be invalidated.
+In these cases, you will need to delete the affected runtime caches using the
+`system caches delete` command.
+
+!!! warning
+
+  After deleting a cache, the first time affected content is visited, Connect
```
aronatkins, line 1041: In the case of rendered content, it is on first run, not first visit. Maybe just say that the rebuild happens when it is first executed / run.

README.md
```diff
@@ -978,3 +978,86 @@ $ rsconnect bootstrap --server https://connect.example.org:3939 --jwt-keypath /p
...
+```bash
+❯ rsconnect system caches delete --name myserver --language Python --version 5.6.7 --image-name teapot
+Deleting runtime cache
+Successfully deleted runtime cache
+```
+
+You should run this command for each cache you wish to delete.
+
+To determine whether your a deletion request would fail without risking
+deletion, you can use the dry run option (`-d, --dry-run`) to surface any errors
```
aronatkins, line 1062: Suggest the dry-run as an integral part of the workflow, not an afterthought.

1. dry-run (with example)
2. delete (with example)

README.md
```diff
@@ -978,3 +978,86 @@ $ rsconnect bootstrap --server https://connect.example.org:3939 --jwt-keypath /p
...
+
+To delete a runtime cache, call the `system caches delete` command, specifying a
+Connect server, as well as the language (`-l, --language`), version (`-V,
+--version`), and image name (`-I, --image-name`) for the cache you wish to
+delete. Deleting a large cache might take a while. The command will wait for
+Connect to finish the task.
+
+```bash
+❯ rsconnect system caches delete --name myserver --language Python --version 5.6.7 --image-name teapot
+Deleting runtime cache
```
aronatkins, line 1055: Do not use prompt characters. Output should be prefaced with comment characters.

tdstein (approved):

aronatkins (approved):

toph-allen:
Just noting here that @aronatkins, @ChaitaC and I verified this on Thursday, March 30th in our Refinement meeting. I failed to record that when it happened.

bcwu (approved):

PR #386, opened 2023-03-27
release-1.15.1

bcwu:
## Intent
Update changelog with release date.


tdstein:
Should this be a minor release instead of a patch since the CONNECT_REQUEST_TIMEOUT env var is a new feature?

Issue #376, opened 2023-03-21
rsconnect-python unit and integration tests


bcwu:
have unit test coverage for every deploy option
- [x] deploy voila
    - [x] manifest test
    - [x] bundle tests
- [x] deploy html
    - [x] manifest test
    - [x] bundle tests
- [x] deploy api
    - [x] manifest test
    - [x] bundle tests
- [x] deploy bokeh
    - [x] manifest test
    - [x] bundle tests
- [x] deploy dash
    - [x] manifest test
    - [x] bundle tests
- [x] deploy fastapi
    - [x] manifest test
    - [x] bundle tests
- [x] deploy streamlit
    - [x] manifest test
    - [x] bundle tests
- [x] deploy manifest
    - [x] bundle tests
- [x] deploy notebook
    - [x] manifest test
    - [x] bundle tests
- [x] deploy shiny
    - [x] manifest test
    - [x] bundle tests
- [x] deploy quarto
    - [x] manifest test
    - [x] bundle tests

PR #395, opened 2023-03-28
deploy manifest bundle test

bcwu:
## Intent
Add deploy manifest bundle test as part of #376


tests/test_bundle.py
```diff
@@ -1804,3 +1804,39 @@ def test_make_html_bundle():
             "manifest.json",
         ]
         assert multi_file_index_dir_extras_ans == json.loads(tar.extractfile("manifest.json").read().decode("utf-8"))
+
+
+pyshiny_manifest_dir = os.path.join(cur_dir, "./testdata/pyshiny_with_manifest")
+pyshiny_manifest_file = os.path.join(cur_dir, "./testdata/pyshiny_with_manifest/manifest.json")
+
+
+def test_make_manifest_bundle():
```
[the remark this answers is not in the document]
bcwu: The intent is to replace manifest and bundle related functions with the Manifest and Bundle classes, so the purpose of these specific tests are to document existing end to end behaviors. A refactored make_manifest_bundle most likely won't have the same logic, but its outputs must still be consistent with the existing behavior.

mmarchetti:
This test is a good addition, but limited to the happy path. Can you add some failure cases to the tests? For example, one of the main failure modes I'd expect to see with manifest-based deployments is missing files - ones that existed when the manifest was created but don't any more. Are there other scenarios that might (reasonably) arise in a real client environment that could break manifest deployments?

tests/test_bundle.py
```diff
@@ -2425,3 +2425,26 @@ def test_make_manifest_bundle():
...
+        "metadata": {"appmode": "python-shiny", "entrypoint": "app5"},
+        "python": {
+            "version": "3.8.12",
+            "package_manager": {"name": "pip", "version": "23.0.1", "package_file": "requirements.txt"},
+        },
+        "files": {
+            "requirements1.txt": {"checksum": "c82f1a9894e5510b2f0c16fa63aaa004"},
+        },
+    }
+    with pytest.raises(TypeError):
```
mmarchetti, line 2449: I would have expected this to raise a different error, why `TypeError`?

bcwu: Existing code path has no custom exceptions, so the error we are getting back is from this file handler: https://github.com/rstudio/rsconnect-python/blob/51eb48b9652ca7adede90af89ed351c58234c2a9/rsconnect/bundle.py#L714-L726

mmarchetti: `make_manifest_bundle` is expecting a file path to the manifest, so you'll need to put your manifest into a file. A subdirectory in the `testdata` directory is the usual place for things like that.

bcwu: This is resolved with [79d98b5](https://github.com/rstudio/rsconnect-python/pull/395/commits/79d98b5e5dbc493833eab62635b2932ec9eb8762) and [a7c46d9](https://github.com/rstudio/rsconnect-python/pull/395/commits/a7c46d9ab3f0172d53d9f607a009bd59e66beeef).

tests/test_bundle.py
```diff
@@ -2427,24 +2427,15 @@ def test_make_manifest_bundle():
...
 
+empty_manifest_file = os.path.join(cur_dir, "./testdata/Manifest_data/empty_manifest.json")
+missing_file_manifest = os.path.join(cur_dir, "./testdata/Manifest_data/missing_file_manifest.json")
+
+
 def test_make_bundle_empty_manifest():
-    manifest = {}
     with pytest.raises(TypeError):
-        make_manifest_bundle(manifest)
+        make_manifest_bundle(Exception)
```
mmarchetti, line 2436: Seems like you meant `pytest.raises(Exception)`?

mmarchetti (approved):

PR #415, opened 2023-04-26
start setting record version on metadata records

mslynch:
## Intent

Intended to support compatibility between newer versions of rsconnect-python and older file formats.

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Checklist

### DIFF
--- a/rsconnect/metadata.py
+++ b/rsconnect/metadata.py
@@ -18,7 +18,6 @@ from .exception import RSConnectException
 from .log import logger
 from .models import AppMode, AppModes
 
-APPSTORE_VERSION = 1
 
 def config_dirname(platform=sys.platform, env=os.environ):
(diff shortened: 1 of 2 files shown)


rsconnect/metadata.py
```diff
@@ -18,6 +18,7 @@
 from .log import logger
 from .models import AppMode, AppModes
 
+METADATA_RECORD_VERSION = 1
```
bcwu, line 21: Metadata comprises DataStore, ServerData, AppStore. Unless we are sure all three evolve at the same time and at the same rate, this would probably better off implemented as an appstore_version first.

mslynch:
Closing – folded into https://github.com/rstudio/rsconnect-python/pull/418.

PR #418, opened 2023-05-03
implement publishing static outputs for cloud

mslynch:
## Intent

Adds support for `deploy html` and `deploy manifest` to deploy to Posit Cloud.

Resolves https://github.com/rstudio/rsconnect-python/issues/420.

This includes the changes from https://github.com/mslynch/rsconnect-python/pull/1, plus versioned app store files and more tests.

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach

Allows for static content to be deployed to Posit Cloud.

This PR also introduces file versioning to the app store files with an initial version of 1. For cloud deployments, `app_id` will now store the content id (as in `https://posit.cloud/content/{content_id}` (as opposed to the previous approach of storing the application id).

Also contains a refactor to Cloud's HTTP client to handle errors within the client and return the parsed response body from its calls.

## Automated Tests

Added a test with `rsconnect deploy manifest` using a manifest for a `static`-appmode application. Also added several unit tests for `CloudService.prepare_deploy` to test deploy/redeploy/pre-versioned app store files.

## Directions for Reviewers

To test this manually, you can use:
```shell
rsconnect deploy html \
          --account $YOUR_ACCOUNT \
          --token $YOUR_TOKEN \
          --secret $YOUR_SECRET \
          --server 'https://api.staging.posit.cloud' \
          'some_file.html'
```

To test compatibility with pre-versioned app store files,
- do a deploy on the current version:
  ```shell
  rsconnect deploy shiny \
            --account $YOUR_ACCOUNT \
            --token $YOUR_TOKEN \
            --secret $YOUR_SECRET \
            --server 'https://api.staging.posit.cloud' \
            'my-shiny-app'
  ```
- then check out main, reinstall, and run the command again.

## Checklist

### DIFF
--- a/rsconnect/metadata.py
+++ b/rsconnect/metadata.py
@@ -406,12 +406,13 @@ class AppStore(DataStore):
     are made.
     """
 
-    def __init__(self, app_file):
+    def __init__(self, app_file, version=1):
         base_name = str(basename(app_file).rsplit(".", 1)[0]) + ".json"
         super(AppStore, self).__init__(
             join(dirname(app_file), "rsconnect-python", base_name),
             join(config_dirname(), "applications", sha1(abspath(app_file)) + ".json"),
         )
+        self.version = version
 
     def get(self, server_url):
         """
@@ -449,7 +450,7 @@ class AppStore(DataStore):
                 app_guid=app_guid,
                 title=title,
                 app_mode=app_mode.name() if isinstance(app_mode, AppMode) else app_mode,
-                app_store_version=self.current_app_store_version,
+                app_store_version=self.version,
             ),
         )
 
@@ -457,7 +458,7 @@ class AppStore(DataStore):
         metadata = self.get(server)
         if metadata is None:
             logger.debug("No previous deployment to this server was found; this will be a new deployment.")
-            return app_id, app_mode, self.current_app_store_version
+            return app_id, app_mode, self.version
 
         logger.debug("Found previous deployment data in %s" % self.get_path())
 
@@ -470,10 +471,6 @@ class AppStore(DataStore):
 
         app_store_version = metadata.get("app_store_version")
         return app_id, app_mode, app_store_version
-
-    @property
-    def current_app_store_version(self):
-        return 1
 
 
 DEFAULT_BUILD_DIR = join(os.getcwd(), "rsconnect-build")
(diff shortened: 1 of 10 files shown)


rsconnect/metadata.py
```diff
@@ -380,6 +380,9 @@ def sha1(s):
     return base64.urlsafe_b64encode(m.digest()).decode("utf-8").rstrip("=")
 
 
+CURRENT_APP_STORE_VERSION = 1
```
bcwu, line 383: Are there other use cases for this to be a global variable? If not then this can be replace with just the class initialization.

rsconnect/metadata.py
```diff
@@ -446,14 +453,15 @@ def set(self, server_url, filename, app_url, app_id, app_guid, title, app_mode):
...
+                app_store_version=self.app_store_version,
             ),
         )
 
     def resolve(self, server, app_id, app_mode):
         metadata = self.get(server)
         if metadata is None:
             logger.debug("No previous deployment to this server was found; this will be a new deployment.")
-            return app_id, app_mode
+            return app_id, app_mode, CURRENT_APP_STORE_VERSION
```
bcwu, line 464:
```suggestion
            return app_id, app_mode, self.app_store_version
```

tests/test_api.py
```diff
@@ -6,11 +6,21 @@
 import sys
 import io
 from rsconnect.exception import RSConnectException
+from rsconnect.metadata import CURRENT_APP_STORE_VERSION
```
bcwu, line 9: The testing strategy might be better if we have a scaffold of tests for each version. That way we can just add new tests when version advances instead of needing to modify or delete existing test. We also want to have tests for older versions.

rsconnect/metadata.py
```diff
@@ -475,6 +471,10 @@ def resolve(self, server, app_id, app_mode):
         app_store_version = metadata.get("app_store_version")
         return app_id, app_mode, app_store_version
 
+    @property
+    def current_app_store_version(self):
+        return 1
+
```
bcwu: We should put current_app_store_version - let's rename it 'version' for brevity - as a default parameter with default of 1 in __init__, then save it in self.version. We probably don't need this property at this point since we can just init the class (e.g. app_store = AppStore() ) and check app_store.version directly.

omar-rs (approved):

bcwu (approved):

PR #434, opened 2023-06-26
Updates Changeling with Release 1.18.0

tdstein:
## Intent

Update the Changelog

## Type of Change

- [ ] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change   

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Deploys for Posit Cloud and shinyapps.io now accept the `--visibility` flag.
 
+### Changed
+- Removes redundant client-side compatibility checks in favor of server-side compatibility checks when deploying Python content. Note that the error handling may differ between versions of Connect. See Connect release notes for additional details regarding compatibility.
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.18.0] - 2023-06-26
+
+### Added
+- Deploys for Posit Cloud and shinyapps.io now accept the `--visibility` flag.
```
mmarchetti, line 10: We should have an entry here for the change to how we do preflight checking of Shiny app deployability.

tdstein: Would you consider this a fix or a change?

mmarchetti: I see it as a  change - we're actually removing the preflight check. It might be worth noting that the result will depend on the version of Connect.

tdstein: Do you know which version of Connect the server side check was introduced? If not, I can leave track it down.

mmarchetti (approved):

PR #445, opened 2023-07-06
Adds CONNECT_TASK_TIMEOUT environment variable.

tdstein:
## Intent

Adds a new environment variable for manipulating how long the client should wait for tasks to complete on the server. Previously, the CONNECT_REQUEST_TIMEOUT value was used. This value defaults to 300 seconds, too short for long-running tasks.

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach

Mimics the implementation for CONNECT_REQUEST_TIMEOUT with slight variation to accommodate tasks vs HTTP requests.

## Automated Tests

Adds unit tests for obtaining the timeout value.

## Directions for Reviewers

Decide if the default value I chose is appropriate. Currently, the client will wait indefinitely. Is this what we want?

## Checklist

### DIFF
--- a/rsconnect/api.py
+++ b/rsconnect/api.py
@@ -300,8 +300,6 @@ class RSConnectClient(HTTPServer):
         poll_wait=0.5,
         raise_on_error=True,
     ):
-        last_status = None
-        ending = time.time() + timeout if timeout else 999999999999
 
         if log_callback is None:
             log_lines = []
@@ -309,10 +307,12 @@ class RSConnectClient(HTTPServer):
         else:
             log_lines = None
 
+        last_status = None
+        start_time = time.time()
         sleep_duration = 0.5
         time_slept = 0
         while True:
-            if time.time() >= ending:
+            if (time.time() - start_time) > timeout:
                 raise RSConnectException("Task timed out after %d seconds" % timeout)
             elif abort_func():
                 raise RSConnectException("Task aborted.")
@@ -747,7 +747,7 @@ class RSConnectExecutor:
         task_id: int = None,
         log_callback=connect_logger,
         abort_func: Callable[[], bool] = lambda: False,
-        timeout: int = None,
+        timeout: int = get_task_timeout(),
         poll_wait: float = 0.5,
         raise_on_error: bool = True,
     ):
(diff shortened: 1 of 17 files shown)


rsconnect/api.py
```diff
@@ -292,7 +292,13 @@ def get_content(self, content_guid):
         return results
 
     def wait_for_task(
-        self, task_id, log_callback, abort_func=lambda: False, timeout=get_timeout(), poll_wait=0.5, raise_on_error=True
+        self,
```
aronatkins, line 295: The `ending` calculation can overflow.

```
>>> t = time.time()
>>> ms = sys.maxsize
>>> ms
9223372036854775807
>>> t
1688655596.327841
>>> t + ms
9.223372038543432e+18
```

Should we switch the logic that computes `ending` into something similar to what is in `wait_until_task_is_successful()`, where we `time.time() - start_time < timeout`? This approach avoids using values greater than `sys.maxsize`.

aronatkins (approved):
Do these timeout settings appear in documentation besides the changelog?

rsconnect/api.py
tdstein: Thanks, @aronatkins - I adjusted the logic to avoid the overflow error.

https://github.com/rstudio/rsconnect-python/pull/445/commits/456c6914636768f0ca69b124e01847ee9d07720a

rsconnect/api.py
```diff
@@ -292,21 +292,26 @@ def get_content(self, content_guid):
...
         else:
             log_lines = None
 
+        last_status = None
+        start_time = time.time()
         sleep_duration = 0.5
         time_slept = 0
         while True:
-            if time.time() >= ending:
+            if time.time() - start_time < timeout:
```
aronatkins, line 314: This is backwards, yes? As written, you're saying "if elapsed time is less than the timeout, err". Since timeout defaults to maxsize, that will always be true.

Do you have a test to confirm this reading?

aronatkins (changes requested):

aronatkins (approved):
Would love to see tests exercising the "took too long" task polling.

tdstein:
> Would love to see tests exercising the "took too long" task polling.

Thanks! Yes, I will clean this up and make it more robust. I converted the pull request back to draft, so feel free to remove yourself if you are still getting notifications. I will add you back when I am ready for final review.

PR #446, opened 2023-07-06
Adjust formatting for `system caches delete` option descriptions

toph-allen:
## Intent

Resolves two issues with the `system caches delete` documentation:

1. Options were described below the example, making them less discoverable.
2. Bullet points describing options lacked a preceding double-newline, so, while GitHub rendered the list correctly, the docs site did not. The `mkdocs` site rendered them as a single paragraph.

## Type of Change

Documentation-only.

- [ ] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change   

## Approach

N/A

## Automated Tests

N/A

## Directions for Reviewers

I checked the fix in the rendered `mkdocs` site, but it'd be good for you to do so as well. Within the `docs` folder, run `make all`, then navigate to `site/index.html` and check the output.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1081,7 +1081,15 @@ Connect server, as well as the language (`-l, --language`), version (`-V,
 delete. Deleting a large cache might take a while. The command will wait for
 Connect to finish the task.
 
-Use the dry run option (`-d, --dry-run`) to surface any errors ahead of
+Use the following parameters specify the target cache:
+
+- `language` (required) must name `R` or `Python`. It is case-insensitive.
+- `version` (required) must be a three-part version number, e.g. `3.8.12`.
+- `image-name` (optional) defaults to `Local`, which targets caches used for
+  natively-executed content. Off-host images can be specified using either the
+  literal image name or the sanitized name returned by the `list` command.
+
+Use the dry run flag (`-d, --dry-run`) to surface any errors ahead of
 deletion.
 
 ```bash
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -1081,7 +1081,15 @@ Connect server, as well as the language (`-l, --language`), version (`-V,
 delete. Deleting a large cache might take a while. The command will wait for
 Connect to finish the task.
 
-Use the dry run option (`-d, --dry-run`) to surface any errors ahead of
+Use the following parameters specify the target cache:
+
+- `language` (required) must name `R` or `Python`. It is case-insensitive.
+- `version` (required) must be a three-part version number, e.g. `3.8.12`.
+- `image-name` (optional) defaults to `Local`, which targets caches used for
+  natively-executed content. Off-host images can be specified using either the
```
aronatkins, line 1089: Always using four-space indentation is a good way to make sure that the different processors render the Markdown consistently.

aronatkins (approved):

toph-allen:
Not sure why CI is failing, but it seems likely to be outside of the scope of this PR.

tdstein:
I just started seeing the same error in another pull request. There may be a temporary error with the host. Let's see if it resolves itself shortly.

README.md
toph-allen: In this case, it was just that the list was not in its own paragraph, i.e. not separated from the previous line by an empty line. I don't think indentation would do anything here? Anyway, it seems to be rendering correctly, so I'm going to resolve this conversation.

toph-allen:
Looks like the other issue that was breaking this is https://github.com/rstudio/vetiver-python/issues/184 — `vetiver` did not specify a `pydantic` version in its `setup.cfg` so a major `pydantic` release was breaking `vetiver`.

tdstein (approved):

toph-allen:
Fixed CI failures by adding `pydantic<2.0.0` to `vetiver-requirements.txt`. There is also a fix inbound in Vetiver; see https://github.com/rstudio/vetiver-python/issues/184.

PR #457, opened 2023-07-27
doc: Remove warning about python 3.7

nealrichardson:
3.7 is EOL so we don't need to warn so prominently about it.


edavidaja (approved):
👍🏾

PR #461, opened 2023-08-04
Client flags for writing `environment.environment_management` fields to the bundle's manifest.json

dbkegley:
## Intent

Add `--disable-env-management`, `--disable-env-management-py` and `--disable-env-management-r` flags for all content types that support environment restores. These flags indicate to Connect that the user is responsible for Python/R package installation, and Connect should not install packages during the build. The Python/R packages must still be available in the runtime environment in order to run the content. This is especially useful if off-host execution is enabled when the execution environment (specified by `--image`) already contains the required packages. Requires Posit Connect `>=2023.07.0`.

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach

Adds 3 new flags for all sub-commands that might write a `manifest.json`:

- `--disable-env-management` - shorthand flag to disable both R and Python env management
- `--disable-env-management-r` - disable only R env management
- `--disable-env-management-py` - disable only Python env management.

All 3 flags are Null-able and default to `None`.  None indicates that no value should be written to the manifest.
Disabling environment management means that the respective value in the manifest should be `False`

>**Note:** There is currently no way to explicitly "enable" env management. "enabled" is the current default behavior so users should rely on `None` to mean enabled. This is a potential shortcoming of the current `--disable`-style flag implementation.

Because of the nature of CLI flags (present == True), we use a callback to invert the value of the flag.  The value is then stored on a non-negative variable which we pass around as an argument. (e.g., `--disable-env-management-py` == `env_managment_py=False`). This logic is controlled by the `func env_management_callback(ctx, param, value)`, which is shared by `--disable-env-management-r` and `--disable-env-management-py`.

All 3 new flags are added via the new `@runtime_environment_args` helper annotation. The existing `--image` flag was also added to this helper annotation.  All `@runtime_environment_args` flags are used to write values to the `environment` field in the bundle's `manifest.json`

## Automated Tests

- Updated/Added manifest-writer tests in `bundles_test.py`

## Directions for Reviewers

We should validate all commands that write manifests have `--disable-env-management` flags in the help options. These include (flags listed for `api` are also shared by `fastapi,dash,streamlit,bokeh,shiny` so those don't need to be validated separately):

```
rsconnect write-manifest api       --help  
rsconnect write-manifest notebook  --help  
rsconnect write-manifest quarto    --help  
rsconnect write-manifest voila     --help

rsconnect deploy api       --help  
rsconnect deploy notebook  --help  
rsconnect deploy quarto    --help  
rsconnect deploy voila     --help
```

There are 3 helpers that write manifests in `bundle.py`:

`make_html_manifest()`
`make_source_manifest()`
`class Manifest.__init__()`

The 3 helpers above can all be exercised by one of the following commands. They should be run with different combinations of the 3 `--disable-env-management` flags to make sure the right values are written to the manifest.

- `rsconnect write-manifest api` - calls `make_source_manifest()`
- `rsconnect write-manifest voila ` - calls `class Manifest.__init__()`
- `rsconnect deploy notebook --static` - calls `make_html_manifest()`

When testing `rsconnect deploy notebook --static`, the resulting manifest will only be available on the connect server in the data dir

## Checklist

### DIFF
--- a/rsconnect/actions.py
+++ b/rsconnect/actions.py
@@ -540,10 +540,10 @@ def write_quarto_manifest_json(
     :param extra_files: Any extra files to include in the manifest.
     :param excludes: A sequence of glob patterns to exclude when enumerating files to bundle.
     :param image: the optional docker image to be specified for off-host execution. Default = None.
-    :param env_management_py: False indicates that the user is responsible for Python package installation
-        in the runtime environment. Default = None.
-    :param env_management_r: False indicates that the user is responsible for R package installation
-        in the runtime environment. Default = None.
+    :param env_management_py: False prevents Connect from managing the Python environment for this bundle.
+        The server administrator is responsible for installing packages in the runtime environment. Default = None.
+    :param env_management_r: False prevents Connect from managing the R environment for this bundle.
+        The server administrator is responsible for installing packages in the runtime environment. Default = None.
     """
     warn("This method has been moved and will be deprecated.", DeprecationWarning, stacklevel=2)
 
@@ -661,10 +661,10 @@ def deploy_jupyter_notebook(
     :param hide_tagged_input: If True, will hide input code cells with the 'hide_input' tag when rendering
     output. Previous default = False.
     :param image: the optional docker image to be specified for off-host execution. Default = None.
-    :param env_management_py: False indicates that the user is responsible for Python package installation
-        in the runtime environment. Default = None.
-    :param env_management_r: False indicates that the user is responsible for R package installation
-        in the runtime environment. Default = None.
+    :param env_management_py: False prevents Connect from managing the Python environment for this bundle.
+        The server administrator is responsible for installing packages in the runtime environment. Default = None.
+    :param env_management_r: False prevents Connect from managing the R environment for this bundle.
+        The server administrator is responsible for installing packages in the runtime environment. Default = None.
     :return: the ultimate URL where the deployed app may be accessed and the sequence
     of log lines.  The log lines value will be None if a log callback was provided.
     """
@@ -867,10 +867,10 @@ def deploy_python_api(
     If a log callback is provided, then None will be returned for the log lines part
     of the return tuple. Previous default = None.
     :param image: the optional docker image to be specified for off-host execution. Default = None.
-    :param env_management_py: False indicates that the user is responsible for Python package installation
-        in the runtime environment. Default = None.
-    :param env_management_r: False indicates that the user is responsible for R package installation
-        in the runtime environment. Default = None.
+    :param env_management_py: False prevents Connect from managing the Python environment for this bundle.
+        The server administrator is responsible for installing packages in the runtime environment. Default = None.
+    :param env_management_r: False prevents Connect from managing the R environment for this bundle.
+        The server administrator is responsible for installing packages in the runtime environment. Default = None.
     :return: the ultimate URL where the deployed app may be accessed and the sequence
     of log lines.  The log lines value will be None if a log callback was provided.
     """
@@ -917,10 +917,10 @@ def deploy_python_fastapi(
         If a log callback is provided, then None will be returned for the log lines part
         of the return tuple. Previous default = None.
         :param image: the optional docker image to be specified for off-host execution. Default = None.
-        :param env_management_py: False indicates that the user is responsible for Python package installation
-        in the runtime environment. Default = None.
-        :param env_management_r: False indicates that the user is responsible for R package installation
-        in the runtime environment. Default = None.
+        :param env_management_py: False prevents Connect from managing the Python environment for this bundle.
+        The server administrator is responsible for installing packages in the runtime environment. Default = None.
+        :param env_management_r: False prevents Connect from managing the R environment for this bundle.
+        The server administrator is responsible for installing packages in the runtime environment. Default = None.
         :return: the ultimate URL where the deployed app may be accessed and the sequence
         of log lines.  The log lines value will be None if a log callback was provided.
     """
@@ -1009,10 +1009,10 @@ def deploy_dash_app(
     If a log callback is provided, then None will be returned for the log lines part
     of the return tuple. Previous default = None.
     :param image: the optional docker image to be specified for off-host execution. Default = None.
-    :param env_management_py: False indicates that the user is responsible for Python package installation
-        in the runtime environment. Default = None.
-    :param env_management_r: False indicates that the user is responsible for R package installation
-        in the runtime environment. Default = None.
+    :param env_management_py: False prevents Connect from managing the Python environment for this bundle.
+        The server administrator is responsible for installing packages in the runtime environment. Default = None.
+    :param env_management_r: False prevents Connect from managing the R environment for this bundle.
+        The server administrator is responsible for installing packages in the runtime environment. Default = None.
     :return: the ultimate URL where the deployed app may be accessed and the sequence
     of log lines.  The log lines value will be None if a log callback was provided.
     """
@@ -1059,10 +1059,10 @@ def deploy_streamlit_app(
     If a log callback is provided, then None will be returned for the log lines part
     of the return tuple. Previous default = None.
     :param image: the optional docker image to be specified for off-host execution. Default = None.
-    :param env_management_py: False indicates that the user is responsible for Python package installation
-        in the runtime environment. Default = None.
-    :param env_management_r: False indicates that the user is responsible for R package installation
-        in the runtime environment. Default = None.
+    :param env_management_py: False prevents Connect from managing the Python environment for this bundle.
+        The server administrator is responsible for installing packages in the runtime environment. Default = None.
+    :param env_management_r: False prevents Connect from managing the R environment for this bundle.
+        The server administrator is responsible for installing packages in the runtime environment. Default = None.
     :return: the ultimate URL where the deployed app may be accessed and the sequence
     of log lines.  The log lines value will be None if a log callback was provided.
     """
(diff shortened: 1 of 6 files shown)


rsconnect/main.py
```diff
@@ -244,6 +244,58 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+# inverts bool args if they are provided, otherwise returns None
```
toph-allen, line 247: Doesn't just invert the bool args, but also sets them from the shorthand/"both" flag if that's present.

rsconnect/bundle.py
```diff
@@ -118,10 +120,17 @@ def __init__(
...
+        if image or env_management_py is not None or env_management_r is not None:
+            self.data["environment"] = {}
+            if image:
+                self.data["environment"]["image"] = image
+            if env_management_py is not None or env_management_r is not None:
+                self.data["environment"]["environment_management"] = {}
+                if env_management_py is not None:
+                    self.data["environment"]["environment_management"]["python"] = env_management_py
+                if env_management_r is not None:
+                    self.data["environment"]["environment_management"]["r"] = env_management_r
```
toph-allen, line 133: This block seems to be present 3 (as far as I've seen) times. Is it worth putting it in a small helper function e.g. `make_environment_data(image, env_management_py, env_management_r)`? I don't have the larger context of this file and how manifest generation is factored out so it may not make sense, but worth asking.

rsconnect/main.py
```diff
@@ -244,6 +244,58 @@ def wrapper(*args, **kwargs):
...
+        default=None,
+        help="Shorthand to disable environment management for both Python and R.",
+    )
+    @click.option(
+        "--disable-env-management-py",
+        "env_management_py",
+        is_flag=True,
+        default=None,
+        help="Disable Python environment management. Connect will not perform any Python package installation. "
+        "It is the user's responsibility to ensure required packages are installed in the runtime environment.",
```
toph-allen, line 280: I think there are parts of this which might be unclear to end users. Going to think through it in this comment.

> Disable Python environment management. Connect will not perform any Python package installation. It is the user's responsibility to ensure required packages are installed in the runtime environment.

- "the user's responsibility" — actually probably the server admin's responsibility
- "the runtime environment" — might be unclear for users without a clear mental model of the runtime environment on the Connect server that we're talking about, who might think this is referring to their local runtime environment.

Maybe something like 

> Disable Python environment management for this bundle. Connect will not create an environment or instal packages. An administrator must install any required packages in the correct Python environment on the Connect server.

And similar for the R help.

rsconnect/actions.py
```diff
@@ -538,6 +540,10 @@ def write_quarto_manifest_json(
     :param extra_files: Any extra files to include in the manifest.
     :param excludes: A sequence of glob patterns to exclude when enumerating files to bundle.
     :param image: the optional docker image to be specified for off-host execution. Default = None.
+    :param env_management_py: False indicates that the user is responsible for Python package installation
```
toph-allen, line 543:
```suggestion
    :param env_management_py: False prevents Connect from managing the Python environment for this bundle. The server administrator is responsible for installing packages in runtime environment on the server.
```

Something like that maybe? I wanna get away from saying it's the *user's* responsibility, because users likely won't have permission to do it.

rsconnect/bundle.py
```diff
@@ -656,7 +693,7 @@ def make_notebook_html_bundle(
         bundle_add_buffer(bundle, filename, output)
 
         # manifest
-        manifest = make_html_manifest(filename, image)
+        manifest = make_html_manifest(filename, image, env_management_py, env_management_r)
```
toph-allen, line 696: It's a pity that there are _so_ many functions that this change needs to affect. Not for this PR, but I wonder if there's any way to reduce repetition here, or if it's already as clean as it can be. Just adds more surface area for errors, I feel like. ¯\\\_(ツ)\_/¯

rsconnect/main.py
```diff
@@ -1443,6 +1492,9 @@ def write_manifest_notebook(
     file,
     extra_files,
     image,
+    disable_env_management,
```
toph-allen, line 1495: Checking my understanding: this var isn't actually needed in the function body because the callback that consumes it has already been called before line 1501?

tests/test_bundle.py
```diff
@@ -554,20 +563,76 @@ def test_make_source_manifest(self):
...
-        manifest = make_source_manifest(AppModes.PYTHON_API, None, None, None, "rstudio/connect:bionic")
+        manifest = make_source_manifest(AppModes.PYTHON_API, None, None, None,
+                                        image="rstudio/connect:bionic")
         self.assertEqual(
             manifest,
             {
                 "version": 1,
                 "metadata": {"appmode": "python-api"},
-                "environment": {"image": "rstudio/connect:bionic"},
+                "environment": {
```
toph-allen, line 580: Do we wanna include a test for one of these that uses the shorthand flag?

toph-allen (approved):
This generally looks good! My comments are just nitpicks or thoughts I had in the moment.

I didn't run through the instructions for exercising the code — I'm assuming we'll perform QA on this separately.

There are some areas of the code where I don't have insight into the full context so bear that in mind. Another review wouldn't hurt.

toph-allen:
Also, CI failed. It's the integration tests against full Connect again. I'm not sure what would have changed between the last PR and this one to reintroduce this failure. It could just be a slow container again. It doesn't seem like it's related to these changes though, I don't think.

rsconnect/bundle.py
dbkegley: Completely agree. It would be nice if we relied more on **kwargs throughout. There are a couple places we do this already with **locals() but it’d be nice to be more consistent

rsconnect/main.py
dbkegley: You're right that it's only required by the callback function. `click` will complain if it doesn't know how to invoke `write_manifest_notebook` which is why it needs to be included in the function signature even though it isn't used.

```
TypeError: write_manifest_notebook() got an unexpected keyword argument 'disable_env_management'
```

rsconnect/main.py
```diff
@@ -244,6 +244,58 @@ def wrapper(*args, **kwargs):
...
+def env_management_callback(ctx, param, value):
+    # eval the shorthand flag if it was provided
+    disable_env_management = ctx.params.get('disable_env_management')
+    if disable_env_management is not None:
+        value = disable_env_management
+
+    # invert env_management value if it was provided
+    if value is not None:
+        return not value
+    return value
```
tdstein, line 257: Is this equivalent?

```suggestion
    value = ctx.params.get('disable_env_management')
    return (not value) if (value != None) else value
```

rsconnect/main.py
```diff
@@ -244,6 +244,58 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+# inverts bool args if they are provided, otherwise returns None
+def env_management_callback(ctx, param, value):
```
tdstein, line 248: nit; type hinting is always appreciated.

```suggestion
def env_management_callback(ctx, param, value) -> Optional[bool]:
```

rsconnect/main.py
dbkegley: Not quite; `value` is either `--disable-env-management-r` or `--disable-env-management-py`, so we can't re-assign until we know if `ctx.params.get('disable_env_management') is None`. We also need to check whether `value is None` so that we can handle the null case which means that nothing should be written to the manifest.

```
$ rsconnect write-manifest api --disable-env-management-py ./
env_management_py:  None
env_management_r:  None
$ rsconnect write-manifest api --disable-env-management ./
env_management_py:  False
env_management_r:  False
```

rsconnect/main.py
dbkegley: I need to make this a habit, thanks for the reminder!

rsconnect/main.py
tdstein: Ah, I overlooked `value` being a method argument. Makes sense. Thanks!

rsconnect/main.py
dbkegley: I like this; I think the "for this bundle" distinction is also really important. Since the setting in the `manifest.json`, and thus the bundle, have the highest precedence it makes sense to explicitly say that these changes must be applied to _every bundle_ for them to take effect.  

If the user wants to apply these changes to all bundles then they'll need to use the server API to set an environment management strategy at the Application-level.

tests/test_bundle.py
dbkegley: I added a test case in `main_test.py` that covers the callback func. It's not exactly what you're asking for but it's pretty close without actually having to print/parse a manifest.

7e3b4c68139e26efdf9cc7ff201e17d9212cc183

ChaitaC:
verified on `kegs-skip-restore`

- [ ] Validate `--help` Options
```
rsconnect write-manifest api       --help  
rsconnect write-manifest notebook  --help  
rsconnect write-manifest quarto    --help  
rsconnect write-manifest voila     --help

rsconnect deploy api       --help  
rsconnect deploy notebook  --help  
rsconnect deploy quarto    --help  
rsconnect deploy voila     --help
```

all commands now include the following flags 
```
--disable-env-management     Shorthand to disable environment management for
                               both Python and R.
  --disable-env-management-py  Disable Python environment management for this
                               bundle. Connect will not create an environment
                               or install packages. An administrator must
                               install the  required packages in the correct
                               Python environment on the Connect server.
  --disable-env-management-r   Disable R environment management for this
                               bundle. Connect will not create an environment
                               or install packages. An administrator must
                               install the  required packages in the correct R
                               environment on the Connect server.
```


- [ ] run with different combinations of the 3 --disable-env-management flags to ensure the right values are written to the manifest.

`rsconnect write-manifest bokeh --overwrite --disable-env-management-r .`
```
 "environment": {
    "environment_management": {
      "r": false
    }
```

`rsconnect write-manifest bokeh --overwrite --disable-env-management-py .`
```
  "environment": {
    "environment_management": {
      "python": false
    }
```

`rsconnect write-manifest bokeh --overwrite --disable-env-management .`
```
 "environment": {
    "environment_management": {
      "python": false,
      "r": false
    }
```

ChaitaC:
- [ ] Verified Deploy and manifest options with all 3 flags 
example: 

#### rsconnect deploy notebook --static
- `rsconnect deploy notebook --disable-env-management --new  --server https://rsc.radixu.com/ --api-key <you_api_key>  ./stock-report-jupyter.ipynb quandl-wiki-tsla.json.gz`
`rsconnect deploy notebook --disable-env-management-r --new  --server https://rsc.radixu.com/ --api-key <you_api_key>  ./stock-report-jupyter.ipynb quandl-wiki-tsla.json.gz`
`rsconnect deploy notebook --disable-env-management-py --new  --server https://rsc.radixu.com/ --api-key <you_api_key>  ./stock-report-jupyter.ipynb quandl-wiki-tsla.json.gz`

- `rsconnect write-manifest notebook --disable-env-management-py --overwrite ./stock-report-jupyter.ipynb quandl-wiki-tsla.json.gz`
`rsconnect write-manifest notebook --disable-env-management-r ./stock-report-jupyter.ipynb quandl-wiki-tsla.json.gz`
`rsconnect write-manifest notebook --disable-env-management ./stock-report-jupyter.ipynb quandl-wiki-tsla.json.gz`


#### rsconnect write-manifest api 
#### rsconnect write-manifest voila

PR #480, opened 2023-09-12
Update http_support.py with NO_PROXY support

sergeytalash:
## Intent

New versions of rsconnect-python fail if the server is located in an intranet and the host has a proxy being enabled.
These changes include support of the NO_PROXY environment variable.

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach

-

## Automated Tests

-

## Directions for Reviewers

The command to check: rsconnect add -n name -s server

## Checklist

- [ ] I have updated [CHANGELOG.md](../CHANGELOG.md) to cover notable changes.
- [x] I have updated all related GitHub issues to reflect their current state.

PR #488, opened 2023-10-03
Add NO_PROXY support

### LINKED ISSUE #400: rsconnect-python doesn't support no_proxy variable
Dear Team,

I'm raising the issue regarding the feature request.

### The issue
We are observing that `rsconnect-python` package doesn't support `no_proxy` environment variables.

### Desired feature
`rsconnect-python` package should support both `HTTP_PROXY`, `HTTPS_PROXY` and `no_proxy` variables in order to be able to access external and internal resources from the same environment.

### Justification
It is important for big production environments that restrict access to internal services with configured proxy variables. Right now users should unset `HTTPS_PROXY` before deployment, 

### Additional notes
I see that `rsconnect-python` ignores `https_proxy` and `https_proxy` lowercase variables. Is it a feature or it is better to fix this behaviour?

Thank you for your consideration, hope to hear from you soon!

Best regards,
Yulia

mmarchetti:
This PR adds support for NO_PROXY. It's based on #480, but on a local branch.

## Intent
Fixes #400 

## Type of Change

- [x] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Directions for Reviewers
Deploy content:
* Without a proxy
* With a proxy: Proxy should be used.
* With a proxy, and NO_PROXY set but not matching the host. Proxy should be used.
* With a proxy, and NO_PROXY set and matching the host: Proxy should NOT be used.

## Checklist


rsconnect/http_support.py
```diff
@@ -75,7 +75,12 @@ def _create_ssl_connection(host_name, port, disable_tls_check, ca_data):
     """
     if ca_data is not None and disable_tls_check:
         raise ValueError("Cannot both disable TLS checking and provide a custom certificate")
-    _, _, proxyHost, proxyPort = _get_proxy()
+
+    no_proxy = os.environ.get("no_proxy", os.environ.get("NO_PROXY", "#"))
```
tdstein, line 79: Why is the default value `'#'`?

mmarchetti: It won't ever match a hostname. Since the check is `endswith`, an empty string would be interpreted as matching all hosts.

tdstein (approved):

PR #492, opened 2023-10-11
Infer entrypoints from additional filename patterns

jcheng5:
Currently, rsconnect-python will infer a variety of entrypoints if none is specified by the user, depending on the files in the directory. In priority order, these are:

- app.py
- application.py
- main.py
- api.py
- If there's only a single .py file in the directory, use it

To the end of this list, I'd like to add:

- If there's one and only one file matching one of [`app-*.py`, `app_*.py`, `*-app.py`, `*_app.py`], use it and append `:app` (whereas the others assume the attribute is the same as the filename).

I've also made two other changes:

1. If no entrypoint is detected, don't assume `app:app` but raise a fatal error.
2. The `actions` module had a deprecated `get_default_entrypoint` implementation that was identical to the `bundle` one. I now have `actions.get_default_entrypoint` delegating to `bundle.get_default_entrypoint` (after raising its deprecation warning).

## Intent
Shiny for Python has some upcoming scenarios which will make a single `foo-app.py` much more common in the future. We'd like to support deploying these without requiring an explicit `--entrypoint` argument.

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [x] Breaking Change   

The breaking change is that we would no longer default to `app:app` if inference fails, but raise an error. I can't think of a case where it's preferable to go through with a whole deployment that's definitely not going to work. (It's definitely not going to work, right?)

## Approach
Pretty much copied what was already there.

## Automated Tests
I added unit tests that cover both my new file patterns, and some already-existing cases that had not been covered. I also refactored the relevant unit test to make it easier to add more cases.

## Directions for Reviewers
In an environment with `rsconnect-python` and `shiny` installed:

```
mkdir testapp
cd testapp
shiny create
mv app.py example-app.py
rsconnect deploy shiny .
```

## Checklist

### DIFF
--- a/rsconnect/bundle.py
+++ b/rsconnect/bundle.py
@@ -1404,6 +1404,10 @@ def validate_manifest_file(file_or_directory):
     return file_or_directory
 
 
+re_app_prefix = re.compile(r"^app[-_].+\.py$")
+re_app_suffix = re.compile(r".+[-_]app\.py$")
+
+
 def get_default_entrypoint(directory):
     candidates = ["app", "application", "main", "api"]
     files = set(os.listdir(directory))
@@ -1417,6 +1421,12 @@ def get_default_entrypoint(directory):
     python_files = list(filter(lambda s: s.endswith(".py"), files))
     if len(python_files) == 1:
         return python_files[0][:-3]
+
+    # try app-*.py, app_*.py, *-app.py, *_app.py
+    app_files = list(filter(lambda s: re_app_prefix.match(s) or re_app_suffix.match(s), python_files))
+    if len(app_files) == 1:
+        # In these cases, the app should be in the "app" attribute
+        return app_files[0][:-3] + ":app"
 
     logger.warning("Can't determine entrypoint; defaulting to 'app'")
     return "app"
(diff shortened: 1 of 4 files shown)


rsconnect/bundle.py
```diff
@@ -1418,8 +1422,13 @@ def get_default_entrypoint(directory):
     if len(python_files) == 1:
         return python_files[0][:-3]
 
-    logger.warning("Can't determine entrypoint; defaulting to 'app'")
-    return "app"
+    # try app-*.py, app_*.py, *-app.py, *_app.py
+    app_files = list(filter(lambda s: re_app_prefix.match(s) or re_app_suffix.match(s), python_files))
+    if len(app_files) == 1:
+        # In these cases, the app should be in the "app" attribute
+        return app_files[0][:-3] + ":app"
```
mmarchetti, line 1429: If you omit the attribute name, Connect will use the Flask import rules to look for common attribute names in the imported module (`app`, `application`, or the `create_app` or `make_app` factory functions).

Connect will also strip the `.py` suffix from the entrypoint module.

mmarchetti (approved):

rsconnect/bundle.py
jcheng5: OK, thanks! It looks like this docstring might be wrong then? If so, I can update it. https://github.com/rstudio/rsconnect-python/blob/f4def0c590fef96f6ee8aa8ea31512629b94a50c/rsconnect/bundle.py#L1428C60-L1429C74

mmarchetti: Yeah, that docstring looks wrong.

PR #495, opened 2023-10-16
Include API key with initial server check

### LINKED ISSUE #487: Include API key in server detection
When adding a new server, rsconnect-python first checks the unauthenticated /server_settings endpoint to see if the new server is actually a Connect server, then uses an authenticated endpoint to see if the API key is valid. 

Some customers have authenticating proxies in front of Connect, and that prevents access to the unauthenticated endpoint. We should only access the authenticated endpoint, and handle 401 status from that call to mean that the API key is invalid.

mmarchetti:
## Intent
Fixes #487 

## Type of Change

- [X] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change   

## Approach
The initial server check accesses /server_settings to verify that the target is a Connect server. That endpoint doesn't require authentication, so the API key was omitted. However, Connect servers using proxy authentication sometimes have the proxy configured to only pass requests that include an authentication header. This change allows rsconnect-python to work in those configurations.

## Automated Tests
This change is in an area of code that does not include tests.

## Directions for Reviewers
Add a new Connect server and verify that the initial request to /server_settings includes an Authorization header containing the provided API key. A couple ways to do this:
* Go through a proxy that logs requests including headers
* Add a log/print statement [here](https://github.com/rstudio/rsconnect-python/blob/api-key-in-server-check/rsconnect/http_support.py#L301)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - The `https_proxy` environment variable is recognized as a synonym for
   `HTTPS_PROXY`.
+- When adding a new server, the initial request now includes an
+  Authorization header containing the API key. This is needed
+  for Connect installations behind a proxy that only passes
+  authenticated requests.
 
 ### Added
 - Added support for the `no_proxy` or `NO_PROXY` environment variables to specify
--- a/rsconnect/main.py
+++ b/rsconnect/main.py
@@ -351,7 +351,7 @@ def _test_server_and_api(server, api_key, insecure, ca_cert):
     me = None
 
     with cli_feedback("Checking %s" % server):
-        real_server, _ = test_server(api.RSConnectServer(server, None, insecure, ca_data))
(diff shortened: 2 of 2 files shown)


tdstein (approved):
Looks good.

I see 

```console
$rsconnect add -n dogfood -k adsf -s https://rsc.radixu.com
...
Checking https://rsc.radixu.com...               /__api__/server_settings {'User-Agent': 'rsconnect-python/1.20.1.dev21+gdc8094f', 'Authorization': 'Key adsf'}
```

kgartland-rstudio:
Verified.

Printed out the headers as in the directions here and this branch now includes Authorization Headers in the server check step:

Previous output:
```
Checking https://connect.localtest.me/rsc/dev-pam/... {'User-Agent': 'rsconnect-python/1.20.1.dev23+g07ce90b.d20231017'}
```

Current output:
```
Checking https://connect.localtest.me/rsc/dev-pam/... {'User-Agent': 'rsconnect-python/1.20.1.dev25+g62b41b4.d20231017', 'Authorization': 'Key {API_KEY}'}
```

PR #511, opened 2023-10-19
Verify deployments

### LINKED ISSUE #503: Verify deployments
When deploying an app/api, access the app's direct URL at the end of deployment to verify that the app can start up successfully. Connect does not start a worker until there's an access, unless MinWorkers > 0.

When performing that access, tolerate certain HTTP errors - for example, the `/` endpoint in an API might expect a POST rather than a GET, or might expect parameters that aren't provided. But if you get back a 5xx code, that indicates a problem.

Also provide a `--no-verify` flag to allow opting out of this new behavior.

mmarchetti:
This PR introduces a new feature that verifies that content items are live and accessible after deployment. This is most useful for apps and APIs. Accessing the content causes Connect to launch a worker process to service the request; this will catch apps that fail to start up because of import errors or exceptions raised on import.

There is a new CLI flag `--no-verify` to prevent this behavior in cases where it's not wanted (for example, if app startup is expensive or might have side effects).

## Intent

Fixes #503

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Automated Tests
There are integration tests that cover the new behavior and the flag.

## Directions for Reviewers
Deploy all kinds of content, including some apps that fail to start. Check that the verification catches those. Also deploy with `--no-verify` and check that even faulty apps can be deployed without rsconnect detecting the error.

Also deploy to a server with a path prefix (like Yeti or Colorado), since the path prefix is used when accessing the app.

## Checklist


rsconnect/api.py
```diff
@@ -195,6 +195,25 @@ def app_publish(self, app_id, access):
...
 
+    def is_app_failed_response(self, response):
+        return isinstance(response, HTTPResponse) and response.status >= 500
+
+    def app_access(self, app_guid):
+        method = "GET"
+        base = dirname(self._url.path)  # remove __api__
+        path = f"{base}/content/{app_guid}/"
+        response = self._do_request(method, path, None, None, 3, {}, False)
+        print(response.status, end=None)
```
tdstein, line 206: Are these print statements still needed?

mmarchetti: Good catch, thanks!

tdstein (approved):

kgartland-rstudio:
Verified on a number of apps.

```
Launching Dash application...
         Dashboard content URL: https://rsc.radixu.com/connect/#/apps/54798b06-793a-4fe5-9ffc-28861b6b16ec
         Direct content URL: https://rsc.radixu.com/content/54798b06-793a-4fe5-9ffc-28861b6b16ec/
Verifying deployed content...   [ERROR]: Could not access the deployed content. The app might not have started successfully. Visit it in Connect to view the logs.
```

It also found a potential bug! When deploying a sanic api to Connect, we don't display anything at the root directory but we can hit the endpoint via curl so our current testing has been passing.  I may need to set `--no-verify` on those tests until we get that fixed.  Other than that I think this looks good. I'll have a PR out on the Connect side to avoid failures in the meantime.

kgartland-rstudio:
Also concerning the issue I hit above, I wonder if we want to make `no-verify` the default behavior and have a flag for `--verify` so we don't suddenly impact customer's current workflow.

kgartland-rstudio:
One more note...`sanic` actually _does_ eventually load but the sainc-routing can take its time. So we're really calling something a failure that will eventually succeed. I'll test this further with longer running apps and notebooks. 

We may want to hold off on this or adjust the default behavior as I mentioned.

nealrichardson:
Personally, I think there are lots of benefits to verifying such that it should be the default. But if we wanted to do a release now where it is opt-in, and then change the default to opt-out in a following release after we've had more time to catch issues, that seems reasonable.

mmarchetti:
The verify check should honor `CONNECT_REQUEST_TIMEOUT`, so there is a way to control how patient it is when waiting for the app to come up/respond.

kgartland-rstudio:
How about we adjust the failure from an ERROR to a WARNING:

```
#   Verifying deployed content... 	[ERROR]: Could not access the deployed content. The app might not have started successfully. Visit it in Connect to view the logs.
#   Error: Could not access the deployed content. The app might not have started successfully. Visit it in Connect to view the logs.
```

mmarchetti:
> I'll test this further with longer running apps and notebooks.

Notebooks should be OK, since we wait until the rendering is complete so the "verify" step is loading the rendered notebook HTML. You would need to set `CONNECT_TASK_TIMEOUT` if the total deployment process, including rendering the notebook, will take >24 hours (the default).

kgartland-rstudio:
Okay all looks good.  Also looks like my `sanic` issue is only occuring in the Dockerized Connect instance, it works as expected on dogfood.  We can merge this and then i'll add the `--no-verify` flag for the `sanic` deploy in Connect to avoid failures in CI.

PR #521, opened 2023-11-13
Add support for shiny express apps

wch:
## Intent

This PR adds support for Shiny express apps.

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach

The main change was pretty straightforward. It involves copying the contents of one file from the shiny package, and then adding three lines of code that change the `entrypoint` and set an environment variable.

However, there were more changes that I made, with some help from @mmarchetti. The use of `kwargs = locals()`, and passing the `kwargs` along to `RSConnectExecutor(**kwargs)` was confusing and caused some problems, so I made the arguments explicit.

## Automated Tests

I didn't add any tests.

## Directions for Reviewers

This will require a close look to make sure that the `extra_args` being passed to `RSConnectExecutor()` contain the correct items. With the previous code, it was extremely difficult to tell which arguments from the `kwargs` were meant to go to `RSConnectExecutor()`, and which ones just happened to be along for the ride.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+### Added
+- Added support for deploying Shiny Express applications.
+
 ## [1.21.0] - 2023-10-26
 
 ### Fixed
--- a/rsconnect/main.py
+++ b/rsconnect/main.py
@@ -80,7 +80,7 @@ from .json_web_token import (
     produce_bootstrap_output,
     parse_client_response,
 )
-from .shiny_express import is_express_app
+from .shiny_express import escape_to_var_name, is_express_app
 
 server_store = ServerStore()
 future_enabled = False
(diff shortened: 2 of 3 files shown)


wch:
That there is a type-checking error in CI -- this issue actually existed in the code before but wasn't picked up on by the type checker because it was obscured by passing in the `kwargs`; now that the arguments are explicit, the type checker is seeing the problem.

The problem is that in `deploy_app()`, `cacert` is typed as `IO[Any]`, but then it is passed to `RSConnectExecutor()`, and the class constructor there expects `cacert` to be a `str`.

Looking at the code, I suspect that it should be `str` in both places, but I don't know for sure.

mmarchetti (approved):

wch:
I also didn't update the changelog because there wasn't a section since the previous release. Should I add one?

karangattu:
Verified that deployments for shiny express app works as expected for **connect**, **shinyapps.io** and **posit.cloud**.
Connect url for deployed app - [link](https://rsc.radixu.com/content/e6f1392e-3bcd-47db-97c0-7da98e7dd24a/)
Shinyapps.io url for deployed app - [link](https://nafcillincat.shinyapps.io/shiny-express/)
Posit.cloud url for deployed app - [link](https://posit.cloud/content/7087624)

karangattu:
Tests for deploying shiny express apps via the rsconnect-python package to connect will be added in the next week.

PR #525, opened 2023-11-30
Include ctx in executor init

mmarchetti:
## Intent
PR #521 cleaned up the executor parameters (on my recommendation). However, we lost the `ctx` parameter, and this causes parameter source checking to return `<unknown source>` instead of the parameter source.

## Type of Change

- [x] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change

### DIFF
--- a/rsconnect/main.py
+++ b/rsconnect/main.py
@@ -1439,6 +1439,7 @@ def generate_deploy_python(app_mode: AppMode, alias: str, min_version: str, desc
         )
 
         ce = RSConnectExecutor(
+            ctx=ctx,
             name=name,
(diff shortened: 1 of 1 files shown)


mmarchetti:
As expected: 
```
$ rsconnect deploy api -A foo -n test connect/test/examples/python-flask
    Warning: the existing manifest.json file will not be used or considered.
Error: -n/--name (from COMMANDLINE) cannot be specified in conjunction with options -A/--account (from COMMANDLINE). See command help for further details.
```

kgartland-rstudio (approved):
LGTM

PR #527, opened 2023-12-01
Make it more unlikely for content builds to get stuck in a running state

### LINKED ISSUE #467: Client has incorrect knowledge of server state when long-running build process is interrupted
I ran into this when I was trying to rebuild content on dogfood after migrating from ubuntu18 to ubuntu22. 

I followed our cache rebuild workflow listed here: https://docs.posit.co/connect/admin/server-management/runtime-caches/#step-2-rebuild-content

After collecting the content that needed to be rebuilt, I filtered down to approximately the last year of deployments. This produced a list of 923 guids which would be used in the rebuild step. 

I kicked off the rebuild on my local machine (following our docs) and let it run overnight. When I came back, I was seeing periodic connectivity errors being produced in stdout, like the following: 

```
[ERROR] 2023-08-23T18:09:51-0400 b2311f56-82f1-4e26-aee5-4fc606627d8b,38498 generated an exception: Exception trying to connect to https://rsc.radixu.com - [Errno 54] Connection reset by peer

[ERROR] 2023-08-23T20:08:25-0400 4d71e092-06da-4c1d-ac91-756314096272,38420 generated an exception: Received an unexpected response from Posit Connect (calling /__api__/v1/content/4d71e092-06da-4c1d-ac91-756314096272/build): 502 Bad Gateway
```

Assuming my local machine going into hibernation could have been the source of connectivity issues, I stopped the process and figured I could restart it, given that rsconnect-python intializes a local store that is tracking the remaining pending tasks. 

This fails with the following: 

```
➜  Downloads rsconnect content build run --name dogfood
[INFO] 2023-08-24T10:23:54-0400 Initializing ContentBuildStore for https://rsc.radixu.com
Error: There is already a build running on this server: https://rsc.radixu.com
```

Investigating the local store's json blob, I see the following flag:
```
    "rsconnect_build_running": true
```

And looking through the source code (cross-checked with @tdstein), I don't think the client attempts to sync its `rsconnect_build_running` flag at startup - it just looks at the current state of the local store. At this point, it has de-synced from the state of the server and isn't correct. 

This means that I can't continue to run the rebuild off of the current state of the store. It is always going to think that a build is currently running and fail out.

dbkegley:
## Intent

Interrupting a long-running `rsconnect content build run` command with `^C`
will now update the local state file before attempting graceful cleanup. This
should help prevent users from getting stuck a "build already running" state.

Fixes #467

## Type of Change

- [x] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach

- Set global build state before attempting graceful background thread shutdown. This may cause individual build status to still say "RUNNING"
- RUNNING builds can now be re-started with `content build run --running`
- Added a convenience flag `--retry`  that will start a build for all content marked as `RUNNING`, `ABORTED`, `NEEDS_BUILD` or `ERROR`
  - Marks the `--running`, `--aborted`, and `--error` flags as hidden. Most people will probably want to use `--retry` anyway but the more specific flags can still be used if needed.

## Automated Tests

Added a new test for `--retry` in `test_main_content.py` 

## Directions for Reviewers

> alternatively we can merge https://github.com/rstudio/rsconnect-python/pull/529 first and then these tests will run in CI

The tests in `test_main_content.py` require `mock_connect`
I would recommend updating `scripts/runtests` to say: `pytest ${PYTEST_ARGS} --mypy ./tests/test_main_content.py`
so that you can run only the content tests locally since they are skipped in CI.

```
# start mock_connect in a new terminal
cd mock_connect && make up

# run tests
CONNECT_CONTENT_BUILD_DIR="rsconnect-build-test" \
CONNECT_SERVER=http://localhost:3939 \
CONNECT_API_KEY="0123456789abcdef0123456789abcdef" \
make test
```

I've also created a followup to deprecate mock_connect. We should move to `httpretty` instead for mocking which is already used by `test_main.py` and part of this follow up we can start running these tests in CI again without mock_connect.

## Checklist


Makefile
```diff
@@ -55,7 +55,7 @@ mock-test-%: clean-stores
 	trap "$(MAKE) -C mock_connect down" EXIT; \
 	CONNECT_CONTENT_BUILD_DIR="rsconnect-build-test" \
 	CONNECT_SERVER="http://$(HOSTNAME):3939" \
-	CONNECT_API_KEY="0123456789abcdef0123456789abcdef" \
+	CONNECT_API_KEY="21232f297a57a5a743894a0e4a801fc3" \
```
zackverham, line 58: did we rotate credentials at some point? Just making sure I'm following why this change is needed

rsconnect/actions_content.py
```diff
@@ -172,13 +194,28 @@ def build_start(connect_server, parallelism, aborted=False, error=False, all=Fal
...
+        )
+        logger.info(
+            "To retry the content build, including items that were interrupted "
+            + "or failed, use: rsconnect content build run --retry"
+        )
     finally:
+        # make sure that we always mark the build as complete but note
+        # there's no guarantee that the content_executor or build_monitor
+        # were allowed to shut down gracefully, they may have been interrupted.
+        _content_build_store.set_build_running(False)
```
zackverham, line 214: with this setup, its still possible to ^C into a weird state, right? Just significantly less likely if we set the state of the build first thing.

tests/test_main_content.py
```diff
@@ -116,6 +117,51 @@ def test_build(self):
...
+        self.assertEqual(listing[0]["guid"], _content_guids[0])
+        self.assertEqual(listing[0]["bundle_id"], "176")
+        self.assertEqual(listing[0]["rsconnect_build_status"], BuildStatus.NEEDS_BUILD)
+
+        # set the content build status to RUNNING so it looks like it was interrupted
+        # and the cleanup did not have time to finish, otherwise it would be marked as ABORTED
+        store = ContentBuildStore(RSConnectServer(connect_server, api_key))
+        store.set_content_item_build_status(_content_guids[0], BuildStatus.RUNNING)
+
+        # run the build
```
zackverham, line 150: I don't think its a blocker or anything, but you could also add some tests for the other hidden flags that are considered when a retry is executed if you wanted to.

zackverham (approved):

zackverham:
do we need to document the new flag somewhere?

Makefile
dbkegley: That was a static api key that mock_connect was expecting. I changed it to the insecure admin api key (`echo -n admin | md5sum`) since that's what we use in dev, but this is removed in my next PR anyway

rsconnect/actions_content.py
dbkegley: Yep, that's right. I'm hoping this covers most cases. It could still happen on slow filesystems but this makes it less likely in most scenarios. If the problem persists after this change then we can look at removing the safety check.

The reason it exists in the first place is to stop users from modifying the state file while there's a build running. The 3 commands that modify the local start are `content build [run, add, rm]`.

zackverham: yeah, I completely agree with the approach here - just making sure I am following what's going on.

PR #536, opened 2024-01-18
Use requirements.txt in docker build

mmarchetti:
## Intent
This PR fixes the `make image-*` target, which currently fails because the `.[test]` is not installable within the docker build. Instead, we're using `pip install -r requirements.txt`.

## Type of Change

- [x] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change   

## Automated Tests
This is used in as part of running `make test` locally.

## Directions for Reviewers
```
make image-3.8
make test-3.8
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,4 +3,5 @@ FROM ${BASE_IMAGE}
 
 WORKDIR /rsconnect
 COPY scripts/build-image build-image
+COPY requirements.txt .
 RUN bash build-image && rm -vf build-image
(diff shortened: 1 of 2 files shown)


aronatkins (approved):
I totally missed that we still had a `requirements.txt`.

nit: The requirements file says:

```
# This file is just for Snyk scanning,
# because it doesn't understand pyproject.toml
# unless it uses Poetry.
# https://github.com/snyk/snyk-python-plugin/issues/147
```

PR #544, opened 2024-02-08
Fix Nightly CI Runs

kgartland-rstudio:
## Intent

Nightly CI runs failed the last two nights. The Cypress Docker image just needed to be updated.  I also noticed, while it wasn't causing the workflow to fail, the pre-release step was failing so that's fixed here too. 

Failed Run:
https://github.com/rstudio/rsconnect-python/actions/runs/7827139255

Passing Run from this branch:
https://github.com/rstudio/rsconnect-python/actions/runs/7831212084

## Type of Change

- [ ] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change   

## Checklist

### DIFF
--- a/integration-testing/docker/cypress.Dockerfile
+++ b/integration-testing/docker/cypress.Dockerfile
@@ -1,4 +1,4 @@
-FROM cypress/included:12.7.0
+FROM cypress/included:latest
 
 RUN apt-get update -y && apt-get install -y --no-install-recommends \
     jq \
--- a/integration-testing/docker/cypress.Dockerfile
+++ b/integration-testing/docker/cypress.Dockerfile
@@ -1,4 +1,4 @@
-FROM cypress/included:latest
+FROM cypress/included:13.6.4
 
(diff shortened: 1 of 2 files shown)


integration-testing/docker/cypress.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM cypress/included:12.7.0
+FROM cypress/included:latest
```
aronatkins, line 1: We have seen breaking changes with previous Cypress releases. They often require configuration or layout changes, especially across major versions. Are you against pinning to specific versions?

aronatkins (approved):

integration-testing/docker/cypress.Dockerfile
kgartland-rstudio: Oh not at all. I'll pin to the latest working version.

aronatkins: Thanks. That's essentially what we do elsewhere -- pin to a working version and then occasionally update after confirming that the new version doesn't break our use.

PR #545, opened 2024-02-15
Shiny starlette warning

wch:
This PR is a workaround for an incompatibility between Connect<2024.01.0 and starlette>=0.35.0.

## Intent

This PR implements the rsconnect-python part of https://github.com/posit-dev/py-shiny/issues/1114#issuecomment-1942757757.

It also updates many type annotations to satisfy pyright/pylance's strict mode, which in turn made it easier to implement the starlette-related changes.

## Type of Change

- [x] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change   

## Approach

The important changes are in main.py and in the new file utils_package.py. (Most of the rest of the changes are adding/updating type annotations.)

## Automated Tests

Many of the functions in utils_package.py have unit tests in tests/test_utils_package.py.


rsconnect/api.py
```diff
@@ -1,11 +1,14 @@
 """
 Posit Connect API client and utility functions
 """
+
+from __future__ import annotations
```
wch, line 5: On Python 3.7 and above, adding this import allows you to do the following:

- Use `list` (which does not require an import) instead of `List` (which requires `from typing import List`), and the same for `dict` and some others.\
- Use `A | B` instead of `Union[A, B]`.

rsconnect/api.py
```diff
@@ -661,7 +667,7 @@ def validate_rstudio_server(
                 raise RSConnectException("Failed to verify with {} ({}).".format(server.remote_name, exc))
 
     @cls_logged("Making bundle ...")
-    def make_bundle(self, func: Callable, *args, **kwargs):
+    def make_bundle(self, func: Callable[..., object], *args: object, **kwargs: object):
```
wch, line 670: I added the `object` type to a number of annotations, so that the places which call this function won't be flagged.

I used `object` because it makes the fewest assumptions about the object, but it may make sense to refine these these types in the future.

rsconnect/api.py
```diff
@@ -390,20 +396,20 @@ def output_task_log(task_status, last_status, log_callback):
 class RSConnectExecutor:
     def __init__(
         self,
-        ctx: click.Context = None,
-        name: str = None,
-        url: str = None,
-        api_key: str = None,
+        ctx: Optional[click.Context] = None,
```
wch, line 399: These instances of `Optional` were necessary to satisfy pyright's strict mode, and they make the types more accurately reflect the actual possible values.

rsconnect/api.py
```diff
@@ -710,20 +716,20 @@ def upload_rstudio_bundle(self, prepare_deploy_result, bundle_size: int, content
...
-        title: str = None,
+        app_id: Optional[str | int] = None,
+        deployment_name: Optional[str] = None,
+        title: Optional[str] = None,
         title_is_default: bool = False,
-        bundle: IO = None,
-        env_vars=None,
-        app_mode=None,
-        visibility=None,
+        bundle: Optional[BinaryIO] = None,
```
wch, line 723: I think `BinaryIO` is the correct type here, but I am not 100% sure.

rsconnect/environment.py
```diff
@@ -6,38 +6,30 @@
...
-)
+class Environment(NamedTuple):
+    contents: Optional[str]
+    error: Optional[str]
+    filename: Optional[str]
+    locale: Optional[str]
+    package_manager: Optional[str]
+    pip: Optional[str]
+    python: Optional[str]
+    source: Optional[str]
```
wch, line 32: The `class` syntax for NamedTuple is the preferred syntax since Python 3.7.

rsconnect/environment.py
```diff
@@ -6,38 +6,30 @@
...
 import os
 import re
 import subprocess
 import sys
-
-try:
-    import typing
-    from typing import Optional
-except ImportError:
-    typing = None
```
wch, line 22: I removed this `try` statement because `typing` has been part of the standard library since Python 3.5.

pyproject.toml
```diff
@@ -87,3 +87,6 @@ write_to = "rsconnect/version.py"
 
 [tool.pytest.ini_options]
 markers = ["vetiver: tests for vetiver"]
+
+[tool.pyright]
+typeCheckingMode = "strict"
```
wch, line 92: I added this configuration for pyright/pylance, setting it to strict mode. For developers who work in VS Code and use pylance, this will encourage adding more accurate types over time. (I'm happy to undo this if it's too much.)

rsconnect/utils_package.py
```diff
@@ -0,0 +1,265 @@
...
+        environment = environment._replace(contents=requirements_txt)
+
+    elif len(starlette_req.specs) == 1 and starlette_req.specs[0].operator == "==":
+        if compare_semvers(starlette_req.specs[0].version, "0.35.0") >= 0:
+            # starlette is in requirements.txt, but with a version spec that is
+            # not compatible with this version of Connect.
+            print(
+                "Starlette version is 0.35.0 or greater, but this version of Connect "
+                "requires starlette<0.35.0. Setting to <0.35.0."
+            )
```
wch, line 256: This uses `print()` to emit a message for the user. Is that the right way to print messages?

mmarchetti: Elsewhere we are using:
```
from .log import logger
logger.warn(...)
```

mmarchetti (approved):

rsconnect/bundle.py
```diff
@@ -1509,7 +1506,7 @@ def _warn_if_no_requirements_file(directory):
         )
 
 
-def _warn_if_environment_directory(directory):
+def _warn_if_environment_directory(directory: str):
```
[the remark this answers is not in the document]
wch: Great, I'll add that. I also like `pathlib.Path`.

rsconnect/main.py
```diff
@@ -1393,25 +1396,25 @@ def deploy_app(
...
-        python,
+        python: Optional[str],
         force_generate: bool,
         verbose: int,
-        directory,
-        extra_files,
-        visibility: typing.Optional[str],
-        env_vars: typing.Dict[str, str],
+        directory: str,
+        extra_files: tuple[str, ...] | list[str],
```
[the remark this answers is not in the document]
wch: This is one spot where the type checker made things a bit more difficult. There's this line below:

```py
extra_files = validate_extra_files(directory, extra_files)
```

But that function returns a list instead of a tuple, and if we have `extra_files: tuple[str, ...]`, the type checker doesn't like it. But maybe I can change that function to return a tuple instead of a list -- I'll check if that looks safe from other places where that function is called.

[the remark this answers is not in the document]
wch: /

rsconnect/main.py
```diff
@@ -1393,30 +1396,30 @@ def deploy_app(
...
-        secret: str = None,
+        account: Optional[str] = None,
+        token: Optional[str] = None,
+        secret: Optional[str] = None,
         no_verify: bool = False,
     ):
         set_verbosity(verbose)
         entrypoint = validate_entry_point(entrypoint, directory)
-        extra_files = validate_extra_files(directory, extra_files)
+        extra_files_list = validate_extra_files(directory, extra_files)
```
wch, line 1422: I found it very difficult to tell whether all uses of `validate_extra_files()` could safely accept a tuple as the return value, so instead of changing that function, I decided to create a new variable here, `extra_files_list`, which the type checker knows is a list, as opposed to re-using `extra_files`, because the type checker thinks that variable is a tuple from the function signature.

kgartland-rstudio:
Tested on Connect Versions:
2022.08.0 (oldest supported version) 👍 
2023.12.0 (last version before fix) 👍 
2024.01.0 (fixed version) ✅ 

👍 `starlette<0.35.0` is added to requirements.txt
✅  requirements.txt file is left unaltered

If starlette is added manually to the requirements.txt file prior to deployment:
✅ _without_ a version pin: we properly alter and pin `starlette<35.0`.
✅ `starlette<0.30.0` we respect the pin.
❓  `starlette>0.35.0` we respect the pin.

In the case where:
* a user has manually pinned to `starlette > 0.35`
* a user hasn't supplied a requirements.txt file but has `starlette > 0.35` installed in their environment.

Should we throw a big yellow warning for that case?

We should still alert Support of this case and see if they can put something on their Support form.

wch:
> Tested on Connect Versions: 2022.08.0 (oldest supported version) 👍 2023.12.0 (last version before fix) 👍 2024.01.0 (fixed version) ✅

Note that with Connect 2024.01.0 and greater, the changes in this PR should have no effect on deployments; these changes only affect older versions.

> 👍 `starlette<0.35.0` is added to requirements.txt
> ✅ requirements.txt file is left unaltered
> 
> If starlette is added manually to the requirements.txt file prior to deployment:
 ✅ _without_ a version pin: we properly alter and pin `starlette<35.0`.
> ✅ `starlette<0.30.0` we respect the pin. 
>❓ `starlette>0.35.0` we respect the pin.


In these cases, this PR currently changes the pinned version and emits a warning:
 https://github.com/rstudio/rsconnect-python/blob/0a2fef1a0eb6be37e7a190f9779bdb2136cd6ebb/rsconnect/utils_package.py#L250-L259

As you noted, these are the two cases where the `requirements.txt` would have `starlette >= 0.35.0`:

> In the case where:
> 
> * a user has manually pinned to `starlette > 0.35`
> * a user hasn't supplied a requirements.txt file but has `starlette > 0.35` installed in their environment.
> 
> Should we throw a big yellow warning for that case?

I'll call these case A and case B.

Case A: If they have manually pinned to `starlette >= 0.35`, I think it would be reasonable to emit a warning and _not_ change the file.

Case B: If they have _not_ supplied a `requirements.txt` and `pip` automatically generates a `requirements.txt` with `starlette >= 0.35`, then I think we should still set `starlette<0.35.0`. I believe this will be a fairly common situation -- if someone does a `pip install shiny`, they will automatically get the latest version of starlette (after we make our next shiny release which un-pins starlette). I don't have a strong preference either way about emitting a warning in this situation.


Currently, the code in this PR does not distinguish between case A and case B. The current behavior in this PR in both cases is to set `starlette<0.35.0` and emit a warning. I think this behavior is also OK.


So in summary, we could leave the PR as is, or distinguish between case A and B as described above... or something else. I'm happy to do any of these options.

kgartland-rstudio:
I think those options sound perfect to cover these last two cases.
Case A: emit a warning but do not alter requirements.txt file

mmarchetti:
I'd prefer in both cases to warn and inject the compatible version of starlette (< 0.35.0).

wch:
> I'd prefer in both cases to warn and inject the compatible version of starlette (< 0.35.0).

If we go with this (which is what the code currently does), then I can edit the warning message to be a bit more informative. Currently it says:

> Starlette version is 0.35.0 or greater, but this version of Connect requires starlette<0.35.0. Setting to <0.35.0.

But I could change it to something like this, if it will reduce the need for users to ask support about it:

> Starlette version is 0.35.0 or greater, but this version of Posit Connect requires starlette<0.35.0. Setting to <0.35.0. Note that newer versions of Posit Connect (2024.01.0 and later) are compatible with starlette>=0.35.0, and will not make this change.

kgartland-rstudio:
> If we go with this (which is what the code currently does), then I can edit the warning message to be a bit more informative. Currently it says:
> 
> > Starlette version is 0.35.0 or greater, but this version of Connect requires starlette<0.35.0. Setting to <0.35.0.

I haven't seen this warning during my testing.  I just deployed a shiny app with this in my requirements.txt file:
```
starlette>0.35.0
```

Here is the output deploying to Connect version 2022.11.0:

```
> rsconnect deploy shiny . -s http://100.27.28.165:3939 -k $ADMIN_API_KEY -N -t grthan30
Validating server... 	[OK]
Validating app mode... 	[OK]
Making bundle ... 	[OK]
Deploying bundle ... 	[OK]
Saving deployed information... 	[OK]
Building Python Shiny application...
Bundle created with Python version 3.10.11 is compatible with environment Local with Python version 3.8.2 from /opt/Python/3.8.2/bin/python3.8
Bundle requested Python version 3.10.11; using /opt/Python/3.8.2/bin/python3.8 which has version 3.8.2
2024/02/21 17:49:26.616713158 Running on host: perftest-connect
2024/02/21 17:49:26.616726757 Working directory: /opt/rstudio-connect/mnt/app
2024/02/21 17:49:26.616788281 Environment will be built with Python "3.8.2 (default, May  7 2020, 20:00:49)  [GCC 7.3.0]" at /opt/Python/3.8.2/bin/python3.8
2024/02/21 17:49:26.616911580 Running as user: rstudio-connect
2024/02/21 17:49:26.617576083 Using cached environment: 37RirkOh1r3tnhffyx9OwA
2024/02/21 17:49:27.055026981 Packages in the environment: aiofiles==23.2.1, altair==5.2.0, annotated-types==0.6.0, anyio==4.3.0, appdirs==1.4.4, argon2-cffi==23.1.0, argon2-cffi-bindings==21.2.0, arrow==1.3.0, asgiref==3.7.2, asttokens==2.4.1, async-lru==2.0.4, attrs==23.2.0, Babel==2.14.0, backcall==0.2.0, beautifulsoup4==4.12.3, bleach==6.1.0, certifi==2020.12.5, cffi==1.16.0, charset-normalizer==3.3.2, click==8.1.7, comm==0.2.1, debugpy==1.8.1, decorator==5.1.1, defusedxml==0.7.1, distlib, exceptiongroup==1.2.0, executing==2.0.1, fastapi==0.109.2, fastjsonschema==2.19.1, filelock, fqdn==1.5.1, h11==0.14.0, htmltools==0.5.1, httpcore==1.0.4, httpx==0.27.0, idna==3.6, importlib-metadata==7.0.1, importlib-resources==6.1.1, ipykernel==6.29.2, ipytablewidgets==0.3.1, ipython==8.12.3, ipywidgets==8.1.2, isoduration==20.11.0, jedi==0.19.1, Jinja2==3.1.3, json5==0.9.17, jsonpointer==2.4, jsonschema==4.21.1, jsonschema-specifications==2023.12.1, jupyter==1.0.0, jupyter-console==6.6.3, jupyter-events==0.9.0, jupyte
r-lsp==2.2.2, jupyter_client==8.6.0, jupyter_core==5.7.1, jupyter_server==2.12.5, jupyter_server_terminals==0.5.2, jupyterlab==4.1.2, jupyterlab_pygments==0.3.0, jupyterlab_server==2.25.3, jupyterlab_widgets==3.0.10, linkify-it-py==2.0.3, lz4==4.3.3, markdown-it-py==3.0.0, MarkupSafe==2.1.5, matplotlib-inline==0.1.6, mdit-py-plugins==0.4.0, mdurl==0.1.2, mistune==3.0.2, nbclient==0.9.0, nbconvert==7.16.1, nbformat==5.9.2, nest-asyncio==1.6.0, notebook==7.1.0, notebook_shim==0.2.4, numpy==1.24.4, overrides==7.7.0, packaging==23.2, pandas==1.5.3, pandocfilters==1.5.1, parso==0.8.3, pexpect==4.9.0, pickleshare==0.7.5, pkgutil_resolve_name==1.3.10, platformdirs==4.2.0, prometheus_client==0.20.0, prompt-toolkit==3.0.36, psutil==5.9.8, ptyprocess==0.7.0, pure-eval==0.2.2, pycparser==2.21, pydantic==2.6.1, pydantic_core==2.16.2, Pygments==2.17.2, python-dateutil==2.8.2, python-json-logger==2.0.7, python-multipart==0.0.9, pytz==2024.1, PyYAML==6.0.1, pyzmq==25.1.2, qtconsole==5.5.1, QtPy==2.4.1, questionary==2.0.1, r
eferencing==0.33.0, requests==2.31.0, rfc3339-validator==0.1.4, rfc3986-validator==0.1.1, rpds-py==0.18.0, Send2Trash==1.8.2, shiny==0.6.1.1, shinywidgets==0.2.4, six, sniffio==1.3.0, soupsieve==2.5, stack-data==0.6.3, starlette==0.36.3, terminado==0.18.0, tinycss2==1.2.1, tomli==2.0.1, toolz==0.12.1, tornado==6.4, traitlets==5.14.1, traittypes==0.2.1, types-python-dateutil==2.8.19.20240106, typing_extensions==4.9.0, uc-micro-py==1.0.3, uri-template==1.3.0, urllib3==2.2.1, uvicorn==0.27.1, vega==4.0.0, virtualenv, watchfiles==0.21.0, wcwidth==0.2.13, webcolors==1.13, webencodings==0.5.1, websocket-client==1.7.0, websockets==12.0, widgetsnbextension==4.0.10, zipp==3.17.0,
2024/02/21 17:49:27.055148663 Creating lockfile: python/requirements.txt.lock
Completed Python build against Python version: '3.8.2'
Launching Python Shiny application...
	 Dashboard content URL: http://100.27.28.165:3939/connect/#/apps/688a0ee6-9854-490a-b933-b9765637d4d0
	 Direct content URL: http://100.27.28.165:3939/content/688a0ee6-9854-490a-b933-b9765637d4d0/
Verifying deployed content... 	[OK]
```

I also deployed with `-v` and `-vv` and didn't see a warning being emitted.

wch:
Just to be clear, the warning _does_ appear if you use a version specifier with `==` instead of `>`. When I deploy to Colorado, I see the warning message:

```
❯ rsconnect deploy shiny -n colorado foo
    Warning: Capturing the environment using 'pip freeze'.
             Consider creating a requirements.txt file instead.
[WARNING] 2024-02-26T12:03:07-0600 Starlette version is 0.35.0 or greater, but this version of Connect requires starlette<0.35.0. Setting to <0.35.0.
Validating server... 	[OK]
Validating app mode... 	[OK]
Making bundle ... 	[OK]
Deploying bundle ... 	[OK]
Saving deployed information... 	[OK]
Building Python Shiny application...
Bundle created with Python version 3.11.6 is compatible with environment Kubernetes::ghcr.io/rstudio/content-pro:r4.2.2-py3.11.3-ubuntu2204 with Python version 3.11.3 from /opt/python/3.11.3/bin/python3 
Bundle requested Python version 3.11.6; using /opt/python/3.11.3/bin/python3 from Kubernetes::ghcr.io/rstudio/content-pro:r4.2.2-py3.11.3-ubuntu2204 which has version 3.11.3
Determining session server location ...
...
```


*****

I think there was some ambiguity in the way that we were talking about the version requirements here. If my memory serves, in a conversation I had with @mmarchetti, we had talked about doing the version checks only when the `==` operator is used in the requirements.txt, because that is by far the most likely situation -- `pip freeze` will generate a requirements.txt with the `==` for version numbers.

So, for example, they will see the warning if:
- There is no requirements.txt (and so it is automatically generated by `rsconnect deploy`)
- The requirements.txt has `==`, as in `starlette==0.37.1` (this is likely what the user would get if they ran `pip freeze`)
- The requirements.txt has `starlette` without a version specification
- The requirements.txt has `shiny` (with or without a version specification), but does not list `starlette` specifically

If there is a requirements.txt with a version specifier other than `==`, like `starlette>0.35.0`, they almost certainly are doing that manually and intentionally, and we'll just let them do it.

If we try to account for all possible version specifiers, then the problem of determining whether the starlette version is OK becomes significantly more complicated.

So IMO this PR covers the vast majority of real-world use cases, and the underlying issue will go away over time as users upgrade their Connect installations.

mmarchetti:
> If my memory serves, in a conversation I had with @mmarchetti, we had talked about doing the version checks only when the `==` operator is used in the requirements.txt, because that is by far the most likely situation -- `pip freeze` will generate a requirements.txt with the `==` for version numbers.
> ...
> If there is a requirements.txt with a version specifier other than ==, like starlette>0.35.0, they almost certainly are doing that manually and intentionally, and we'll just let them do it.

Yes, we dicussed handling the common/automatic cases that get users in trouble and letting advanced users specify what they need. Sorry I didn't document it here in the PR, it would have made things clearer.

kgartland-rstudio:
I hadn't realized the different behavior wrt use `>` vs `==`.  This makes sense now.  I think we can move ahead with what we have. Thanks for the explanation!

PR #546, opened 2024-02-22
Follow symlinks when building bundle

### LINKED ISSUE #414: support following symlinks
> In my case, the app.py is a symlink to a file in the parent directory (long story, but there is a good reason for this). When rsconnect deploys an app, does it follow local symlinks so that the content is deployed to the server?
> The reason I have symlinks is because I have many apps in the same directory (named app_basic.py, app_recipes.py, etc.) and I want to deploy the apps individually, but I think rsconnect wants them to be named app.py. So I made a directory for each with symlinks from foo/app.py to the original file, like app_basic.py.
Why are they all in the same directory? Because they share common submodules, like utils.py, chat/_init_.py, and so on.

When constructing the list of files to bundle we current rely on [`os.path.isfile`](https://github.com/rstudio/rsconnect-python/blob/bcwu-deploy-manifest-tests/rsconnect/bundle.py#L962) so symlinks are not followed.

jcheng5:
I have NOT tested this on Windows, only Mac. Please nobody merge until I've tested.

## Intent
Fixes #414. When building a bundle, follow symlinks. This allows apps to include directories that might live somewhere else for deduplication/centralization reasons.

## Type of Change
- [x] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change   

## Approach
`os.walk` takes a `followlinks` argument that defaults to `False`. I simply set this to `True`. Note that from googling, `os.walk` doesn't check for circular references--I'm happy to add that if we think it's important.

## Automated Tests
TODO

## Directions for Reviewers
Create two sibling directories, one called `data` and one called `shinytest`.

shinytest/app.py:
```
from pathlib import Path
from shiny.express import input, ui

msg_path = Path(__file__).parent / "data/message.txt"

with open(msg_path) as f:
    f.read()
```

data/message.txt:
```
Hello, world!
```

Then, from within the shinytest directory, run `ln -s ../data data`.

Deploy from within the shinytest directory as usual, and see if it crashes on startup or the message correctly appears instead.

## Checklist

### DIFF
--- a/rsconnect/bundle.py
+++ b/rsconnect/bundle.py
@@ -578,6 +578,7 @@ def list_files(
     base_dir: str,
     include_sub_dirs: bool,
     walk: Callable[[str], Iterator[tuple[str, list[str], list[str]]]] = os.walk,
+    followlinks: bool = True,
 ) -> list[str]:
     """List the files in the directory at path.
 
@@ -1259,7 +1260,7 @@ def create_file_list(
         file_set.add(path_to_add)
         return sorted(file_set)
 
-    for cur_dir, _, files in os.walk(path):
+    for cur_dir, _, files in os.walk(path, followlinks=True):
         if Path(cur_dir) in exclude_paths:
             continue
         if any(parent in exclude_paths for parent in Path(cur_dir).parents):
--- a/rsconnect/bundle.py
+++ b/rsconnect/bundle.py
@@ -574,37 +574,6 @@ def write_manifest(
     return created, skipped
 
 
-def list_files(
-    base_dir: str,
-    include_sub_dirs: bool,
-    walk: Callable[[str], Iterator[tuple[str, list[str], list[str]]]] = os.walk,
-    followlinks: bool = True,
-) -> list[str]:
-    """List the files in the directory at path.
-
-    If include_sub_dirs is True, recursively list
-    files in subdirectories.
-
-    Returns an iterable of file paths relative to base_dir.
-    """
-    skip_dirs = [".ipynb_checkpoints", ".git"]
-
-    def iter_files():
-        for root, sub_dirs, files in walk(base_dir):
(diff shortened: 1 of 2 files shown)


rsconnect/bundle.py
```diff
@@ -498,7 +498,7 @@ def write_manifest(
     return created, skipped
 
 
-def list_files(base_dir, include_sub_dirs, walk=os.walk):
+def list_files(base_dir, include_sub_dirs, walk=os.walk, followlinks=True):
```
nealrichardson, line 501: Just reading the diff, does this get used? Or does it belong where walk is called in L512?

jcheng5: 🤦

jcheng5: @mmarchetti It actually looks to me like this function has never been used, except from its own unit test. Safe to delete?

If not, I think I'd need to change it to effectively `walk=functools.partial(os.walk, followlinks=True)` or equivalent.

mmarchetti: > Safe to delete?

Seems like it.

jcheng5:
I played around with Windows and it's pretty weird.

The good news is symlinks work fine, the command is [`mklink`](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/mklink) and Python handles them just fine. I haven't tried with git but it's supposed to work; there's a `core.symlinks` option that defaults to true.

They're not supported on FAT filesystems but that's probably OK for our purposes.

However, by default, you don't have permissions to call `mklink` unless you are at an Administrator prompt. That's true even if you have write permissions to the directory where you're trying to create the link--it's the creation of a symlink that's considered a privileged operation. You can put Windows into [developer mode](https://learn.microsoft.com/en-us/windows/apps/get-started/enable-your-device-for-development) (doesn't even require a restart) and then anyone can call mklink.

PR #549, opened 2024-03-15
Add support for shiny mode magic comment

wch:
## Intent

This PR adds support for shiny mode magic comments which are generated by interactive Quarto dashboards that use Shiny for Python. This change allows users to create Quarto dashboards that have `from shiny.express import ...` instead of `from shiny ...`; previously those apps would not run.

See the following for more information:
- https://github.com/posit-dev/py-shiny/pull/1217 
- https://github.com/quarto-dev/quarto-web/pull/1049

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach

This copies over code from the py-shiny repository, from this file:

https://github.com/posit-dev/py-shiny/blob/2b66c898cb9d0fc827e587b11a839e270b2e814f/shiny/express/_is_express.py

## Checklist

### DIFF
--- a/rsconnect/shiny_express.py
+++ b/rsconnect/shiny_express.py
@@ -96,14 +96,14 @@ def find_magic_comment_mode(content: str) -> Literal["core", "express"] | None:
     Returns
     -------
     :
-        `True` if Shiny Express comment is found, `False` if Shiny Core comment is
+        `"express"` if Shiny Express comment is found, `"core"` if Shiny Core comment is
         found, and `None` if no magic comment is found.
     """
     m = re.search(r"^#[ \t]*shiny_mode:[ \t]*(\S*)[ \t]*$", content, re.MULTILINE)
     if m is not None:
         shiny_mode = cast(str, m.group(1))
         if shiny_mode in ("express", "core"):
-            return shiny_mode
+            return shiny_mode # type: ignore
         else:
             print(f'Invalid shiny_mode: "{shiny_mode}"', file=sys.stderr)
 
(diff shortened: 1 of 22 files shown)


wch:
The linting test is failing because in Python 3.7, the `typing` module doesn't have `Literal`. (It is present in 3.8+.)

I could work around this by importing from `typing_extensions`. That said, my understanding is that we've dropped Python 3.7 support, so can we stop testing on that platform?

mmarchetti:
rsconnect-python hasn't removed [Python 3.7 support](https://github.com/rstudio/rsconnect-python/blob/master/pyproject.toml#L8), but Connect has removed it so it's probably time.

nealrichardson:
> rsconnect-python hasn't removed [Python 3.7 support](https://github.com/rstudio/rsconnect-python/blob/master/pyproject.toml#L8), but Connect has removed it so it's probably time.

I made https://github.com/rstudio/rsconnect-python/issues/556 for that.

rsconnect/shiny_express.py
```diff
@@ -56,25 +66,52 @@ def __init__(self):
...
+    def visit_Module(self, node: ast.Module) -> None:
         super().generic_visit(node)
 
     # Don't recurse into any nodes, so the we'll only ever look at top-level nodes.
-    def generic_visit(self, node: ast.AST):
+    def generic_visit(self, node: ast.AST) -> None:
         pass
 
 
+def find_magic_comment_mode(content: str) -> Literal["core", "express"] | None:
```
nealrichardson, line 88: Would it be worth adding a test for this function?

aronatkins:
@wch - if you pull in master, you should see CI pass.

rsconnect/shiny_express.py
wch: I've added tests for Shiny Express app detection in general in [cc3766e](https://github.com/rstudio/rsconnect-python/pull/549/commits/cc3766e3b30b5863a309153020e736c251cff6aa).

PR #554, opened 2024-03-19
quarto: adjust manifest construction

### LINKED ISSUE #320: exclude glob patterns do not work on Windows
Encountered while fixing a separate issue:

https://github.com/rstudio/rsconnect-python/pull/294#issuecomment-1329510093

### LINKED ISSUE #551: doc: explain "**" glob
In order to exclude everything other than explicitly enumerated files, I ended up doing something like:

```bash
rsconnect write-manifest quarto . _quarto.yml index.qmd requirements.txt --exclude '**'
```

The `**` and `**/*` glob syntaxes are not documented. Only `*` is explained.

### LINKED ISSUE #552: quarto: configure as "site" only with multiple render targets
Bundles should not be marked as "site" when there is only one render target.

https://github.com/rstudio/rsconnect-python/blob/18bfc5a5ac5b0ea03a2b9610dd04685e3919bd7f/rsconnect/bundle.py#L98-L103
https://github.com/rstudio/rsconnect-python/blob/18bfc5a5ac5b0ea03a2b9610dd04685e3919bd7f/rsconnect/bundle.py#L328-L334

Single-document projects are incorrectly marked as site. The existence project configuration is not enough to call the bundle a "site".

### LINKED ISSUE #553: quarto: manifest includes derived files
Given a Quarto project:

The `_quarto.yml`:

```yaml
project: {}
```

A `requirements.txt`:

```
jupyter
matplotlib
numpy
```

The `index.qmd`:

````markdown
---
title: "matplotlib demo"
format:
  html:
    code-fold: true
jupyter: python3
---

For a demonstration of a line plot on a polar axis, see @fig-polar.

```{python}
#| label: fig-polar
#| fig-cap: "A line plot on a polar axis"

import numpy as np
import matplotlib.pyplot as plt

r = np.arange(0, 2, 0.01)
theta = 2 * np.pi * r
fig, ax = plt.subplots(
  subplot_kw = {'projection': 'polar'}
)
ax.plot(theta, r)
ax.set_rticks([0.5, 1, 1.5, 2])
ax.grid(True)
plt.show()
```
````

Render this project, then create a manifest for this content using rsconnect.

```bash
quarto render
rsconnect write-manifest quarto .
```

The generated `manifest.json` enumerates `index.html` and the entirety of `index_files/**`. This is because manifest construction is attempting to automatically filter derived files, but uses `config.project.render` rather than `input.files`.

https://github.com/rstudio/rsconnect-python/blob/18bfc5a5ac5b0ea03a2b9610dd04685e3919bd7f/rsconnect/bundle.py#L1307-L1312

This is because the earliest test projects forced rendering with configurations like:

```yaml
project:
  render:
    - index.qmd
```

The use of this pattern led us to assume that the `config.project.render` inspect output enumerated files. It does not; it enumerates patterns.

aronatkins:
## Intent

* multi-file project according to files.input
* automatic ignores according to files.input
* glob handling allows posix OS path separators (`/` and optionally `\\`)
* documentation for '**'

fixes #552
fixes #553
fixes #551
fixes #320

## Type of Change

- [x] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change   

## Approach

Actively iterating and deploying different types of Quarto content; filed each of the linked issues based on observations.

## Automated Tests

* Expanded Quarto single-file, simple project and multi-file project tests. Confirm that intermediate and rendered output is ignored, that simple projects are not classified as sites.
* Expanded glob tests to test using "/" and host-specific ("/" and "\\") path separators.

## Directions for Reviewers

> If your content is within a Git repository, Quarto may create a `.gitignore` file.
> If it exists, `.gitignore` will show up in the `manifest.json` for projects.

Simple Quarto document

```bash
mkdir simple
cd simple
touch index.qmd
rsconnect write-manifest quarto index.qmd
# confirm that this is quarto-static content and only index.qmd is a target file
```

Simple Quarto document, rendered

```bash
mkdir simple-rendered
cd simple-rendered
touch index.qmd
quarto render index.qmd
rsconnect write-manifest quarto index.qmd
# confirm that this is quarto-static content and only index.qmd is a target file
```

Simple Quarto project

```bash
mkdir simple-project
cd simple-project
echo "project:\n  type: default" > _quarto.yml
touch index.qmd
rsconnect write-manifest quarto .
# confirm that this is quarto-static content with _quarto.yml and index.qmd target files
```

Simple Quarto project, rendered

```bash
mkdir simple-project-rendered
cd simple-project-rendered
echo "project:\n  type: default" > _quarto.yml
touch index.qmd
quarto render
rsconnect write-manifest quarto .
# confirm that this is quarto-static content with _quarto.yml and index.qmd target files
```

Simple Quarto website

```bash
mkdir simple-website
cd simple-website
echo "project:\n  type: website" > _quarto.yml
touch index.qmd
touch about.qmd
rsconnect write-manifest quarto .
# confirm that this is quarto-static content with _quarto.yml, index.qmd, about.qmd target files
# confirm that this is marked as a "site"
```

Simple Quarto project, rendered

```bash
mkdir simple-website-rendered
cd simple-website-rendered
echo "project:\n  type: website" > _quarto.yml
touch index.qmd
touch about.qmd
quarto render
rsconnect write-manifest quarto .
# confirm that this is quarto-static content with _quarto.yml, index.qmd, about.qmd target files
# confirm that this is marked as a "site"
```

## Checklist

### DIFF
--- a/rsconnect/bundle.py
+++ b/rsconnect/bundle.py
@@ -94,12 +94,10 @@ class Manifest:
                 "version": quarto_inspection.get("quarto", {}).get("version", "99.9.9"),
                 "engines": quarto_inspection.get("engines", []),
             }
-            project_config = quarto_inspection.get("config", {}).get("project", {})
-            render_targets = project_config.get("render", [])
-            if len(render_targets):
-                self.data["metadata"]["primary_rmd"] = render_targets[0]
-            project_type = project_config.get("type", None)
-            if project_type or len(render_targets) > 1:
+
+            files_data = quarto_inspection.get("files", {})
+            files_input_data = files_data.get("input", [])
+            if len(files_input_data) > 1:
                 self.data["metadata"]["content_category"] = "site"
 
         if environment:
@@ -325,12 +323,10 @@ def make_source_manifest(
             "version": quarto_inspection.get("quarto", {}).get("version", "99.9.9"),
             "engines": quarto_inspection.get("engines", []),
         }
-        project_config = quarto_inspection.get("config", {}).get("project", {})
-        render_targets = project_config.get("render", [])
-        if len(render_targets):
-            manifest["metadata"]["primary_rmd"] = render_targets[0]
-        project_type = project_config.get("type", None)
-        if project_type or len(render_targets) > 1:
+
+        files_data = quarto_inspection.get("files", {})
+        files_input_data = files_data.get("input", [])
+        if len(files_input_data) > 1:
             manifest["metadata"]["content_category"] = "site"
 
     if environment:
@@ -1303,13 +1299,19 @@ def make_quarto_manifest(
         output_dir = project_config.get("output-dir", None)
         if output_dir:
             excludes = excludes + [output_dir]
-        else:
-            render_targets = project_config.get("render", [])
-            for target in render_targets:
-                t, _ = splitext(target)
-                # TODO: Single-file inspect would give inspect.formats.html.pandoc.output-file
-                # For foo.qmd, we would get an output-file=foo.html, but foo_files is not available.
-                excludes = excludes + [t + ".html", t + "_files"]
+
+        files_data = quarto_inspection.get("files", {})
+        files_input_data = files_data.get("input", [])
+        # files.input is a list of absolute paths to input (rendered)
+        # files. Automatically ignore the most common derived files for
+        # those inputs.
+        #
+        # These files are ignored even when the project has an output
+        # directory, as Quarto may create these files while a render is
+        # in-flight.
+        for each in files_input_data:
+            t, _ = splitext(os.path.relpath(each, file_or_directory))
+            excludes = excludes + [t + ".html", t + "_files/**/*"]
 
         # relevant files don't need to include requirements.txt file because it is
         # always added to the manifest (as a buffer) from the environment contents
--- a/rsconnect/models.py
+++ b/rsconnect/models.py
@@ -2,7 +2,7 @@
 Data models
 """
 
-import os
+import pathlib
 import re
 
 import fnmatch
@@ -163,6 +163,7 @@ class GlobMatcher(object):
     """
 
     def __init__(self, pattern):
+        pattern = pathlib.PurePath(pattern).as_posix()
         if pattern.endswith("/**/*"):
             # Note: the index used here makes sure the pattern has a trailing
             # slash.  We want that.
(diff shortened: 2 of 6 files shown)


rsconnect/bundle.py
```diff
@@ -94,12 +94,10 @@ def __init__(
                 "version": quarto_inspection.get("quarto", {}).get("version", "99.9.9"),
                 "engines": quarto_inspection.get("engines", []),
             }
-            project_config = quarto_inspection.get("config", {}).get("project", {})
-            render_targets = project_config.get("render", [])
-            if len(render_targets):
-                self.data["metadata"]["primary_rmd"] = render_targets[0]
```
mmarchetti, line 100: Do we not need to set `primary_rmd`?

rsconnect/models.py
```diff
@@ -163,6 +163,7 @@ class GlobMatcher(object):
     """
 
     def __init__(self, pattern):
+        pattern = pathlib.PurePath(pattern).as_posix()
```
mmarchetti, line 166: +1 for `Pathlib`!

mmarchetti (approved):

rsconnect/bundle.py
aronatkins: It's not used by Connect for Quarto content, today.

With R Markdown and Jupyter content, we tell the tool the specific file that we want to render, and the output file is written to the `.index` file.

This path only applies for projects, and we want to always rely on the Quarto discovery rules in that case. Only projects have the `config.project.render` structure in the output from `quarto inspect`.

For Quarto standalone documents, there is a stronger case for communicating the primary document/entrypoint. Connect currently transforms content into a project and then has Quarto use its default discovery rules. If the primary file is deployed along with secondary Markdown files, those would need to be named `_filename.md` to be ignored during rendering. Otherwise, Quarto could render both `primary.qmd` and `do-not-render.qmd`. They could also deploy a `_quarto.yml` file alongside the file, and deploy it as a simple "project". Both of those options feel OK, right now.

We may want to eventually annotate the manifest with `metadata.entrypoint`, but I'd like to see if we can avoid it.

kgartland-rstudio:
Validated! 
✅ Tested all scenarios above
✅ Tested on Windows + MacOS
✅ Tested inside and oustide target directory
✅ Subdirectories are created with proper path
✅ Excluding with wildcards work as expected (test.txt, *.txt,  subdir/*, subdir/*/*.txt, subdir/*/*)

kgartland-rstudio (approved):

PR #558, opened 2024-03-20
Add support for Shiny Express apps when writing manifest files

wch:
This PR does the following:

- Adds support for Shiny Express apps when writing manifest files.
- When deploying apps via `rsconnect deploy`, only try to detect Shiny Express for Shiny apps. (Previously it would attempt the detection for other types of apps as well, including `api`, `streamlit`, `voila`, etc.)
- Adds and corrects type annotations. (This was a helpful step before modifying the code to do the things above.) These changes only modify type annotations and should have no run-time effect.

## Intent

This addresses item (2) in https://github.com/posit-dev/py-shiny/issues/1232: `rsconnect write-manifest shiny` does not support Shiny Express apps.

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Checklist

### DIFF
--- a/rsconnect/main.py
+++ b/rsconnect/main.py
@@ -9,7 +9,13 @@ import textwrap
 import click
 from os.path import abspath, dirname, exists, isdir, join
 from functools import wraps
-from typing import Callable, ItemsView, Literal, Optional, ParamSpec, Sequence, TypeVar, cast
+from typing import Callable, ItemsView, Literal, Optional, Sequence, TypeVar, cast
+
+if sys.version_info >= (3, 10):
+    from typing import ParamSpec
+else:
+    from typing_extensions import ParamSpec
+
 
 from rsconnect.certificates import read_certificate_file
 
--- a/rsconnect/main.py
+++ b/rsconnect/main.py
@@ -1999,6 +1999,11 @@ def _write_framework_manifest(
     with cli_feedback("Inspecting Python environment"):
         _, environment = get_python_env_info(directory, python, force_generate)
 
+    if app_mode == AppModes.PYTHON_SHINY:
+        with cli_feedback("Inspecting Shiny for Python app"):
+            if is_express_app(entrypoint + ".py", directory):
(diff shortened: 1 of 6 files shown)


rsconnect/main.py
```diff
@@ -103,7 +111,7 @@ def failed(err):
         except EnvironmentException as exc:
             failed("Error: " + str(exc))
         except Exception as exc:
-            if click.get_current_context("verbose"):
+            if click.get_current_context(True):
```
wch, line 114: I believe this was an existing bug, which mypy started noticing after I made some other changes.

According to the docs, `click.get_current_context` wants a boolean, not a string. If `False`, it will throw an error if there's no context; if `True`, it'll return `None` in that situation. I set it to `True` because I think the intent here was not to throw another error.

rsconnect/main.py
```diff
@@ -1979,6 +2006,11 @@ def _write_framework_manifest(
     with cli_feedback("Inspecting Python environment"):
         _, environment = get_python_env_info(directory, python, force_generate)
 
+    if app_mode == AppModes.PYTHON_SHINY:
+        with cli_feedback("Inspecting Shiny for Python app"):
+            if is_express_app(entrypoint + ".py", directory):
+                entrypoint = "shiny.express.app:" + escape_to_var_name(entrypoint + ".py")
+
```
wch, line 2013: This is the part that adds support for Shiny Express when writing manifest files.

rsconnect/main.py
mmarchetti: Let's drop the `if` condition entirely and always print the traceback in this case. It looks like the original intent was to check the `verbose` flag from the command line and only print the traceback in verbose mode. The checks for `RSConnectException` and `EnvironmentException` suppress tracebacks for known kinds of exceptions (where we provide meaningful error messages); the rest can have trackbacks for debugging.

mmarchetti (approved):

PR #559, opened 2024-03-21
remove python 3.7 tests

kgartland-rstudio:
## Intent

`black` was updated to 24.3.0 here:
https://github.com/rstudio/rsconnect-python/commit/818d393c3b802136014839c927698ef5a7352779

but it's not available for Python 3.7:
https://github.com/rstudio/rsconnect-python/actions/runs/8372241141/job/22922828215

This PR removes Python 3.7 from the testing matrix.  The nightly tests now pass again:
https://github.com/rstudio/rsconnect-python/actions/runs/8375941880

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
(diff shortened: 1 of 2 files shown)


aronatkins (approved):
Are we missing something from our `main` workflow that caused this to only show up as a nightly failure?

kgartland-rstudio:
> Are we missing something from our `main` workflow that caused this to only show up as a nightly failure?

I have no idea why it passed `main` initially.

mmarchetti (commented):
:+1:

PR #555, opened 2024-03-20
Bump version of black in requirements.txt

nealrichardson:
## Intent

Snyk picked up this CVE in `black`: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2024-21503

This PR bumps to the latest version, which fixes it.

PR #560, opened 2024-03-21
black version constraint is in two places. happy days.

aronatkins:
This is why https://github.com/rstudio/rsconnect-python/pull/559 was needed and https://github.com/rstudio/rsconnect-python/pull/555 did not fail on master.


kgartland-rstudio (approved):
LGTM

Issue #448, opened 2023-07-07
GitHub Action fails due to Docker Compose Call Sequence


tdstein:
## Summary

The CI GitHub Action occasionally fails when executing the following step: https://github.com/rstudio/rsconnect-python/blob/master/.github/workflows/main.yml#L156-L182

See https://github.com/rstudio/rsconnect-python/actions/runs/5478386182/jobs/9999333833

In some instances, this results in the following error:

```
docker-compose up -d
rsconnect-python_rsconnect_1 is up-to-date
docker-compose exec -T rsconnect bash < vetiver-testing/setup-rsconnect/add-users.sh
Error response from daemon: Container 9bdc42da9e0c00adaa7765fde5cb31926403c905f2cd590f640bcc427092031e is restarting, wait until the container is running
make: *** [Makefile:173: dev] Error 1
Error: Process completed with exit code 2.
```

## Reasoning

During the execution of the `make dev` target, two consecutive commands to `docker-compose` are issued here: https://github.com/rstudio/rsconnect-python/blob/master/.github/workflows/main.yml#L156-L182

The second command, `docker-compose exec,` causes an asynchronous container restart. This is not mentioned in the [documentation](https://docs.docker.com/engine/reference/commandline/compose_exec/) and no logs are produced. But the restart can be viewed by watching `docker ps` during the execution of `make dev`.

Due to the asynchronous nature of the issued restart on "fast" machines, the shell script `vetiver-testing/setup-rsconnect/add-users.sh` will execute before the restart is complete. In some instances, this is not observed since the initialization sequence is observed before the shell script executes.

When this occurs, the shell script will log errors related to the `cracklib-runtime`. This is due to the `cracklib-runtime` not having finished initializing before the `chpasswd` command being issued.

## RCA

On my local machine, I can recreate this issue using the following steps:

1. Open a terminal window and run `while true; do docker ps; echo; done |  while IFS= read -r line; do printf '[%s] %s\n' "$(date '+%H:%M:%S')" "$line"; done > log`. 
4. Execute `make dev`
6. After `make dev` completes. Stop the first command.
7. Inspect the log file and search for "Less than a second ago".

Note that lines containing this message are issued in two distinct blocks. The first block is the initial startup via `docker-compose up -d`. The second block comes from `docker-compose exec ...`.

On my local machine, I can manually execute the shell script `add-users.sh` by reinstalling `cracklib-runtime`.

[...]

PR #566, opened 2024-03-28
Add comprehensive type annotations

wch:
This PR follows up on a discussion I had with @mmarchetti. It does the following: 
- Turns on strict type checking with pyright.
- Adds many type annotations. These by and large should not affect run-time behavior, and so should be safe.

In the changes here, there are still many, many issues flagged by pyright in strict mode.
The purpose of this is to set things up for a second pull request, which will be made against this pull request instead of master. The second PR will have changes that are not quite as safe as this PR -- it will add more type annotations and makes some changes to the logic, and is therefore not as safe as this PR. The reason they're done with two separate pull requests is so that the substantive changes in the second PR are more visible and not drowned out by the changes in this PR, which are safe. This will make the second PR easier to review. 

After the second PR is merged into this one and tests are passing, we can merge this PR into master.

The overall goal is to get the code passing with pyright in strict mode in CI, and disable mypy. Once this is all done, it will be much easier to contribute to this repository and refactor the code here.


mmarchetti:
This is awesome! A huge step - no, actually a leap! - forward toward modernizing this codebase.

rsconnect/actions.py
```diff
@@ -488,767 +265,16 @@ def validate_quarto_engines(inspect):
...
-        The server administrator is responsible for installing packages in the runtime environment. Default = None.
-    :param env_management_r: False prevents Connect from managing the R environment for this bundle.
-        The server administrator is responsible for installing packages in the runtime environment. Default = None.
-    :return: the ultimate URL where the deployed app may be accessed and the sequence
-    of log lines.  The log lines value will be None if a log callback was provided.
-    """
-    return deploy_app(app_mode=AppModes.PYTHON_API, **locals())
-
-
-def deploy_python_fastapi(
```
mmarchetti, line 832: This function is imported by `vetiver-python` and its removal may be why the tests are failing.

wch: Ah, that's good to know! This makes me a little concerned, though: I removed many of the functions in `actions.py` because they were not used internally and seemed to essentially be duplicates of functions in `main.py`. Do we need to keep any other of those functions around in case someone external was using them?

Also, in this particular case, I wonder if we can just have `actions.deploy_python_fastapi` be an alias to `main.deploy_python_fastapi`.

mmarchetti: Currently the only non-CLI clients are `vetiver-python` and `rsconnect-jupyter`.

https://github.com/rstudio/rsconnect-jupyter/blob/0c3d679d9ec799330134aa7710646870ec35e9c9/rsconnect_jupyter/__init__.py#L13C1-L29C45

mmarchetti: > I wonder if we can just have `actions.deploy_python_fastapi` be an alias to `main.deploy_python_fastapi`

Yes, as long as we're careful to match the type signature that Vetiver is expecting.

wch: Hm, there currently isn't a `main.deploy_python_fastapi`; the command line command is generated by the following and isn't saved in a variable anywhere:
```py
generate_deploy_python(app_mode=AppModes.PYTHON_FASTAPI, alias="fastapi", min_version="2021.08.0")
```

Also, the function signature is somewhat different and I don't want to risk breaking things, so for now I'll just bring back this code.

wch:
From some testing I did in #571, it looks like commit c476d69 made the test break. It looks like something on the server side stopped working? 

In that commit I added many type annotations, but didn't make any functional changes. I suppose it's possible that some of the syntax I used there was not compatible with Python 3.8.

This is the line that cause the error:

```
python vetiver-testing/setup-rsconnect/dump_api_keys.py vetiver-testing/rsconnect_api_keys.json
```

As far as I can tell, that code does not cause `rsconnect` to be imported at all, at least not on the client side. But the server side gives errors in response to requests. Does `rsconnect` get loaded on the server side somehow?

wch:
Now when I run the tests, the results are inconsistent. For example, on the same commit, I ran the workflow twice, and the failure happened differently between the two runs.

- First run: https://github.com/rstudio/rsconnect-python/actions/runs/8562649287/job/23466276020
- Second run: https://github.com/rstudio/rsconnect-python/actions/runs/8562649287/job/23466519638

I don't understand why the first one failed, but the second run failed in a way that makes sense to me and that I can fix.

wch:
I fixed the bug mentioned in the previous comment, and now here's a case where it fails on the first run and succeeds on the second:

1. https://github.com/rstudio/rsconnect-python/actions/runs/8562909967/job/23467049557
2. https://github.com/rstudio/rsconnect-python/actions/runs/8562909967/job/23467122442

I see that in #569, where the command was switched from `docker-compose` to `docker compose`, it had the same failure on the first attempt and then passed on the second attempt, so that makes me more confident that the failures aren't related to the content of this PR.

1. https://github.com/rstudio/rsconnect-python/actions/runs/8528435857/job/23362057518
2. https://github.com/rstudio/rsconnect-python/actions/runs/8528435857/job/23365251904


I suspect there's a timing problem with line `docker compose up -d` and the commands that come after it -- it might not have enough time to reliably start the container before the subsequent requests are made to the container. In 916535fd3433da8ce13993379b018072af78b31d I added a `sleep 4` after that line, and now it seems to be running more reliably -- two successes in a row.

aronatkins:
The timing problem is likely https://github.com/rstudio/rsconnect-python/issues/448. CC @tdstein

PR #567, opened 2024-03-28
Type cleanup

wch:
This is the second pull request described in #566. It adds many types and changes to the logic. The goal is to pass pyright strict mode type checks.

- [ ] Before merging, change main.yml back so that workflow is run only on pull requests that target `master`.


rsconnect/metadata.py
```diff
@@ -128,7 +128,7 @@ def _get_by_key(self, key: str, default: T | None = None) -> T | None:
         """
         return self._data.get(key, default)
 
-    def _get_by_value_attr(self, attr: str, value: str) -> T | None:
+    def _get_by_value_attr(self, attr: str, value: T) -> T | None:
```
mmarchetti, line 131: `value` here seems like it can't be a `T`, but instead should be a new type variable.

`T` is usually something like `AppMetadata` which is a `TypedDict`. `self._data` is a mapping from `str` to `T`.

`value` has to be the type of `attr` within a `T`; we're selecting a matching `T` based on the value of a specific attribute.

rsconnect/models.py
```diff
@@ -225,16 +225,20 @@ def _to_parts_list(pattern: str):
...
                 if depth_wildcard_index is not None:
                     raise ValueError('Only one occurrence of the "**" pattern is allowed.')
                 depth_wildcard_index = index
             elif any(ch in name for ch in "*?["):
-                parts[index] = re.compile(r"\A" + fnmatch.translate(name))
-        return parts, depth_wildcard_index
+                value = re.compile(r"\A" + fnmatch.translate(name))
+            parts_result.append(value)
+
+        return parts_result, depth_wildcard_index
```
mmarchetti, line 241: This function would benefit from a return type annotation.

rsconnect/models.py
```diff
@@ -457,16 +513,74 @@ class TaskStatusResult(TypedDict):
...
+    last: int
+    result: TaskStatusResult | None
+
+
+class BootstrapOutputDTO(TypedDict):
+    api_key: str
+
+
+# This not the complete specification of the server settings data structure, but it is
+# sufficient for the purposes of this package.
```
mmarchetti, line 543: :+1:

mmarchetti:
Forgot to mention in my comments: THANK YOU!!!

rsconnect/actions_content.py
```diff
@@ -81,29 +87,29 @@ def build_remove_content(
...
         raise RSConnectException(
             "There is a build running on this server, " + "please wait for it to finish before removing content."
         )
     guids: list[str] = [guid]
     if all:
-        guids = [c["guid"] for c in _content_build_store.get_content_items()]
+        guids = [c["guid"] for c in content_build_store().get_content_items()]
     for guid in guids:
-        _content_build_store.remove_content_item(guid, purge)
+        content_build_store().remove_content_item(guid, purge)
```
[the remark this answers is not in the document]
wch: What do you think about changing `init_content_build_store()` to a signature like this, where inits if necessary, and then always returns the store:

```py
def ensure_content_build_store(connect_server: RSConnectServer) -> ContentBuildStore:
    global _content_build_store
    if not _content_build_store:
        logger.info("Initializing ContentBuildStore for %s" % connect_server.url)
        _content_build_store = ContentBuildStore(connect_server)
    return _content_build_store
```

Then in usage in the various functions, it would look like this:

```py
build_store = ensure_content_build_store()
```

rsconnect/actions_content.py
```diff
@@ -455,13 +468,13 @@ def do_filter(item: ContentItem) -> bool:
...
-) -> Iterator[ContentItem] | list[ContentItem]:
+) -> list[ContentItemV1]:
     result = iter(content_list)
     if order_by == "last_deployed":
         pass  # do nothing, content is ordered by last_deployed by default
     elif order_by == "created":
         result = sorted(result, key=lambda c: c["created_time"], reverse=True)
 
-    return result
+    return list(result)
```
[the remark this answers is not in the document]
wch: I just tried that change, but it turns out `order_by` could be `None`, in which case `result` would still be an `Iterator`. Thanks to the type checker for noticing the issue! :)

image

rsconnect/api.py
```diff
@@ -807,40 +875,45 @@ def deploy_bundle(
         visibility = visibility or self.get("visibility")
 
         if isinstance(self.remote_server, RSConnectServer):
-            result = self.client.deploy(
+            # If we got here, the client must have been a RSConnectClient.
```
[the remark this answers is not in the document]
wch: Sounds good, I've replace this `cast()` and similar ones with an `isinstance()` check.

rsconnect/bundle.py
```diff
@@ -465,7 +465,7 @@ def buffer_checksum(buf: str | bytes) -> str:
 def to_bytes(s: str | bytes) -> bytes:
     if isinstance(s, bytes):
         return s
-    elif hasattr(s, "encode"):
+    elif isinstance(s, str):
```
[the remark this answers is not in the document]
wch: 🎉

rsconnect/api.py
```diff
@@ -203,45 +246,68 @@ def _tweak_response(self, response: HTTPResponse) -> JsonData | HTTPResponse:
...
         return self.post(
             "applications/%s" % app_id,
             body={"access_type": access, "id": app_id, "needs_config": False},
         )
 
-    def app_config(self, app_id: str):
-        return self.get("applications/%s/config" % app_id)
+    def app_config(self, app_id: str) -> ConfigureResult:
+        response = cast(ConfigureResult | HTTPResponse, self.get("applications/%s/config" % app_id))
+        response = self._server.handle_bad_response(response)
```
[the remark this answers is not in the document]
wch: Sounds good -- the places where `handle_bad_response` is used is a bit awkward and it'll be nice to clean it up.

rsconnect/api.py
```diff
@@ -536,17 +576,19 @@ def __init__(
         secret: Optional[str] = None,
         timeout: int = 30,
         logger: Optional[logging.Logger] = console_logger,
-        **kwargs: Any,
+        extra_kwargs: ExecutorKwargs | None = None,
```
mmarchetti, line 612: Can we add the rest of the arguments as optionals?

rsconnect/api.py
```diff
@@ -536,17 +576,19 @@ def __init__(
...
+        if extra_kwargs is None:
+            extra_kwargs = {}
+        self.state: ExecutorState = {**extra_kwargs}
         self.logger: logging.Logger | None = logger
         self.ctx = ctx
         self.setup_remote_server(
             ctx=ctx,
             name=name,
-            url=url or kwargs.get("server"),
+            url=url or extra_kwargs.get("server"),
```
mmarchetti, line 624: then we don't need to do this

rsconnect/api.py
```diff
@@ -667,19 +727,14 @@ def setup_client(self, cookies: Optional[CookieJar] = None, **kwargs: object):
...
-        return kwargs.get(key) or self.state.get(key)
-
     def pipe(self, func: Callable[P, T], *args: P.args, **kwargs: P.kwargs):
         return func(*args, **kwargs)
 
     @cls_logged("Validating server...")
     def validate_server(
         self,
+        # TODO: In practice, these arguments are not used. Should they be removed?
+        # And same for the downstream functions that are called from this function.
```
mmarchetti, line 770: Yes. We can instantiate this object with everything it needs. If anyone wants to validate a different server, they can create a different object and call its `validate_server` method.

rsconnect/api.py
```diff
@@ -796,17 +851,21 @@ def validate_posit_server(
...
-        )
-        app_id = self.get("app_id", **kwargs)
-        title = self.get("title", **kwargs)
-        app_store = self.get("app_store", *args, **kwargs)
+    def make_bundle(
+        self,
+        func: Callable[..., IO[bytes]],
+        *args: object,
+        **kwargs: object,
+        # These are the actual kwargs that appear to be present in practice
```
mmarchetti, line 892: Let's make these named arguments and stop accepting `**kw`.

rsconnect/api.py
```diff
@@ -520,6 +520,46 @@ class ServerDetails(TypedDict):
     python: ServerDetailsPython
 
 
+class ExecutorKwargs(TypedDict):
```
mmarchetti, line 556: I'd be very happy if no functions in this module accepted `**kwargs`.

rsconnect/api.py
```diff
@@ -520,6 +520,46 @@ class ServerDetails(TypedDict):
...
+    exclude: NotRequired[tuple[str, ...]]
+    new: NotRequired[bool]
+    app_id: NotRequired[str | None]
+    title: NotRequired[str | None]
+    visibility: NotRequired[str | None]
+    disable_env_management: NotRequired[bool | None]
+    env_vars: NotRequired[dict[str, str]]
+
+
+class ExecutorState(ExecutorKwargs, TypedDict):
```
mmarchetti, line 572: I'm hopeful that these can become normal members of the executor class, and drop the concept of the state dictionary.

rsconnect/api.py
wch: Sure thing.

rsconnect/api.py
wch: Agreed, that is a good goal.

rsconnect/api.py
wch: Done.

rsconnect/api.py
wch: I just realized that removing the `kwargs` here is not so straightforward, since `func` may or may not accept the specific arguments, like `multi_notebook`. I've added a `ParamSpec` so at least the `args` and `kwargs` will be checked where `make_bundle()` is called.

But looking at how the code is used, I think we can remove this very dynamic use of `func`.

This is a typical way that `make_bundle` is used right now:
```
        ce.make_bundle(
            make_html_bundle,
            path,
            entrypoint,
            extra_files,
            exclude,
        )
```

But looking at the content of this function, I think it can be restructured to be used this way, where `make_bundle()` just gets the built bundle instead of all those args:

```py
        .make_bundle(
            make_html_bundle(
                path,
                entrypoint,
                extra_files,
                exclude,
            )
        )
```

Although there also needs to be some error handling for `make_html_bundle()`.

But also, it looks like there's some more cleanup and refactoring possible for `make_bundle()`.

rsconnect/api.py
wch: My comment was incorrect -- these arguments are used in the downstream functions. The docstring was in the wrong function, though: it was in `validate_connect_server()` below but it documents the params for this function, so I've moved it back to this function.

tests/test_main.py
```diff
@@ -113,6 +112,7 @@ def test_deploy_manifest_shinyapps(self):
             httpretty.GET,
             "https://api.shinyapps.io/v1/users/me",
             body=open("tests/testdata/rstudio-responses/get-user.json", "r").read(),
+            adding_headers={"Content-Type": "application/json"},
```
wch, line 115: For tests to pass, I added this `"Content-Type": "application/json"` header to some of the mock HTTP endpoints. I believe that's the correct behavior to mock the endpoint.

wch:
This is ready for a review and then we can merge it into the `type-cleanup` branch.

There are still some typing issues that remain to be fixed, but the vast majority of the code now is fully typed. I think it's better to merge this sooner than later, because the typing updates will benefit other developers, and if we leave this branch open for a long time, there will be merge conflicts. After this is merged, then we can gradually fix the remaining issues.

Note that the pyright test in the Makefile is currently run with a leading `-`, so that even if it fails, `make` will not interpret it as an error, and CI will keep running. Once the typing work is completed, this can be removed so that any pyright failures will lead to CI failures.

rsconnect/api.py
```diff
@@ -62,7 +88,43 @@ def __init__(self, url: str, remote_name: str):
         self.url = url
         self.remote_name = remote_name
 
-    def handle_bad_response(self, response: HTTPResponse | object) -> None:
+    @overload
+    def handle_bad_response(self, response: HTTPResponse, is_httpresponse: Literal[True]) -> HTTPResponse: ...
+
+    @overload
+    def handle_bad_response(self, response: HTTPResponse | T, is_httpresponse: Literal[False] = False) -> T: ...
```
mmarchetti, line 95: Can you clarify the purpose of the overloads vs. just having the one method with a boolean parameter?

mmarchetti (approved):

rsconnect/api.py
wch: The purpose of the overloads is for type narrowing. Keep in mind that this function takes a `response` object and either throws or returns that same object.

- When `is_httpresponse=False` (the default), then the type checker can be certain that, the returned object is _not_ a `HTTPResponse`. (If the object was an `HTTPResponse`, the function throws.) In all cases where this is used, the input type is either a `TypedDict` or a `HTTPResponse`, and so when it returns, we can be certain that the result is the `TypedDict`.
- When `is_httpresponse=True`, then the returned object _could_ be an `HTTPResponse`. I added this because there are some instances where the input is expected to be an `HTTPResponse`. These are cases where the reponse status is 2xx, but the payload is _not_ JSON; instead, it may be the contents of a file.

This stuff is sort of necessary because the `_do_request` method may return either an HTTPResponse or, if the reponse body is JSON, it will convert that to a dict and return that. The mix of possible return values makes some things downstream of that a bit awkward. I think that code would benefit from being restructured in a future PR.

PR #572, opened 2024-04-05
Integration Flakey Test Fixes

kgartland-rstudio:
The Intergration Tests have been flakey lately. The first issue was, the docker container wasn't getting the RSC_LICENSE variable for Connect so Connect never started.
Like here:
https://github.com/rstudio/rsconnect-python/actions/runs/8567414234/job/23488396171#step:6:12

I moved the `RSC_LICENSE` variable up in the Github Actions workflow and that seems to have solved the license issue.  However, I still saw some race conditions where the test would start before Connect was ready, so I added a script to wait for Connect to start.  There were also some warnings about password security when the users were being created, so I updated those as well.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -175,6 +175,8 @@ jobs:
   test-rsconnect:
     name: "Integration tests against latest Connect"
     runs-on: ubuntu-latest
+    env:
+      RSC_LICENSE: ${{ secrets.RSC_LICENSE }}
     steps:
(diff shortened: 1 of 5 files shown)


kgartland-rstudio:
I ran it several times and it's still failing occasionally with nothing in the Connect logs:
https://github.com/rstudio/rsconnect-python/actions/runs/8572474436/job/23495692721

nealrichardson:
Testing has evolved quite a bit since 2024, closing this.

PR #577, opened 2024-04-06
Fix path for deploying notebooks and manifests

### LINKED ISSUE #575: Deploy notebook and voila create a `rsconnect-python` record in PWD rather than deployment directory
We now create an `rsconnect-python` directory for notebooks and voila in the PWD, previously those were stored as a per-deploy record here: ~/Library/Application Support/rsconnect-python/applications/

Each deployment would it get its own deployment json file, i.e. BWRzKUJz949Cz64x0DGJgG0_q84.json. Now we store the deployment records in an `rsconnect-python` folder in the PWD.

So when you deploy mutliple notebooks or voila apps from a parent directory, they both try to read/write to the same file so you get collisions.

I would expected the `rsconnect-python` record to be created in the directory of the deployment file(s) rather than the PWD to avoid such collisions.

wch:
This fixes #575, where, when deploying notebooks and voila apps (which are a form of notebook), the `rsconnect-python/` directory would be created in the wrong place.

In the code prior to the typing fixes, the `deploy_notebook()` function has a parameter named `file`, which got passed to `RSConnectExecutor()` via `kwargs`.
https://github.com/rstudio/rsconnect-python/blob/1.23.0/rsconnect/main.py#L898-L939

Later on, inside of `RSConnectExecutor.validate_app_mode()`, it did a `.get()` on `"path"`, `"file"`, `"file_name"`, etc. from the `kwargs` that were passed to the `RSConnectExecutor()` constructor, and used the first one that matched.  Then it used the resulting path when it created the `AppStore()`, and that's where the directory was set that caused the problem described in #575.
https://github.com/rstudio/rsconnect-python/blob/1.23.0/rsconnect/api.py#L857-L868

In the changes that I made when adding type annotations, I made all of the argument passing explicit instead of using `kwargs`, and inside of the `RSConnectExecutor()` constructor, I made it so that it explicitly stored each of the values that were passed in.
https://github.com/rstudio/rsconnect-python/blob/550b3733335d060d3cf9c360e2e598e8cb171699/rsconnect/main.py#L933-L945

As part of the change, instead of accepting args named `path`, `file`, `file_name`, etc., the `RSConnectExecutor()` constructor expected just `path`.
https://github.com/rstudio/rsconnect-python/blob/550b3733335d060d3cf9c360e2e598e8cb171699/rsconnect/api.py#L562-L587

But in those changes, I had missed that `deploy_notebook()` needed to call `RSConnectExecutor()` with `path=file`. That's the change I've made here.


rsconnect/main.py
```diff
@@ -1147,7 +1148,7 @@ def deploy_manifest(
         account=account,
         token=token,
         secret=secret,
-        path=file_name,
+        path=file,
```
wch, line 1151: This is for `deploy_manifest()`. In the previous code, the `kwargs` passed to `RSConnectExecutor(**kwargs)` contained both `file` and `file_name`:
https://github.com/rstudio/rsconnect-python/blob/1.23.0/rsconnect/main.py#L1094-L1121

And then when `validate_app_mode()` was called, it would look first for `path`, then `file`, then `file_name`. With that ordering, `file` took precedence over `file_name`.
https://github.com/rstudio/rsconnect-python/blob/1.23.0/rsconnect/api.py#L857-L868

This change makes it more consistent with the previous behavior.

kgartland-rstudio:
It looks like `deploy notebook` is now creating the rsconnect-python directory back at the old path: `~/Library/Application\ Support/rsconnect-python/applications/` so there should be no conflicts there.

However `deploy voila` is still creating the rsconnect-python directory in the PWD rather than the application directory.  After deploying voila from a parent directory I get an rsconnect-python dir here:

version: `1.23.1.dev88+g4c6f3dc`
```
> tree
.
├── python-voila
│   ├── index.ipynb
│   ├── manifest.json
│   ├── requirements.txt
│   └── voila.json
├── rsconnect-python
│   └── notebook-test.json
```

On the previous version of rsconnect-python, I get the rsconnect-python record in the application directory as expected.

version: `1.23.0`
```
> tree
.
├── python-voila
│   ├── index.ipynb
│   ├── manifest.json
│   ├── requirements.txt
│   ├── rsconnect-python
│   │   └── python-voila.json
│   └── voila.json
```

wch:
Oops, after all that text I wrote, it looks like I still missed passing the `path` for voila apps.

I've just pushed a fix.

kgartland-rstudio:
> Oops, after all that text I wrote, it looks like I still missed passing the `path` for voila apps.
> 
> I've just pushed a fix.

Ah that did it.  Looks good!

kgartland-rstudio (approved):
Looks good, verified fix on both `notebook` and `voila`.

PR #581, opened 2024-04-12
Some typing fixes

wch:
This PR makes some typing fixes, and also adds some runtime checks that `ce.remote_server` is a `RSConnectServer`.

### DIFF
--- a/rsconnect/main.py
+++ b/rsconnect/main.py
@@ -643,6 +643,8 @@ def details(
     set_verbosity(verbose)
 
     ce = RSConnectExecutor(ctx, name, server, api_key, insecure, cacert).validate_server()
+    if not isinstance(ce.remote_server, RSConnectServer):
+        raise RSConnectException("This command requires a Posit Connect server.")
 
     click.echo("    Posit Connect URL: %s" % ce.remote_server.url)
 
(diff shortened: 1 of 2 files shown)


rsconnect/main.py
```diff
@@ -2215,6 +2210,8 @@ def content_describe(
     output_params(ctx, locals().items())
     with cli_feedback("", stderr=True):
         ce = RSConnectExecutor(ctx, name, server, api_key, insecure, cacert, logger=None).validate_server()
+        if not isinstance(ce.remote_server, RSConnectServer):
+            raise RSConnectException("This command requires a Posit Connect server.")
```
wch, line 2214: I used a pretty generic exception message here, but am willing to customize it for each command, if you think that would help.

For example, in this case, it could be:

```
"`rsconnect content describe` requires a Posit Connect server."
```

rsconnect/main.py
```diff
@@ -101,7 +106,7 @@
 def cli_exception_handler(func: Callable[P, T]) -> Callable[P, T]:
     @wraps(func)
     def wrapper(*args: P.args, **kwargs: P.kwargs):
-        def failed(err: str):
+        def failed(err: str) -> Never:
```
wch, line 109: The explicit `Never` return helps pyright recognize that the calls to `failed()` will never return.

mmarchetti (approved):
I'm OK with the more generic messages, but the specific ones would be a nice touch.

PR #583, opened 2024-04-17
Pass `path` to `RSConnectExecutor.fromConnectServer()`

### LINKED ISSUE #582: deploy html creates an rsconnect-python record in PWD
It looks like we have one more case of this https://github.com/rstudio/rsconnect-python/issues/575.

When you `deploy html ...` the rsconnect-python record is created in the present directory rather than the deployment directory.

wch:
This should fix #582.

### DIFF
--- a/rsconnect/main.py
+++ b/rsconnect/main.py
@@ -1390,6 +1390,7 @@ def deploy_html(
             account=account,
             token=token,
             secret=secret,
+            path=path,
             server=server,
             exclude=exclude,
             new=new,
--- a/rsconnect/main.py
+++ b/rsconnect/main.py
@@ -1408,6 +1408,7 @@ def deploy_html(
             account=account,
             token=token,
             secret=secret,
+            path=path,
             server=server,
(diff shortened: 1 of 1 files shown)


kgartland-rstudio:
I'm still seeing the same behavior:
```
rsconnect version
1.23.1.dev97+g2460f7d
```

Before deploy:
```
> tree
.
└── test
    └── index.html
```
then run:
```> rsconnect deploy html test/index.html -n dogfood```

After deploy:
```
> tree
.
├── rsconnect-python
│   └── html-test.json
└── test
    └── index.html
```

In rsconnect-python 1.23.0, the `rsconnect-python` record is stored at `~/Library/Application Support/rsconnect-python/applications/`.

wch:
Shoot, sorry about that. I had only passed the `path` argument to one of the two code paths. Should be fixed now. Sorry I missed that -- I was a bit too hasty there.

kgartland-rstudio (approved):
Looks good, thanks!

PR #588, opened 2024-05-10
Add IMAGE_OS variable

kgartland-rstudio:
This PR just adds an IMAGE_OS variable to avoid running a script that can only be run in Connect.

I also updated `just` while I was in here.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -261,7 +261,7 @@ jobs:
           - 3.8.10
           - 3.9.5
     steps:
-          - uses: extractions/setup-just@v1
+          - uses: extractions/setup-just@v2
             env:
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -303,6 +303,7 @@ jobs:
 
               # This allows us to start Connect separately in our own docker container
               CONNECT_SERVER: "http://localhost:3939"
+              IMAGE_OS: "ubuntu22"
```
aronatkins, line 306: `IMAGE_OS` should be something like `rstudio/connect:ubuntu22`. It's `TARGET_OS` that contains just the distribution.

aronatkins (changes requested):

aronatkins (approved):

PR #589, opened 2024-05-13
deploy/write-manifest tensorflow

aronatkins:
## Intent

Allow manifests/deploys of TensorFlow models.

Fixes https://github.com/rstudio/connect/issues/26884

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach
Add entrypoints for TensorFlow commands. Removed R/Python/image arguments in situations where they do not apply (e.g. static content).

## Automated Tests
Test assembly of manifest and bundle; will add integration tests separately.

## Checklist

### DIFF
--- a/rsconnect/main.py
+++ b/rsconnect/main.py
@@ -61,6 +61,7 @@ from .bundle import (
     make_notebook_html_bundle,
     make_notebook_source_bundle,
     make_voila_bundle,
+    make_tensorflow_bundle,
     read_manifest_app_mode,
     validate_entry_point,
     validate_extra_files,
@@ -70,6 +71,7 @@ from .bundle import (
     write_environment_file,
     write_notebook_manifest_json,
     write_quarto_manifest_json,
+    write_tensorflow_manifest_json,
     write_voila_manifest_json,
 )
 from .environment import EnvironmentException
@@ -971,9 +973,6 @@ def deploy_notebook(
             python,
             hide_all_input,
             hide_tagged_input,
-            image=image,
-            env_management_py=env_management_py,
-            env_management_r=env_management_r,
         )
     else:
         ce.make_bundle(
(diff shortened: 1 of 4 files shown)


rsconnect/main.py
```diff
@@ -1325,6 +1324,98 @@ def deploy_quarto(
...
 
 
+# noinspection SpellCheckingInspection,DuplicatedCode
+@deploy.command(
+    name="tensorflow",
+    short_help="Deploy TensorFlow models to Posit Connect [v2024.05.0+].",
+    help=(
+        "Deploy a TensorFlow model to Posit Connect. Requires Posit Connect 2024.05.0 or later."
+        "\n\n"
+        "DIRECTORY is the path containing a TensorFlow model."
```
toph-allen, line 1334: Should `DIRECTORY` be in all caps here? If this is just a convention I'm unaware of, ignore this.

tests/test_bundle.py
```diff
@@ -2079,90 +2065,6 @@ def test_create_html_manifest():
     )
     assert single_file_index_file_ans == json.loads(manifest.get_flattened_copy().json)
 
-    # check all runtime_environment vars
-    single_file_index_file_ans = {
```
toph-allen, line 2083: What were these tests doing?

toph-allen (approved):

tests/test_bundle.py
aronatkins: These tests were all trying to confirm that R/Python environment and image fields were retained for static (HTML) content. We do not execute HTML content, so these fields are unnecessary. I removed them as inputs and removed the tests as appropriate.

rsconnect/main.py
aronatkins: The all-caps is a convention we used for Quarto content commands. Its presence feels a little redundant, but folks found it helpful during early testing.

PR #590, opened 2024-05-13
More type fixes

wch:
This PR makes more typing fixes.

NOTE: This PR also removes use of the deprecated function `datetime.datetime.utcnow()`. This will result in a behavior change in logging in one place, where instead of writing `2024-05-13 17:03:09.857000`, it will write `2024-05-13 17:03:09.857000+00:00`. If this is undesirable, I can get it to format the string without the `+00:00`.


rsconnect/api.py
```diff
@@ -1347,7 +1347,7 @@ def _tweak_response(self, response: HTTPResponse) -> JsonData | HTTPResponse:
     def get_extra_headers(self, url: str, method: str, body: str | bytes):
         canonical_request_method = method.upper()
         canonical_request_path = parse.urlparse(url).path
-        canonical_request_date = datetime.datetime.utcnow().strftime("%a, %d %b %Y %H:%M:%S GMT")
+        canonical_request_date = datetime.datetime.now(datetime.timezone.utc).strftime("%a, %d %b %Y %H:%M:%S GMT")
```
wch, line 1350: Note that this change will not result in changed output, because `strftime()` is being called with a format string.

rsconnect/environment.py
```diff
@@ -169,7 +169,10 @@ def pip_freeze():
     pip_stdout = filter_pip_freeze_output(pip_stdout)
 
     pip_stdout = (
-        "# requirements.txt generated by rsconnect-python on " + str(datetime.datetime.utcnow()) + "\n" + pip_stdout
+        "# requirements.txt generated by rsconnect-python on "
+        + str(datetime.datetime.now(datetime.timezone.utc))
```
wch, line 173: This will result in `+00:00` being added to the output.

rsconnect/metadata.py
```diff
@@ -721,7 +731,7 @@ def update_content_item_last_build_time(self, guid: str, defer_save: bool = Fals
         """
         with self._lock:
             content = self.get_content_item(guid)
-            content["rsconnect_last_build_time"] = datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ")
+            content["rsconnect_last_build_time"] = datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
```
wch, line 734: Note that this change will not result in changed output, because strftime() is being called with a format string.

rsconnect/api.py
```diff
@@ -16,6 +16,7 @@
 import typing
 import webbrowser
 from os.path import abspath, dirname
+from ssl import SSLError
```
wch, line 19: `ssl.SSLError` is the same object as `_ssl.SSLError`.

See:

https://github.com/python/cpython/blob/v3.8.0/Lib/ssl.py#L102-L105

https://github.com/python/cpython/blob/v3.12.3/Lib/ssl.py#L104-L107See

wch:
The GitHub Actions workflows don't seem to be running on the latest commit. Any idea why? (It won't let me merge until they run.)

nealrichardson:
> The GitHub Actions workflows don't seem to be running on the latest commit. Any idea why? (It won't let me merge until they run.)

It's because I renamed the default branch to `main`, but the workflow only is triggered on `master`. Sorry. Do you want to fix that here or want me to make a separate patch for that?

nealrichardson:
Ironically, the workflow is called `main` :joy: https://github.com/posit-dev/rsconnect-python/blob/main/.github/workflows/main.yml#L4-L9

PR #599, opened 2024-05-29
Let it build under Fedora-39

tuxmaster5000:
## Intent
Use it under Fedora-.39

## Type of Change
Version of required modules

 [X ] Bug Fix


mmarchetti:
We've opened a repo-local PR (https://github.com/posit-dev/rsconnect-python/pull/607) which will have access to the resources needed to pass CI.

PR #600, opened 2024-06-03
add commands to list and fetch jumpstart examples

edavidaja:


aronatkins:
related: https://github.com/rstudio/rsconnect/issues/840

nealrichardson:
Thoughts on adding some basic test coverage for this? Seems like a useful thing to have, but I'd feel more comfortable with :shipit: if we had some assurance that it was doing what we expected.

PR #602, opened 2024-06-03
Update "rsconnect content build" sub-command docs in the readme

dbkegley:
## Intent

Update readme to be more clear about what to do when local state becomes out of sync with server side state during a `content build` operation.

## Type of Change

- [ ] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change   
- [x] docs

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -800,7 +800,10 @@ Commands:
 ```
 
 To build a specific content item, first `add` it to the list of content that is
-"tracked" for building using its GUID.
+"tracked" for building using its GUID. Content that is "tracked" in the local state
+may become out-of-sync with what exists remotely on the Connect server (the result of
+`rsconnect content search`). When this happens, it is safe to remove the locally tracked
+entries with `rsconnect content build rm`.
 
 > **Note**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -800,7 +800,10 @@ Commands:
 ```
 
 To build a specific content item, first `add` it to the list of content that is
-"tracked" for building using its GUID.
+"tracked" for building using its GUID. Content that is "tracked" in the local state
+may become out-of-sync with what exists remotely on the Connect server (the result of
+`rsconnect content search`). When this happens, it is safe to remove the locally tracked
+entries with `rsconnect content build rm`.
```
aronatkins, line 806: Is there any other documentation for `rsconnect content build rm`?

aronatkins (commented):
Could we document `content build rm` alongside `content build add`?

aronatkins (approved):
Maybe add an example showing `add --guid ABCD` and `rm --guid ABCD` to help folks see that one undoes the other.

PR #615, opened 2024-08-15
add ability to override python version

cfhammill:
## Intent

I am unable to deploy to connect from my development server because the python I'm using is newer than the versions available on connect. I would like to be able to tell rsconnect-python to use a different python version, so I patched in that feature.

## Type of Change

- [x] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach

This PR modifies the environment detection code to add an override option where the `Environment` object is cloned
with the updated python version. This was weaved out to the CLI.

## Automated Tests

No tests written, would like feedback on idea / style / etc. before adding tests. Someone more familiar with the project could also provide direction for testing this feature.

## Directions for Reviewers

Try deploying something and using `--override-python-version <version>` or `-op <version>`

## Checklist

### DIFF
--- a/rsconnect/bundle.py
+++ b/rsconnect/bundle.py
@@ -1732,6 +1732,7 @@ def get_python_env_info(
     file_name: str,
     python: str | None,
     force_generate: bool = False,
+    override_python_version: str | None = None,
 ) -> tuple[str, Environment]:
     """
     Gathers the python and environment information relating to the specified file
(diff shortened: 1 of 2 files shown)


mmarchetti:
Thank you for contributing this PR. I've opened #617 on a local branch so that CI can pass; we'll add documentation there and do some testing before we merge.

PR #624, opened 2024-12-17
update supported Connect version for gradio apps

edavidaja:


tdstein (approved):
Why was it previously 2024.11.0?

edavidaja:
Support for the app mode landed in 2024.11 but some further changes to make Gradio app serving more robust landed after. There have also been some changes in the framework such that 2024.11 _might_ work, but we prefer for people to use 2024.12+.

PR #629, opened 2024-12-18
fix: add validation for `system caches delete` flags

lucasrod16:
## Intent

Fixes https://github.com/posit-dev/rsconnect-python/issues/628

## Type of Change

- [x] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change   

## Approach

Use click's built-in validation, `required=True`

## Automated Tests

Added tests to the `TestSystemCachesDelete` class

## Checklist

### DIFF
--- a/rsconnect/main.py
+++ b/rsconnect/main.py
@@ -2854,8 +2854,8 @@ def system_caches_delete(
     insecure: bool,
     cacert: Optional[str],
     verbose: int,
-    language: Optional[str],
-    version: Optional[str],
+    language: str,
+    version: str,
     image_name: str,
     dry_run: bool,
 ):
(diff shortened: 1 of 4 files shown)


rsconnect/main.py
```diff
@@ -2825,11 +2825,13 @@ def system_caches_list(
     "--language",
     "-l",
     help="The language of the target cache.",
+    required=True,
 )
 @click.option(
     "--version",
     "-V",
     help="The version of the target cache.",
+    required=True,
```
aronatkins, line 2834: The `system_caches_delete()` function claims that `language` and `version` have `Optional[..]`. The `delete_runtime_cache()` function does the same. Should we update them to not have `Optional[..]` typing?

aronatkins (approved):
CC @toph-allen - do you recall why we did not mark the language and version arguments as required? 

The Connect API enforces the presence of language, version, and image name.

dbkegley (approved):

PR #637, opened 2025-02-18
chore(ci): Remove IMAGE_OS setting

### LINKED ISSUE #633: Use public Docker image for testing
CI should be able to use a similar setup to what posit-sdk-py does for testing against live Connect, using the public Docker images for releases and previews: https://github.com/posit-dev/posit-sdk-py/blob/main/.github/workflows/ci.yaml#L42

nealrichardson:
Closes #633. Turns out we already are using public images for testing, and from what I infer from #588, this was just being added to prevent a script from looking for it and erroring if it wasn't set. Due to changes in the Connect source, this variable isn't being used anymore, so we can drop the reference.

### DIFF
--- a/.github/workflows/deploy_tests.yml
+++ b/.github/workflows/deploy_tests.yml
@@ -13,7 +13,7 @@ jobs:
             fetch-depth: 0
           env:
             GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-            
+
         - uses: actions/setup-python@v4
           with:
             python-version: 3.8.x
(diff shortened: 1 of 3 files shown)


nealrichardson:
The integration testing job that was modified in this PR passes: https://github.com/posit-dev/rsconnect-python/actions/runs/13401139629/job/37437096302?pr=637

IDK what's up with the `test-jupyter` job, but it's failing on the nightly `main` run too: https://github.com/posit-dev/rsconnect-python/actions/runs/13386951455

Seems to be failing since January 29: https://github.com/posit-dev/rsconnect-python/actions/runs/13027921854

edavidaja (approved):

Issue #636, opened 2025-02-11
Link directly to the deployment log when a deployment has failed


toph-allen:
> - When a (Python?) deployment fails (from rsconnect-python?), we don't surface the logs, or a link to the logs even, so it's hard to debug.
> - It returns the content URL regardless, so when you go there, you just get a blank page or other error display.

When `rsconnect-python` was implemented, you couldn't link directly to logs in the Connect dashboard. You can now, so when a deployment fails, we should construct the URL to the log for the current deployment task and print it, rather than printing the link to the content URL.

Issue #256, opened 2022-06-15
refactor issues


kgartland-rstudio:
This ticket is a placeholder for issues/questions/comments about the refactor.

### Issues
- [x] [deploying content without a server results in a NoneType error](https://github.com/rstudio/rsconnect-python/issues/256#issuecomment-1156531151)
- [x] [when trying to hit a server without Connect installed we don't catch and handle the error](https://github.com/rstudio/rsconnect-python/issues/256#issuecomment-1156544323)
- [x] [we don't catch the error when deploying content with an unavailable package.](https://github.com/rstudio/rsconnect-python/issues/256#issuecomment-1156565707)
- [x] [The `--static` flag is ignored when deploying a static notebook](https://github.com/rstudio/rsconnect-python/issues/256#issuecomment-1156874531)
- [x] [We aren't catching the mismatched python version](https://github.com/rstudio/rsconnect-python/issues/256#issuecomment-1156877817)
- [x] [when deploying api, from outside the api directory, extra files cannot be included](https://github.com/rstudio/rsconnect-python/issues/256#issuecomment-1156889516)
- [x] [redeploying always creates new content](https://github.com/rstudio/rsconnect-python/issues/256#issuecomment-1158873085)
- [x] [deploying while specifying the `--app-id` always deploys using Python API mode](https://github.com/rstudio/rsconnect-python/issues/256#issuecomment-1158887047)

### Comments (nit-picks)
- [ ] [[OK] loglines no longer match up](https://github.com/rstudio/rsconnect-python/issues/256#issuecomment-1156894111)

kgartland-rstudio:
When trying to hit a server without Connect installed we don't catch and handle the error.

Actual:
```
> rsconnect deploy streamlit top-5-income-share-streamlit/ -N -s http://google.com -k blah
Validating server... 	[ERROR]: Received an unexpected response from RStudio Connect: 404 Not Found
Traceback (most recent call last):
  File "/Users/kgartland/.pyenv/versions/rsconnect-tag/bin/rsconnect", line 8, in <module>
    sys.exit(cli())
  File "/Users/kgartland/.pyenv/versions/3.8.2/envs/rsconnect-tag/lib/python3.8/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/Users/kgartland/.pyenv/versions/3.8.2/envs/rsconnect-tag/lib/python3.8/site-packages/click/core.py", line 1053, in main
[...]

PR #642, opened 2025-03-07
Add logs url on failure

jonkeane:
## Intent
Add the logs URL to the error message so that folks find the logs easier.

Resolves: #636

## Type of Change
- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach
Fetch and add the logs URL to the error message

## Automated Tests
None were broken or added.

## Directions for Reviewers
Output from testing this locally:

```
% rsconnect deploy shiny --name radixu .
Validating server... 	[OK]
Validating app mode... 	[OK]
Making bundle ... 	[OK]
Deploying bundle ... 	[OK]
Saving deployed information... 	[OK]
Building Python Shiny application...
...
Launching Python Shiny application...
	 Dashboard content URL: https://rsc.radixu.com/connect/#/apps/09bfbb6a-7e7c-4a37-bafd-9a0bd7b4b67c/access
	 Direct content URL: https://rsc.radixu.com/content/09bfbb6a-7e7c-4a37-bafd-9a0bd7b4b67c/
Verifying deployed content... 	[ERROR]: Could not access the deployed content. The app might not have started successfully.
	For more information: https://rsc.radixu.com/connect/#/apps/09bfbb6a-7e7c-4a37-bafd-9a0bd7b4b67c/logs
Error: Could not access the deployed content. The app might not have started successfully.
	For more information: https://rsc.radixu.com/connect/#/apps/09bfbb6a-7e7c-4a37-bafd-9a0bd7b4b67c/logs
```

## Checklist

### DIFF
--- a/rsconnect/api.py
+++ b/rsconnect/api.py
@@ -355,7 +355,7 @@ class RSConnectClient(HTTPServer):
             raise RSConnectException(
                 "Could not access the deployed content. "
                 + "The app might not have started successfully."
-                + f"\n\t For more information: {self.app_config(app_guid)['logs_url']}"
+                + f"\n\t For more information: {self.app_config(app_guid).get('logs_url')}"
             )
 
     def bundle_download(self, content_guid: str, bundle_id: str) -> HTTPResponse:
--- a/rsconnect/api.py
+++ b/rsconnect/api.py
@@ -354,8 +354,8 @@ class RSConnectClient(HTTPServer):
         if self.is_app_failed_response(response):
             raise RSConnectException(
                 "Could not access the deployed content. "
-                + "The app might not have started successfully. "
-                + "Visit it in Connect to view the logs."
+                + "The app might not have started successfully."
+                + f"\n\tFor more information: {self.app_config(app_guid)['logs_url']}"
             )
 
     def bundle_download(self, content_guid: str, bundle_id: str) -> HTTPResponse:
--- a/rsconnect/api.py
+++ b/rsconnect/api.py
@@ -355,7 +355,7 @@ class RSConnectClient(HTTPServer):
             raise RSConnectException(
                 "Could not access the deployed content. "
                 + "The app might not have started successfully."
-                + f"\n\tFor more information: {self.app_config(app_guid)['logs_url']}"
+                + f"\n\t For more information: {self.app_config(app_guid)['logs_url']}"
             )
 
     def bundle_download(self, content_guid: str, bundle_id: str) -> HTTPResponse:


jonkeane:
I will add changelog after initial review / confirmation this looks like the direction we want.

rsconnect/api.py
```diff
@@ -354,8 +354,8 @@ def app_access(self, app_guid: str) -> None:
         if self.is_app_failed_response(response):
             raise RSConnectException(
                 "Could not access the deployed content. "
-                + "The app might not have started successfully. "
-                + "Visit it in Connect to view the logs."
+                + "The app might not have started successfully."
+                + f"\n\tFor more information: {self.app_config(app_guid)['logs_url']}"
```
jonkeane, line 358: New line + tab to match the other URLs that get put in the output, but that's not strictly necessary.

nealrichardson:
> Output from testing this locally:

Looks like the error message is printed twice? And is indented one space less than the other messages?

jonkeane:
> Looks like the error message is printed twice? 

Yeah, it is. I tried to find where/why this was once, but didn't get to the bottom of it and wanted to get this up. I will keep looking though (or at least create an issue to fix it). The duplication is *not* a result of these changes though.

> And is indented one space less than the other messages?

Ah, right indeed. those other messages have `"\t D" and this one has "\tF". I'll match the other ones for now (so this chang ei s less disruptive) but think if we're going to `\t` we shouldn't also space immediately after.

jonkeane:
> > Looks like the error message is printed twice?

> Yeah, it is. I tried to find where/why this was once, but didn't get to the bottom of it and wanted to get this up. I will keep looking though (or at least create an issue to fix it). The duplication is not a result of these changes though.

Ok, with some more digging, I think this was introduced in #244 and ultimately what happens is the [logger is printed to](https://github.com/posit-dev/rsconnect-python/blob/bf9ade854a7b4ed483fa6033fb0cab1a490aaae4/rsconnect/log.py#L231) and then it is [printed to the console](https://github.com/posit-dev/rsconnect-python/blob/bf9ade854a7b4ed483fa6033fb0cab1a490aaae4/rsconnect/main.py#L112). Though I cannot find any discussion in the PR or linked [refactor issues issue](https://github.com/posit-dev/rsconnect-python/issues/256) about if this is the desired behavior and why. 

I would be supportive of (probably in a separate PR) exploring if the `cli_exception_handler` needs to exist at all.

rsconnect/api.py
```diff
@@ -354,8 +354,8 @@ def app_access(self, app_guid: str) -> None:
         if self.is_app_failed_response(response):
             raise RSConnectException(
                 "Could not access the deployed content. "
-                + "The app might not have started successfully. "
-                + "Visit it in Connect to view the logs."
+                + "The app might not have started successfully."
+                + f"\n\t For more information: {self.app_config(app_guid)['logs_url']}"
```
tdstein, line 358: The paranoid programmer in me says to avoid the possible `KeyError` before the exception is raised.

```suggestion
                + f"\n\tFor more information: {self.app_config(app_guid).get('logs_url')}"
```

jonkeane: (1) yes, agreed about the key error (2) the ` ` there is (sadly) intentional to avoid needing to `s/\t /\t/g` elsewhere

tdstein (approved):

jonkeane:
@tdstein any thoughts on:

> Ok, with some more digging, I think this was introduced in https://github.com/posit-dev/rsconnect-python/pull/244 and ultimately what happens is the [logger is printed to](https://github.com/posit-dev/rsconnect-python/blob/bf9ade854a7b4ed483fa6033fb0cab1a490aaae4/rsconnect/log.py#L231) and then it is [printed to the console](https://github.com/posit-dev/rsconnect-python/blob/bf9ade854a7b4ed483fa6033fb0cab1a490aaae4/rsconnect/main.py#L112). Though I cannot find any discussion in the PR or linked https://github.com/posit-dev/rsconnect-python/issues/256 about if this is the desired behavior and why.
> 
> I would be supportive of (probably in a separate PR) exploring if the cli_exception_handler needs to exist at all.

tdstein:
> @tdstein any thoughts on:
> 
> > Ok, with some more digging, I think this was introduced in #244 and ultimately what happens is the [logger is printed to](https://github.com/posit-dev/rsconnect-python/blob/bf9ade854a7b4ed483fa6033fb0cab1a490aaae4/rsconnect/log.py#L231) and then it is [printed to the console](https://github.com/posit-dev/rsconnect-python/blob/bf9ade854a7b4ed483fa6033fb0cab1a490aaae4/rsconnect/main.py#L112). Though I cannot find any discussion in the PR or linked #256 about if this is the desired behavior and why.
> > I would be supportive of (probably in a separate PR) exploring if the cli_exception_handler needs to exist at all.

This predates me as well. I think a separate PR to remove the duplication is correct. 

My guess is that the intent is for `logger` to write to stderr and for `click` to write to stdout. In this case, I can see why a log message and `print` call would be necessary. But, I think it's overkill for this package. It should either use [`click.echo`](https://click.palletsprojects.com/en/stable/quickstart/#echoing) or Python logging.

jonkeane:
The failures are on main as well, I made https://github.com/posit-dev/rsconnect-python/issues/644 for those.

PR #643, opened 2025-03-11
Foundation for detecting python version requirements from pyproject.toml

amol-:
## Intent

Prepare foundations for supporting detection of python version requirements when generating manifest.json.
This implements detect of which metadata file is available and parsing of the requirement from `pyproject.toml`.
Further changes will provide support for `setup.cfg` and `.python-version`, those will then be used to set the proper requirement field in the `manifest.json`.

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach

The support for adding python version constraint to `manifest.json` is being built in small incremental steps that won't change user facing behaviour until the final PR that will enable the new behaviour.

## Automated Tests

A new test suite in `test_pyproject.py` was added.

## Directions for Reviewers

There isn't a directly corresponding issue in the `rsconnect-python` repository as the issue governing this work is in Connect itself.


amol-:
The failure seems to be already happening on main: https://github.com/posit-dev/rsconnect-python/actions/runs/13663461265/job/38199851576

So marking this as ready for review

tests/test_pyproject.py
```diff
@@ -0,0 +1,55 @@
...
+
+from rsconnect.pyproject import lookup_metadata_file, parse_pyproject_python_requires
+
+import pytest
+
+HERE = os.path.dirname(__file__)
+PROJECTS_DIRECTORY = os.path.abspath(os.path.join(HERE, "testdata", "python-project"))
+
+
+@pytest.mark.parametrize(
```
sagerb, line 12: It's a nit, but adding a simple comment here that you are testing to make sure that tests against the sample directories return the expected files would be great.

tests/test_pyproject.py
```diff
@@ -0,0 +1,55 @@
...
+        (os.path.join(PROJECTS_DIRECTORY, "allofthem"), ("pyproject.toml", "setup.cfg", ".python-version")),
+    ],
+    ids=["pyproject.toml", "setup.cfg", ".python-version", "allofthem"],
+)
+def test_python_project_metadata_detect(project_dir, expected):
+    expectation = [(f, pathlib.Path(project_dir) / f) for f in expected]
+    assert lookup_metadata_file(project_dir) == expectation
+
+
+@pytest.mark.parametrize(
```
sagerb, line 33: another nit, same comment to add a simple comment.

tests/test_pyproject.py
```diff
@@ -0,0 +1,55 @@
...
+        os.path.join(PROJECTS_DIRECTORY, "empty"),
+        os.path.join(PROJECTS_DIRECTORY, "missing"),
+    ],
+    ids=["empty", "missing"],
+)
+def test_python_project_metadata_missing(project_dir):
+    assert lookup_metadata_file(project_dir) == []
+
+
+@pytest.mark.parametrize(
```
sagerb, line 45: and just one more. ;-)

sagerb (approved):
This looks great. I only had a little nit about some comments in the test file to better describe what you're doing. I was able to certainly extract what was going on within the code, but for maintenance concerns, I would recommend adding some simple comments as indicated. But I will leave that in your hands to decide!

amol-:
@sagerb I added the comments in https://github.com/posit-dev/rsconnect-python/pull/643/commits/5ae6d17eaaa6c97207f0dccd6500deca9fe2baeb it should now be more explicit about what the tests are testing 👍

PR #645, opened 2025-03-12
Parser for interpreter version requirements from setup.cfg .python-version 

amol-:
## Intent

Second step toward supporting python interpreter version requirements in `manifest.json` generation.
This is a follow-up of https://github.com/posit-dev/rsconnect-python/pull/643 that adds support for the other two planned file formats.

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach

Continuing with small incremental PRs approach to add support for each new format keeping the review effort minimal.

## Automated Tests

New tests were added for every function in the `test_pyproject.py` file.

## Directions for Reviewers

This continues in the direction of building all the necessary blocks without exposing the feature to the end user until the very end. The PR is based on the https://github.com/posit-dev/rsconnect-python/pull/643 code, so it should be reviewed only once the other PR is merged as it will influence this one.

### DIFF
--- a/rsconnect/pyproject.py
+++ b/rsconnect/pyproject.py
@@ -7,6 +7,7 @@ but not from setup.py due to its dynamic nature.
 
 import pathlib
 import typing
+import configparser
 
 try:
     import tomllib
@@ -24,12 +25,29 @@ def lookup_metadata_file(directory: typing.Union[str, pathlib.Path]) -> typing.L
     directory = pathlib.Path(directory)
 
     def _generate():
-        for filename in ("pyproject.toml", "setup.cfg", ".python-version"):
+        for filename in (".python-version", "pyproject.toml", "setup.cfg"):
             path = directory / filename
             if path.is_file():
                 yield (filename, path)
 
     return list(_generate())
+
+
+def get_python_requires_parser(
+    metadata_file: pathlib.Path,
+) -> typing.Optional[typing.Callable[[pathlib.Path], typing.Optional[str]]]:
+    """Given the metadata file, return the appropriate parser function.
+
+    The returned function takes a pathlib.Path and returns the parsed value.
+    """
+    if metadata_file.name == "pyproject.toml":
+        return parse_pyproject_python_requires
+    elif metadata_file.name == "setup.cfg":
+        return parse_setupcfg_python_requires
+    elif metadata_file.name == ".python-version":
+        return parse_pyversion_python_requires
+    else:
+        return None
 
 
 def parse_pyproject_python_requires(pyproject_file: pathlib.Path) -> typing.Optional[str]:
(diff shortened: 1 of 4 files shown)


rsconnect/pyproject.py
```diff
@@ -20,18 +21,38 @@ def lookup_metadata_file(directory: typing.Union[str, pathlib.Path]) -> typing.L
...
+    The returned function takes a pathlib.Path and returns the parsed value.
+    """
+    if metadata_file.name == "pyproject.toml":
+        return parse_pyproject_python_requires
+    elif metadata_file.name == "setup.cfg":
+        return parse_setupcfg_python_requires
+    elif metadata_file.name == ".python-version":
+        return parse_pyversion_python_requires
+    else:
+        return None
```
amol-, line 53: This could be done with a lookup table, but given the options aren't many I preferred to be explicit and avoid dynamic lookups with are harder to follow when debugging.

tests/test_pyproject.py
```diff
@@ -37,6 +44,23 @@ def test_python_project_metadata_detect(project_dir, expected):
     assert lookup_metadata_file(project_dir) == expectation
 
 
+@pytest.mark.parametrize(
```
sagerb, line 47: Maybe consider a comment describing this test?

sagerb (commented):
The code is looking great. I provided a few small suggestions which I think will improve the code. Reach out when you are ready for another review.

rsconnect/pyproject.py
```diff
@@ -20,18 +21,38 @@ def lookup_metadata_file(directory: typing.Union[str, pathlib.Path]) -> typing.L
...
-        for filename in ("pyproject.toml", "setup.cfg", ".python-version"):
+        for filename in (".python-version", "pyproject.toml", "setup.cfg"):
             path = directory / filename
             if path.is_file():
                 yield (filename, path)
 
     return list(_generate())
 
 
+def get_python_requires_parser(
```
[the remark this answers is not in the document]
amol-: Sounds good, I'll rename the function

tests/test_pyproject.py
amol-: There is one at line 58, do think it's not sufficiently informative?

rsconnect/pyproject.py
```diff
@@ -20,18 +21,38 @@ def lookup_metadata_file(directory: typing.Union[str, pathlib.Path]) -> typing.L
...
+    metadata_file: pathlib.Path,
+) -> typing.Optional[typing.Callable[[pathlib.Path], typing.Optional[str]]]:
+    """Given the metadata file, return the appropriate parser function.
+
+    The returned function takes a pathlib.Path and returns the parsed value.
+    """
+    if metadata_file.name == "pyproject.toml":
```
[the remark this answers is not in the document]
amol-: Given 

https://github.com/pypa/pip/blob/main/src/pip/_internal/req/req_install.py#L512-L514

and

https://github.com/pypa/pip/blob/main/src/pip/_internal/pyproject.py#L26-L27

I think it's fairly safe to assume that the filename is expected to be lowercase for pip too.

It's true that the PEP doesn't explicitly say that the filename must be case sensitive, but in every single place it explicitly write `pyproject.toml` lowercase, which I think it's safe to expect it means the filename is expected to always be lowercase and thus work in both case sensitive and case insensitive file systems.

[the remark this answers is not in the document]
sagerb: ok, thanks for considering my suggestion.

tests/test_pyproject.py
sagerb: Yes, sorry, I caught myself seeing that for a different test and suppressed my comment. This is fine.

PR #647, opened 2025-03-19
fix: removes deprecated docker compose field

tdstein:


tdstein:
@jonkeane - this gets rid of the warning message but doesn't fix CI. That's a bigger fix, which I'll do in a separate PR.

PR #648, opened 2025-03-20
Support environment.python.requires option in manifest

amol-:
## Intent

Connect supports python interpreter version requirements from the `environment.python.requires` field in the manifest.
That field should be filled based on the version requirements in the project metadata files:

- pyproject.toml
- setup.cfg
- .python-version

## Type of Change

- [x] New Feature       
- [x] Breaking Change   

## Approach

1. Refactor the various deploy commands to generate the `Environment` information via the same code path
2. Implement support for detecting the interpreter requirement in that shared entry point
3. Write the detected constraint to the generated manifest before submitting it to the server.

## Automated Tests

- Existing test suite checks that the support for current deploys is still working as expected
- New tests check that the manifest.json gets generated with the right `environment.python.requires` field when detected

## Directions for Reviewers

- `--override-python-version` still allows overwriting the detected version, but on long term the plan is to deprecate the option and ask users to create a `.python-version` file to do the same. So that is a more standard approach in the Python ecosystem.

## Checklist

### DIFF
--- a/rsconnect/bundle.py
+++ b/rsconnect/bundle.py
@@ -201,9 +201,7 @@ class Manifest:
                 # If the environment has a python version requirement,
                 # add it to the manifest as environment.python.requires
                 manifest_environment = self.data.setdefault("environment", {})
-                manifest_environment["python"] = {
-                    "requires": environment.python_version_requirement
-                }
+                manifest_environment["python"] = {"requires": environment.python_version_requirement}
 
         if image or env_management_py is not None or env_management_r is not None:
             manifest_environment = self.data.setdefault("environment", {})
@@ -677,9 +675,10 @@ def make_html_manifest(
     filename: str,
 ) -> ManifestData:
     # noinspection SpellCheckingInspection
+    appmode = "static"
     manifest: Manifest(
         metadata=ManifestDataMetadata(
-            appmode="static",
+            appmode=appmode,
             primary_html=filename,
         )
     )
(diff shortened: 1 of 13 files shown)


rsconnect/bundle.py
```diff
@@ -378,56 +388,16 @@ def make_source_manifest(
...
-            manifest["environment"]["environment_management"] = {}
-            if env_management_py is not None:
-                manifest["environment"]["environment_management"]["python"] = env_management_py
-            if env_management_r is not None:
-                manifest["environment"]["environment_management"]["r"] = env_management_r
-
-    manifest["files"] = {}
-
-    return manifest
+    manifest: Manifest = Manifest(
```
amol-, line 391: All places where we were manually creating `ManifestData` have been migrated to use the `Manifest` class. This ensures that the same behaviour is applied independently from the command used, and thus that the python version requirements are always applied to the manifest without having to set them in every single place.

This is also more robust to future changes to the manifest.

rsconnect/subprocesses/inspect_environment.py
```diff
@@ -0,0 +1,235 @@
+#!/usr/bin/env python
```
amol-, line 1: Didn't change, just renamed from `rsconnect.environment`

rsconnect/pyproject.py
```diff
@@ -16,6 +16,23 @@
     import toml as tomllib  # type: ignore[no-redef]
 
 
+def detect_python_version_requirement(directory: typing.Union[str, pathlib.Path]) -> typing.Optional[str]:
```
amol-, line 19: This is more or less the entry point for the whole version requirement detectiong process, based on the previously merged metadata parsing functions.

amol-:
There are the two failures that are also happening on `main` and independent from this PR

Lytol (approved):
Outstanding! Really appreciate the refactor of both the `Environment` class as well as extracting the environment inspection into `subprocesses`. I did not review line-by-line on the refactors that were simply moving things around, so I'm trusting that the test coverage is enough there.

My comments are trivial nits and/or copy edits, so do with them what you'd like.

PR #651, opened 2025-03-25
spcs support

edavidaja:
initial support for interacting with Connect in SPCS

## Type of Change

- [ ] Bug Fix           
- [x] New Feature       
- [ ] Breaking Change   

## Approach

Take an optional dependency on the snow CLI
commands requiring interaction with remote servers gain an optional --snowflake-connection-name argument
The ingress URL, account name, and keypair configured out of band are sufficient information
to obtain the OAuth token required to interact with Connect in SPCS.

supports rsconnect add, details, and deploy_html

## Checklist

### DIFF
--- a/rsconnect/api.py
+++ b/rsconnect/api.py
@@ -235,9 +235,33 @@ class RSConnectServer(AbstractRemoteServer):
         self.ca_data = ca_data
         # This is specifically not None.
         self.cookie_jar = CookieJar()
+        # 🤡
+        self.snowflake_connection_name = None
 
 
-TargetableServer = typing.Union[ShinyappsServer, RSConnectServer, CloudServer]
+class SPCSConnectServer(AbstractRemoteServer):
+    """
+    A class encapsulating the information required to interact with Connect in SPCS.
+    """
+
+    def __init__(
+        self,
+        url: str,
+        snowflake_connection_name: Optional[str],
+        insecure: bool = False,
+        ca_data: Optional[str | bytes] = None,
+    ):
+        super().__init__(url, "Posit Connect")
+        self.insecure = insecure
+        self.ca_data = ca_data
+        self.snowflake_connection_name = snowflake_connection_name
+        # for compatibility with RSConnectClient
+        self.cookie_jar = CookieJar()
+        self.api_key = None
+        self.bootstrap_jwt = None
+
+
+TargetableServer = typing.Union[ShinyappsServer, RSConnectServer, CloudServer, SPCSConnectServer]
 
 
 class S3Server(AbstractRemoteServer):
@@ -254,7 +278,7 @@ class RSConnectClientDeployResult(TypedDict):
 
 
 class RSConnectClient(HTTPServer):
-    def __init__(self, server: RSConnectServer, cookies: Optional[CookieJar] = None):
+    def __init__(self, server: RSConnectServer | SPCSConnectServer, cookies: Optional[CookieJar] = None):
         if cookies is None:
             cookies = server.cookie_jar
         super().__init__(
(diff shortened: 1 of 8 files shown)


rsconnect/api.py
```diff
@@ -234,9 +234,33 @@ def __init__(
         self.ca_data = ca_data
         # This is specifically not None.
         self.cookie_jar = CookieJar()
+        # 🤡
+        self.snowflake_connection_name = None
```
mconflitti-pbc, line 238: Curious as to why this is needed here but not in the other non SPCS targetable servers.

rsconnect/snowflake.py
```diff
@@ -0,0 +1,100 @@
...
+            return True
+        except OSError:
+            return False
+
+
+def list_connections():
+
+    if not is_snow_installed():
+        raise RSConnectException(
+            "The snowflake-cli is required bit not installed."
```
aronatkins, line 27:
```suggestion
            "The snowflake-cli is required but not installed."
```

rsconnect/http_support.py
```diff
@@ -256,6 +260,9 @@ def key_authorization(self, key: str):
     def bootstrap_authorization(self, key: str):
         self.authorization("Connect-Bootstrap %s" % key)
 
+    def snowflake_authorization(self, token: str):
+        self.authorization('Snowflake Token="%s"' % token)
```
aronatkins, line 264: The most important quotes IN THE WORLD.

rsconnect/api.py
```diff
@@ -234,9 +234,33 @@ def __init__(
...
 
 
-TargetableServer = typing.Union[ShinyappsServer, RSConnectServer, CloudServer]
+class SPCSServer(AbstractRemoteServer):
```
aronatkins, line 241:
```suggestion
class SPCSConnectServer(AbstractRemoteServer):
```

rsconnect/snowflake.py
```diff
@@ -0,0 +1,100 @@
...
+    def exchange_token(self, spcs_endpoint: str, snowflake_connection_name: Optional[str] = None):
+
+        payload = self.fmt_payload(spcs_endpoint, snowflake_connection_name)
+
+        response = self.request(
+            method="POST",
+            path="/oauth/token",
+            body=payload,
+            headers={"Content-Type": "application/x-www-form-urlencoded"},
+        )
```
mconflitti-pbc, line 95: maybe this is a snowflake quirk, but is it really the case that the request requires the payload to be url encoded but then passed as the body of the post? Can the dictionary not just be passed directly to the body here?

rsconnect/snowflake.py
```diff
@@ -0,0 +1,100 @@
...
+
+def get_token_endpoint(snowflake_connection_name: Optional[str] = None) -> str:
+    cx = get_connection(snowflake_connection_name)
+    if cx is None:
+        raise RSConnectException("No Snowflake connection found")
+
+    return f"https://{cx["account"]}.snowflakecomputing.com/"
+
+
+class SnowflakeExchangeClient(HTTPServer):
```
mconflitti-pbc, line 70: i see this is done elsewhere but naming is confusing to call the base class HTTPServer but it is really a client

rsconnect/api.py
```diff
@@ -554,6 +586,7 @@ def __init__(
         name: Optional[str] = None,
         url: Optional[str] = None,
         api_key: Optional[str] = None,
+        snowflake_connection_name: Optional[str] = None,
         insecure: bool = False,
         cacert: Optional[str] = None,
         ca_data: Optional[str | bytes] = None,
```
mconflitti-pbc, line 592: Not an issue of this PR, but there are a number of things I think worth refactoring in this repo. The addition of the snowflake use case highlights that the code is not as extensible as we probably want it to be. Again, no action to take here, but depending on who owns this repo, it may be time to consider some refactoring.

mconflitti-pbc: For instance, I think a more concrete strategy pattern would be super beneficial given how this codebase is used.

edavidaja: 💯

rsconnect/snowflake.py
edavidaja: This is what I get when I drop the URL encoding:

```
❯ rsconnect details -n spcs -vv
[DEBUG] 2025-03-26T10:34:29-0400 Request: POST /oauth/token
[DEBUG] 2025-03-26T10:34:29-0400 Headers:
[DEBUG] 2025-03-26T10:34:29-0400 --> User-Agent: rsconnect-python/1.25.3.dev6+gc4775e6.d20250324
[DEBUG] 2025-03-26T10:34:29-0400 --> Content-Type: application/json; charset=utf-8
[DEBUG] 2025-03-26T10:34:29-0400 The HTTPSConnection timeout is set to '300' seconds
[DEBUG] 2025-03-26T10:34:29-0400 Response: 415 Unsupported Media Type
```

`requests` [performs the URL encoding automatically](https://requests.readthedocs.io/en/latest/user/quickstart/#more-complicated-post-requests); I think some of this is quirks of our library.

PR #655, opened 2025-04-08
Warning about --python and --override-python-version deprecation

amol-:
## Intent

The intent is moving the user toward best practices of using a `.python-version` file when a specific version of Python has to be used to deploy the project. That relies on a standard way to specific version requirements and allows the requirement to be packaged/committed with the project thus making deploys reproducible.

* The `--override-python-version` option isn't needed anymore when `.python-version` is supported and using it made the deploy unreproducible
* The `--python` option is confusing in itself, as it refers to the python interpreter used to run `rsconnect` and on latest connect versions (>=2025.03) it doesn't influence the interpreter used to run the content anymore. The fact that it previously did was practically a side effect. Deprecating it gives the user a suggestion for a reliable and predictable way to specify the python interpreter.

As the `--python` option does no longer influence the interpreter in use itself, but it's only used to detect the environment requirements, I made #656 to change the name to something that better conveys its purpose in the future.

## Type of Change

- [ ] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change   

## Approach

Add a deprecation message without disabling the existing option.
This will start suggesting users to embrace the best practice and give them time to migrate.
In the future we will be able to drop the options as users migrated away from them.

## Automated Tests

Two tests were added asserting the the deprecation messages are correctly emitted

## Checklist

### DIFF
--- a/rsconnect/environment.py
+++ b/rsconnect/environment.py
@@ -129,9 +129,10 @@ class Environment:
         _warn_on_missing_python_version(python_version_requirement)
 
         if override_python_version:
-            # TODO: --override-python-version should be deprecated in the future
-            #       and instead we should suggest the user sets it in .python-version
-            #       or pyproject.toml
+            logger.warning(
+                "The --override-python-version option is deprecated, "
+                "please use a .python-version file to force a specific interpreter version."
+            )
             python_version_requirement = f"=={override_python_version}"
 
         # with cli_feedback("Inspecting Python environment"):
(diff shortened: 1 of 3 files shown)


rsconnect/environment.py
```diff
@@ -128,10 +128,20 @@ def create_python_environment(
         python_version_requirement = pyproject.detect_python_version_requirement(directory)
         _warn_on_missing_python_version(python_version_requirement)
 
+        if python is not None:
+            # TODO: Remove the option in a future release
+            logger.warning(
+                "On modern connect versions, the --python option won't influence "
```
Lytol, line 134:
```suggestion
                "On modern Posit Connect versions, the --python option won't influence "
```

Lytol (approved):
Wonderful! I noted a small copy edit to keep consistent with how we refer to Connect.

PR #662, opened 2025-04-28
Adapt .python-version requirements to connect accepted ones

### LINKED ISSUE #659: .python-version is not necessarily a version specifier (per PEP 440) and thus cannot necessarily be used for `environment.python.requires`
Per [PEP 440](https://peps.python.org/pep-0440/#version-specifiers), a version specifier must contain a comparison operator (i.e. `>`, `==`, etc). Connect expects a PEP-440 compliant version specifier for the `environment.python.requires` manifest property.

A few options that I see:

1) Ensure that `.python-version` is a valid version specifier; otherwise, skip.
2) Modify the version that we get from `.python-version` to make it a compliant version specifier (i.e. prepend `==`)
3) Do not use `.python-version` at all 🤷 

I'm fine with any of these, but here's my 2 cents:

Given that `.python-version` appears to be pretty bespoke for `uv` and can contain a wide variety of non-version-specifier values (see <https://docs.astral.sh/uv/concepts/python-versions/#requesting-a-version>), then I'd vote to drop the precedence of `.python-version` below `pyproject.toml` and `setup.cfg`, and then (1) skip if not a valid version specifier.

@amol- What do you think?

amol-:
## Intent

Connect requires a version constraint operator to always be specified, so version requirements like "3.8" are not valid, they need to e "~=3.8" to reproduce the intended behaviour.

fixes #659 

## Type of Change

- [x] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change   

## Approach

Versions loaded from `.python-version` are validated against what connect supports and adapted when necessary

## Automated Tests

Added a dedicated test for version adapting

## Checklist

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -24,4 +24,7 @@
         "build/**": true,
         "venv/**": true,
     },
+    "python.analysis.exclude": [
+        "tests"
+    ],
 }
--- a/rsconnect/pyproject.py
+++ b/rsconnect/pyproject.py
@@ -134,9 +134,7 @@ def adapt_python_requires(
         for constraint in constraints:
             constraint = constraint.strip()
             if "@" in constraint or "-" in constraint or "/" in constraint:
-                raise InvalidVersionConstraintError(
-                    f"python specific implementations are not supported: {constraint}"
-                )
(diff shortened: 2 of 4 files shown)


marcosnav (approved):
🚀

Lytol (approved):
:pinched_fingers: Wonderful

PR #664, opened 2025-05-02
add comprehensive CLI documentation

edavidaja:
## Intent

Add more comprehensive documentation of the CLI

## Type of Change

- [ ] Bug Fix           
- [x] New Feature       
- [ ] Breaking Change   

## Approach

- move the README that currently provides most documentation to `docs/`
  - The readme is currently disjointed and pretty long; it would probably make sense to break it into more comprehensible pieces as part of this PR or separately
  - replace the existing readme with a stripped down version and links to the new docs
- Use `mkdocs-click` to generate pages for each `rsconnect` subcommand.
  - `quartodoc` does not currently provide support for documenting command-line interfaces.
  - we could render all the commands and subcommands as one page, but that seems less helpful than breaking things up by command
    - subcommand pages are rendered without the `rsconnect` command; [this is a limitation of mkdocs-click]](https://github.com/mkdocs/mkdocs-click/issues/54)

## Automated Tests

Github pages preview to be added.

## Directions for Reviewers

```
uv venv
uv pip install -e ".[docs]"
. ./.venv/bin/activate
mkdocs serve
```

## Checklist


.github/workflows/preview-docs.yml
```diff
@@ -0,0 +1,34 @@
+name: preview docs
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - closed
```
tdstein, line 9: What's the target use case for running on closed? You might be able to remove this.

docs/index.md
```diff
@@ -0,0 +1,208 @@
...
+```
+
+You may also build and install a wheel directly from a repository clone:
+
+```bash
+git clone https://github.com/posit-dev/rsconnect-python.git
+cd rsconnect-python
+pip install pipenv
+make dist
+pip install ./dist/rsconnect_python-*.whl
```
tdstein, line 29: If this is aimed at installing pre-release versions, then you could use...

```suggestion
pip install git+https://github.com/posit-dev/rsconnect-python.git
```

.github/workflows/preview-docs.yml
edavidaja: Based on the action docs I think the close event is used to delete the preview--the build is skipped on line 26 if the event is closed.

PR #667, opened 2025-05-16
publish to PyPI as `rsconnect` and `rsconnect-python`

edavidaja:
- update makefile to support building wheels with a different package name
- add a script to rewrite the package name in CI
  - `uv run` is used for this so that the necessary `toml` dependency
    is installed without drama

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -68,6 +68,9 @@ jobs:
 
   distributions:
     needs: test
+    strategy:
+      matrix:
+        package_name: ["rsconnect_python", "rsconnect"]
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
@@ -78,14 +81,19 @@ jobs:
     - uses: actions/setup-python@v4
       with:
         python-version: 3.8.x
+    - name: Install uv # see scripts/temporary-rename
+      uses: astral-sh/setup-uv@v6
     - run: pip install -e '.[test]'
     - run: pip freeze
     - run: make dist
       id: create_dist
+      env:
+        PACKAGE_NAME: ${{ matrix.package_name }}
     - uses: actions/upload-artifact@v4
       with:
         name: distributions
         path: dist/
+      if: matrix.package_name == 'rsconnect_python'
     - run: pip install -vvv ${{ steps.create_dist.outputs.whl }}
     - run: rsconnect version
     - run: rsconnect --help
@@ -101,16 +109,19 @@ jobs:
       with:
         role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
         aws-region: ${{ secrets.AWS_REGION }}
-    - if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    - if: github.event_name == 'push' && github.ref == 'refs/heads/main' && matrix.package_name == 'rsconnect_python'
       run: make sync-latest-to-s3
-    - if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+      env:
+        BDIST_WHEEL: ${{ steps.create_dist.outputs.whl }}
+    - if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags') && matrix.package_name == 'rsconnect_python'
       run: make sync-to-s3
+      env:
+        BDIST_WHEEL: ${{ steps.create_dist.outputs.whl }}
     - if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
         user: __token__
         password: ${{ secrets.PYPI_TOKEN }}
-
   docs:
     needs: test
     runs-on: ubuntu-latest
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,8 @@ VERSION := $(shell python -m setuptools_scm)
 HOSTNAME := $(shell hostname)
 S3_PREFIX := s3://rstudio-connect-downloads/connect/rsconnect-python
 
-BDIST_WHEEL := dist/rsconnect_python-$(VERSION)-py2.py3-none-any.whl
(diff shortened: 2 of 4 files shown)


tdstein:
It might be worth reverting from `pyproject.toml` to setup.py to support this: https://github.com/posit-dev/rsconnect-python/pull/668

.github/workflows/main.yml
```diff
@@ -111,6 +111,39 @@ jobs:
...
+        name: distributions
+        path: dist/
+    - run: pip install -vvv ${{ steps.create_dist.outputs.whl }}
+    - run: rsconnect version
+    - run: rsconnect --help
+    - if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        user: __token__
+        password: ${{ secrets.PYPI_TOKEN }}
```
tdstein, line 145: Should this replace the distributions step above?

.github/workflows/main.yml
```diff
@@ -111,6 +111,39 @@ jobs:
...
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - uses: actions/setup-python@v4
+      with:
+        python-version: 3.8.x
+    - name: Install uv # see scripts/temporary-rename
+      uses: astral-sh/setup-uv@v5
```
tdstein, line 127: Set up Python using `uv`: https://github.com/posit-dev/posit-sdk-py/blob/main/.github/workflows/ci.yaml#L32-L33

Should be faster.

.github/workflows/main.yml
```diff
@@ -111,6 +111,39 @@ jobs:
...
+    - run: pip install -e '.[test]'
+    - run: pip freeze
+    - run: make dist
+      id: create_dist
+      env:
+        PACKAGE_NAME: rsconnect
```
tdstein, line 133: Oh, I see. You are using a separate job to release the 'rsconnect' named version.

Is it possible to parameterize this and use the same release flow for each?

.github/workflows/main.yml
tdstein: Looks like the python version is builtin now as well: 

```yml
        uses: astral-sh/setup-uv@v6
        with:
          python-version: ${{ matrix.python-version }}

**Comment by edavidaja** [comment]
> It might be worth reverting from `pyproject.toml` to setup.py to support this: #668

[PyPA docs suggest pyproject.toml is the way](https://packaging.python.org/en/latest/flow/#the-configuration-file).

File: .github/workflows/main.yml (hunk 3 above)
Comment (edavidaja) [reply .github/workflows/main.yml:133]: I was worried about concurrent changes to the file resulting in publishing one but not the other but I think I just forgot how matrixed jobs work; I'll update with a matrixed approach.

**Comment by tdstein** [comment]
> > It might be worth reverting from `pyproject.toml` to setup.py to support this: #668
> 
> [PyPA docs suggest pyproject.toml is the way](https://packaging.python.org/en/latest/flow/#the-configuration-file).

Ya, it's just unconventional for the name field in pyproject.toml to be dynamic. I guess publishing the same project under two different names is also unconventional.

Why not stop using `rsconnect-python` and wholly migrate to `rsconnect`?

**Comment by edavidaja** [comment]
> Why not stop using rsconnect-python and wholly migrate to rsconnect?

I think it will make sense to keep both versions on PyPI up to date until we have the opportunity to update the places where the CLI is used.

File: .github/workflows/main.yml (hunk 2 above)
Comment (edavidaja) [reply .github/workflows/main.yml]: I think that changes a few assumptions about how package installation works in the rest of the steps, which I'd prefer to leave in place for now. 

I think we could simplify much of the build infrastructure here out of docker and into `uv` and that's probably worth doing on its own.

File: .github/workflows/main.yml [hunk 4]
```diff
@@ -68,6 +68,9 @@ jobs:
 
   distributions:
     needs: test
+    strategy:
+      matrix:
+        package_name: ["rsconnect-python", "rsconnect"]
```
edavidaja, line 73:
```suggestion
        package_name: ["rsconnect_python", "rsconnect"]
```

.github/workflows/main.yml
```diff
@@ -101,16 +108,15 @@ jobs:
       with:
         role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
         aws-region: ${{ secrets.AWS_REGION }}
-    - if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    - if: github.event_name == 'push' && github.ref == 'refs/heads/main' && matrix.package_name == 'rsconnect-python'
```
edavidaja, line 111:
```suggestion
    - if: github.event_name == 'push' && github.ref == 'refs/heads/main' && matrix.package_name == 'rsconnect_python'
```

.github/workflows/main.yml
```diff
@@ -101,16 +108,15 @@ jobs:
...
       run: make sync-latest-to-s3
-    - if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+    - if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags') && matrix.package_name == 'rsconnect-python'
```
edavidaja, line 113:
```suggestion
    - if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags') && matrix.package_name == 'rsconnect_python'
```

edavidaja:
Merging here appears to be blocked because of the `test-connect` failures and https://github.com/posit-dev/rsconnect-python/issues/649.

edavidaja:
matrixing the distributions job caused the branch protection rules to go out of sync [(see this discussion)](https://github.com/orgs/community/discussions/26698); I've disabled just the branch protection rule requiring a passing of the distributions check, which I will re-add if I can get it to pass here or after it merges to main.

PR #675, opened 2025-05-30
remove publishing to s3

edavidaja:


Makefile
```diff
@@ -135,23 +135,6 @@ dist:
 	@echo "::set-output name=whl::$(BDIST_WHEEL)"
 	@echo "::set-output name=whl_basename::$(notdir $(BDIST_WHEEL))"
 
-.PHONY: dist-install
-dist-install: dist
-	pip install $(BDIST_WHEEL)
```
tdstein, line 140: fwiw; I would keep this and rename it to:

```makefile
install:
    pip install $(BDIST_WHEEL)
```

My personal preference is to be able to run `make && make install` and end up with the built package installed in my environment.

Issue #676, opened 2025-06-02
[1.26.0] din't work, because the packaged rsconnect.subprocesses are not installed


mdc-webdb:
In the source tar ball, the module exists but it will not be installed.

amol-:
Is this still a valid issue? Seems the associated PR has been dropped

PR #677, opened 2025-06-02
Fix for #676

tuxmaster5000:
Fix for https://github.com/posit-dev/rsconnect-python/issues/676

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,7 +80,10 @@ extend_ignore = ["E203", "E231", "E302"]
 per-file-ignores = ["tests/test_metadata.py: E501"]
 
 [tool.setuptools]
-packages = ["rsconnect"]
+packages = [
+    "rsconnect",
+    "rsconnect.subprocesses"
+]
 
 [tool.setuptools_scm]
 write_to = "rsconnect/version.py"


tuxmaster5000:
When I try to build it as an rpm package. It looks like Fedora 41 is using an modern version of setuptools, because on the EOL version 40 it was not happens. As I understand the documentation of setuptools, when using this option, than all modules must listed by hand. Or don't use this option.
https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

amol-:
> I have played with find, but this will require an complete reorganisation of the directory structure. See: https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#finding-simple-packages The source code must be placed in an sub folder. But I don't know what then happens with the tests and the rest.
> 
> > [tool.setuptools.packages.find]
> > where = [ "rsconnect" ]
> > will fails.

```
[tool.setuptools.packages.find]
include = ["rsconnect"]
namespaces = false
```
should be what we want. If `where` is not specified, it defaults to `.` out of which we only want to include the `rsconnect` directory.

amol-:
Ah! when you say "build process" you mean the rpm build process, not the python distribution build process.

Building the wheel works as expected with find:

```
% python -m build
...
adding 'rsconnect/__init__.py'
adding 'rsconnect/actions.py'
adding 'rsconnect/actions_content.py'
...
adding 'rsconnect/subprocesses/__init__.py'
adding 'rsconnect/subprocesses/inspect_environment.py'
...
adding 'rsconnect_python-1.26.1.dev1+g1108457.d20250605.dist-info/METADATA'
adding 'rsconnect_python-1.26.1.dev1+g1108457.d20250605.dist-info/RECORD'
```

`rsconnect` and `rsconnect.subprocess` are both included in the wheel.

I haven't been building RPMs in years, but that sounds like a bug in `pyproject_save_files.py` and it's out of the scope of `rsconnect-python` to address it. 
If `pyproject_save_files.py` behaves differently than the official PyPa packages like `build` I think it probably needs to be aligned.

> I have try it, but this will break the build process. But I don't know why:
> 
> ```
> + /usr/bin/python3 /usr/lib/rpm/redhat/pyproject_save_files.py --output-files /builddir/build/BUILD/python-rsconnect-1.26.0-build/python-rsconnect-1.26.0-2.fc41.x86_64-pyproject-files --output-modules /builddir/build/BUILD/python-rsconnect-1.26.0-build/python-rsconnect-1.26.0-2.fc41.x86_64-pyproject-modules --buildroot /builddir/build/BUILD/python-rsconnect-1.26.0-build/BUILDROOT --sitelib /usr/lib/python3.13/site-packages --sitearch /usr/lib64/python3.13/site-packages --python-version 3.13 --pyproject-record /builddir/build/BUILD/python-rsconnect-1.26.0-build/python-rsconnect-1.26.0-2.fc41.x86_64-pyproject-record --prefix /usr rsconnect
```

amol-:
Closing as it hasn't seen updates for months, we can reopen if the issue pops up again.

PR #679, opened 2025-06-06
deploy command support for drafts

amol-:
## Intent

Allows to deploy content without activating it.

## Type of Change

- [x] New Feature       

## Approach

New command line option that injects the parameter only when necessary, this way it will be backward compatible with previous API when the parameter is not provided.

## Automated Tests

`test_deploy_draft` runs the command with all possible permutations of content and activate=True/False

## Checklist

### DIFF
--- a/rsconnect/api.py
+++ b/rsconnect/api.py
@@ -419,10 +419,17 @@ class RSConnectClient(HTTPServer):
         env_body = [dict(name=kv[0], value=kv[1]) for kv in env_vars]
         return self.patch("v1/content/%s/environment" % app_guid, body=env_body)
 
-    def app_deploy(self, app_id: str, bundle_id: Optional[int] = None) -> TaskStatusV0:
+    def app_deploy(self, app_id: str, bundle_id: Optional[int] = None, activate: bool = True) -> TaskStatusV0:
+        body = {"bundle": bundle_id}
+        if not activate:
+            # The default behavior is to activate the app after deployment.
+            # So we only pass the parameter if we want to deactivate it.
+            # That way we can keep the API backwards compatible.
+            body["activate"] = False
+
         response = cast(
             Union[TaskStatusV0, HTTPResponse],
-            self.post("applications/%s/deploy" % app_id, body={"bundle": bundle_id}),
+            self.post("applications/%s/deploy" % app_id, body=body),
         )
         response = self._server.handle_bad_response(response)
         return response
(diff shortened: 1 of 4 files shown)


rsconnect/api.py
```diff
@@ -544,10 +567,10 @@ def deploy(
 
         app_bundle = self.app_upload(app_id, tarball)
 
-        task = self.app_deploy(app_id, app_bundle["id"])
+        task = self.content_deploy(app_guid, app_bundle["id"], activate=activate)
```
marcosnav, line 570: This was the only use of `app_deploy`, it might be worth clearing it up

marcosnav (approved):
Implementation wise, looks good to me, a couple non blocker items:
- It'd be nice to clean up the `app_deploy`
- I'm starting to question the use of the term "draft"

edavidaja:
I thought we were calling this `preview`? or did we change the name upstream as well?

amol-:
> I thought we were calling this `preview`? or did we change the name upstream as well?

The name in Connect is `"active"`, the word `"preview"` is primarily used as a verb to indicate the action of looking at a bundle that is not active, but not to set it as "active" or "not active".

 during design there were a few discussions about glossary and one of the terms for the idea of deploying a new application with 0 active bundles was to deploy a draft. As a `--not-active` or `--inactive` option seemed terrible as a name, I chose to go for the "draft" name which seems to better convey the intent of the action, is short and is positive instead of negative.

edavidaja:
I think "preview deployment" is well-enough established as a noun in similar contexts (e.g. [netlify](https://docs.netlify.com/site-deploys/deploy-previews/), [cloudflare](https://developers.cloudflare.com/pages/configuration/preview-deployments/), [vercel](https://vercel.com/docs/deployments/sharing-deployments), etc)  that `deploy --preview` communicates the intent pretty clearly.

amol-:
> I think "preview deployment" is well-enough established as a noun in similar contexts (e.g. [netlify](https://docs.netlify.com/site-deploys/deploy-previews/), [cloudflare](https://developers.cloudflare.com/pages/configuration/preview-deployments/), [vercel](https://vercel.com/docs/deployments/sharing-deployments), etc) that `deploy --preview` communicates the intent pretty clearly.

Agreed, but "preview" was moved to "activate" as a term in the API.
So if we want to rename it I'd prefer to go for a name that at least has adherence to the same glossary used by the APIs

jonkeane:
Let's go with what's here in the code now (which IIRC, is `draft`) ultimately what we _think_ it should be called and debating it here isn't helpful — what matters is what will land with our users. Let's keep this in mind when we do user testing and see if folks find `... --draft` awkward and we can revisit then.

PR #680, opened 2025-06-09
Require requirements

### LINKED ISSUE #538: Require requirements.txt file
Currently the package generates a requirements.txt file automatically as part of the deployment process which causes a lot of problems for neophite users. A few observations from supporting lots of python users through this process:

1) Virtual environments are essential to python deployments
When people try to deploy to either Connect or Shinyapps without a virtual environment they have a horrible time, and get really bewildering failures after long deploy jobs. I have never seen a real user have a good time with our products without using a virtual environment.

2) People who use virtual environments likely also use requirements.txt files
As a result it's not a burden for these users to be required to either do their own `pip freeze` or maintain their own requirments file, and generating it automatically doesn't help them 

3) People who don't know about requirements.txt files are likely not using a virtual environment. These people seem to benefit from the automatic file generation, but it's actually just making their error worse and more obscure. In particular:
- Their system python is likely a mess so `pip freeze` will generate an unsolvable environment and a deployment failure. The error they get back doesn't help them, and it only occurs after Connect or Shinyapps have spent a long time trying to install things
- Locking things down to particular package versions is too restrictive and also creates errors
- OS-specific packaages like [pywin](https://pypi.org/project/pywin32/) are not installable on Linux servers, but are generated by `pip freeze`

As a result I think we should cancel the autogeneration and require the requirements.txt file. This is what [Streamlit](https://docs.streamlit.io/streamlit-community-cloud/deploy-your-app) and [HuggingFace](https://huggingface.co/new-space?template=posit%2Fshiny-for-python-template) require.

jonkeane:
## Intent
Require a requirements file by default, do not provide one. This is partially an idea to see how others feel about it. I would argue that we actually should remove the pip freeze fallback entirely, honestly. Possibly replacing it with https://github.com/bndr/pipreqs (or we could pull in the go code that the publisher uses).

Resolves #538

## Type of Change
- [ ] Bug Fix            
- [ ] New Feature       
- [x] Breaking Change   

## Approach
Mostly Claude, actually

## Automated Tests
Changed

## Checklist
Will do at some point:

### DIFF
--- a/rsconnect/environment.py
+++ b/rsconnect/environment.py
@@ -340,7 +340,8 @@ def _check_requirements_file(directory: str, requirements_file: typing.Optional[
         raise RSConnectException(
             "The requirements file '%s' does not exist in '%s'.\n"
             "Please create the file or specify a different file with --requirements-file.\n"
-            "To have the requirements file generated using pip freeze, pass --force-generate." % (requirements_file, directory)
+            "To have the requirements file generated using pip freeze, pass --force-generate."
+            % (requirements_file, directory)
         )
 
 
--- a/rsconnect/subprocesses/inspect_environment.py
+++ b/rsconnect/subprocesses/inspect_environment.py
@@ -74,6 +74,7 @@ def detect_environment(dirname: str, requirements_file: Optional[str] = "require
     """
 
     if requirements_file is None:
+        # --force-generate sets requirements_file to None
         result = pip_freeze()
     elif os.path.basename(requirements_file) == "uv.lock":
         result = uv_export(dirname, requirements_file)


rsconnect/subprocesses/inspect_environment.py
```diff
@@ -207,16 +219,36 @@ def main():
...
+        require_requirements_txt = True
+
+        # Check for flags in first argument
+        if len(sys.argv) > 2 and sys.argv[1].startswith("-") and not sys.argv[1].startswith("--"):
             flags = sys.argv[1]
-        if "f" in flags:
-            force_generate = True
-        envinfo = detect_environment(directory, force_generate)._asdict()
+            if "f" in flags:
+                force_generate = True
```
amol-: Should we just switch to `argparse`? 😅

rsconnect/subprocesses/inspect_environment.py
```diff
@@ -60,21 +60,33 @@ class EnvironmentException(Exception):
...
 
     if force_generate:
         result = pip_freeze()
     else:
-        result = output_file(dirname, "requirements.txt", "pip") or pip_freeze()
+        # Try to read requirements.txt file
+        try:
+            result = output_file(dirname, "requirements.txt", "pip")
+        except EnvironmentException as e:
+            # For backwards compatibility in tests
```
amol-: This comment is a bit misleading, it's not just for tests, the `require_requirements_txt` is excplicitly true, so the behavior was requested

rsconnect/subprocesses/inspect_environment.py
```diff
@@ -121,13 +133,13 @@ def output_file(dirname: str, filename: str, package_manager: str):
...
-    if successful, None if the file does not exist,
-    or a dictionary containing 'error' on failure.
+    if successful, or raises an EnvironmentException if the file does not exist
+    or on other failures.
     """
     try:
         path = os.path.join(dirname, filename)
         if not os.path.exists(path):
-            return None
+            raise EnvironmentException(f"{filename} file is required for deployment")
```
amol-: I think this changes the semantic of the method and ends up mixing two different cases. In one case the environment is broken (`Error reading...`) but in the other case the file is missing, which is a perfectly valid outcome.

I suggest we retain `None` in case of missing file and change `detect_environment` to handle separatedly "requirements don't exist" and "requirements are broken". As not exist should trigger a regenerate, but exist and broken should not, it should error.

rsconnect/environment.py
```diff
@@ -101,6 +101,7 @@ def create_python_environment(
         python: typing.Optional[str] = None,
         override_python_version: typing.Optional[str] = None,
         app_file: typing.Optional[str] = None,
+        require_requirements_txt: bool = True,
```
amol-: I suggest we collapse `force_generate` and `require_requirements_file`, it's practically a single option with three possible cases:

- requirements : regenerate
- requirements : require
- requirements : auto

It would be easier to reason about it across the code instead of having to deal with possible interactions of `force_generate`+`require_requirements`

amol-:
@jonkeane mind reviewing the changes when you have a chance?

rsconnect/subprocesses/inspect_environment.py
```diff
@@ -78,7 +78,12 @@ def detect_environment(dirname: str, requirements_file: Optional[str] = "require
     elif os.path.basename(requirements_file) == "uv.lock":
         result = uv_export(dirname, requirements_file)
     else:
-        result = output_file(dirname, requirements_file, "pip") or pip_freeze()
+        result = output_file(dirname, requirements_file, "pip")
+        if result is None:
+            raise EnvironmentException(
+                "The requirements file '%s' was not found in '%s'. "
+                "Please create it or use --force-generate to use pip freeze." % (requirements_file, dirname)
+            )
```
jonkeane, line 87: This might be due to being away from this block for so long, but it took me a second to realize/remember that `requirements_file` here is just the name of the file, so that first `if` will only be false when it's literally passed as `None` (i.e. when someone passes `--force-generate`), and _not_ when that file doesn't exist (that's lower down in the `else`). 

I think this is fine, and maybe this comment is enough details in the git record, but it did take me a second to figure out how the code change here was even accomplishing what we wanted it to!

rsconnect/environment.py
```diff
@@ -336,10 +337,10 @@ def _warn_if_no_requirements_file(directory: str, requirements_file: typing.Opti
...
 
     if not requirements_file_path.exists():
-        click.secho(
-            "    Warning: Capturing the environment using 'pip freeze'.\n"
-            "             Consider creating a %s file instead." % requirements_file,
-            fg="yellow",
+        raise RSConnectException(
+            "The requirements file '%s' does not exist in '%s'.\n"
+            "Please create the file or specify a different file with --requirements-file.\n"
+            "To use pip freeze instead, pass --force-generate." % (requirements_file, directory)
```
tdstein:
```suggestion
            "To have the requirements file generated using pip freeze, pass --force-generate." % (requirements_file, directory)
```

rsconnect/subprocesses/inspect_environment.py
amol-: The logic is more or less:

```
- requirements_file was requested -> use it and error if it doesn't exist
- requirements_file was not requested -> generate a new one (ie `--force-generate`)
```

Happy to improve the comment in any way you feel it might make this more obvious.
Maybe I can just add an explicit comment within the `if` block stating that `--force-generate` causes `requirements_file` to be `None`

jonkeane: > block stating that --force-generate causes requirements_file to be None

Yeah, that would have helped me when I was reading

amol-:
> Is there a simple way to make this use `uv pip freeze` if it's available?

I guess so, is there any specific improvement you think that would provide? 

Environments without pip are very uncommon as in most cases `venv` performs `ensurepip` on standard python builds so the case where you have `uv` and not `pip` sounds very uncommon.

Also the inspect process explicitly runs in a target python environment, so there isn't the need for `uv` environment autodetection for example.

amol-:
> We don't have to here, but I _do_ think we should still consider the following separately
> 
> > Possibly replacing it with https://github.com/bndr/pipreqs (or we could pull in the go code that the publisher uses).

`pipreqs` is more or less in minimal maintenance mode and looking for a maintainer. Not sure how much we could rely on it on long term. Also `pipreqs` does not support `qmd` files which we currently support and for .ipynb it depends on jupyter nbconvert to be installed.
And I wasn't fond of its approach of keeping a manual list of standard library modules and a manual mapping of package_name->distribution_name, that's why I had started https://github.com/posit-dev/picopip/pull/3 but had to pause it due to other priorities. I'll get back to that experiment as soon as I pop other things from my stack

PR #692, opened 2025-07-21
ci: update requirements.txt for snyk

nealrichardson:
## Intent

The requirements.txt claims only exists to facilitate dependency scanning. We're getting spurious alerts from it. Fix that.

## Approach

~Add a reminder comment in pyproject.toml. Copy over to requirements.txt and delete the optional/dev dependencies listed there. Alphabetize.~

Update the `snyk.yml` workflow to `uv pip compile` from `pyproject.toml`, so snyk stays in sync. Then delete the `requirements.txt` because it's only used for snyk, right? Wrong, other things are failing. So we'll need to clean up other references to that requirements file.

## Automated Tests

CI should pass.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,32 @@
+# TODO(#649): this file shouldn't exist, but there currently are some
+# integration test suites that rely on it. The integration tests deserve to be
+# overhauled along the lines of the posit-sdk-py tests, so we can clean this up
+# when we do that.
+
+# build-system.requires
+setuptools>=61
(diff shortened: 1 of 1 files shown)


edavidaja (changes requested):
I think we can `uv pip compile` to generate the requirements file.

edavidaja (approved):

nealrichardson:
Tests are still failing but the only non-comment change in this PR is in snyk.yml so I don't see how they're related.

PR #697, opened 2025-08-01
Update system caches tests

amol-:
## Intent

connect path to system caches has moved inside `python-environments/_packages_cache`, the test suite is still looking in the old path.

## Type of Change

- [x] Bug Fix

### DIFF
--- a/tests/test_main_system_caches.py
+++ b/tests/test_main_system_caches.py
@@ -10,10 +10,10 @@ from rsconnect.main import cli
 CONNECT_SERVER = "http://localhost:3939"
 CONNECT_KEYS_JSON = "vetiver-testing/rsconnect_api_keys.json"
 
-ADD_CACHE_COMMAND = "docker compose exec -u rstudio-connect -T rsconnect mkdir -p /data/python-environments/pip/1.2.3"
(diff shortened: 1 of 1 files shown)


tests/test_main_system_caches.py
```diff
@@ -131,19 +137,6 @@ def test_system_caches_delete_admin(self):
 
         # TODO: Unsure how to test log messages received from Connect.
 
-    # Admins cannot delete caches that do not exist
```
amol-, line 134: This test should not be here imho,
it's testing a connect behaviour not `rsconnect-python`.

We are already testing that deleting a cache does work, and we are already testing that reporting an error message from delete works:

- `test_system_caches_delete_admin`
- `test_system_caches_delete_publisher`

So this test didn't test anything additional and make the testsuite too dependant on a connect behaviours that didn't break compatibility with rsconnect itself.

amol-:
> This is a test-only fix? There's nothing in the code that needs to be adapted to point at the new cache location?

Yes, test only change as tests were failing on main
There shouldn't be any need to change rsconnect itself. When verifying the commands via shell they all worked as expected for me.
In the end the commands only invoke the Connect APIs, so `rsconnect-python` doesn't need to know _where_ those caches are stored.

PR #700, opened 2025-08-12
docs: integration requests

mconflitti-pbc:
## Intent
Docs and example for integration requests in the manifest.json

The plan is to duplicate these in rsconnect and then link out to each tool in Connect docs.. 

I am willing to be persuaded that we instead put these docs in Connect but not sure where.

## Checklist


edavidaja:
> I am willing to be persuaded that we instead put these docs in Connect but not sure where.

- https://docs.posit.co/connect/user/publishing-cli/
- https://docs.posit.co/connect/user/publishing-r/

is where I would put them.

docs/deploying/integration-requests.md
```diff
@@ -0,0 +1,52 @@
...
+- `salesforce`
+- `snowflake`
+- `connect`
+- `aws`
+- `custom`
+
+Possible values for `auth_type` include:
+- `Viewer`
+- `Service Account`
+- `Visitor API Key`
```
edavidaja, line 36: If these are to be provided as values at the command line (which feels a little surprising, spaces and capitalization and such), show an example command line invocation illustrating the necessary quotation, whether an equals sign is required, etc.

edavidaja (changes requested):
I think it would be helpful to provide a little more context about the intended workflow.

docs/deploying/integration-requests.md
joshyam-k: as of right now I don't think there are plans for these types of values to be passed in the command line. these are just possible values that can be listed inside of an "integration_request" in the manifest that can be used to identify the integration that should be associated with the deployed content.

joshyam-k:
closing in favor of https://github.com/posit-dev/connect/pull/33121

PR #711, opened 2025-10-14
feat: simple MCP server

joshyam-k:
## Type of Change

- [ ] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change   

## Checklist


pyproject.toml
```diff
@@ -5,15 +5,15 @@ description = "The Posit Connect command-line interface."
 authors = [{ name = "Posit, PBC", email = "rsconnect@posit.co" }]
 license = { file = "LICENSE.md" }
 readme = { file = "README.md", content-type = "text/markdown" }
-requires-python = ">=3.8"
+requires-python = ">=3.10"
```
joshyam-k: `fastmcp` requires >=3.10 which would be a breaking change. What's the protocol around doing something like this? Is this reason enough to move this work off of `rsconnect-python`?

PR #719, opened 2025-10-30
Send Connect test instance logs to STDOUT for CI jobs to show the log in case of job failure

marcosnav:
The CI step "Get logs in case of failure" was not showing enough info while debugging CI failures.

Setting the logging output to STDOUT, instead of a file, should help to see all the info on GH action runs without the need to replicate locally.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ??
 
+### Added
+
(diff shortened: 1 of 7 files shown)


vetiver-testing/setup-rsconnect/rstudio-connect.gcfg
```diff
@@ -13,6 +13,7 @@ DefaultUserRole = publisher
 
 [Python]
 Enabled = true
+Executable = /opt/python/9.9.9/bin/python
```
marcosnav: Intentionally triggering an error to confirm CI shows full Connect logs. To be removed.

PR #726, opened 2025-11-12
fix: `rsconnect list` bug

### LINKED ISSUE #705: rsconnect-python crashes on rsconnect list when server has null name
**Issue**
The rsconnect list command crashes with TypeError: '<' not supported between instances of 'str' and 'NoneType' when a server entry has a null/None value for its name field.
Re Slack thread: https://positpbc.slack.com/archives/CCD5REGR4/p1756228912248909

**Reproduction Steps**
 1. Create or modify the servers.json file with a null name:
  echo '{"bad_server": {"url": "https://example.com", "name": null}}' > ~/.rsconnect-python/servers.json
  2. Run rsconnect list
  3. Observe the crash with TypeError

Image

#### FROM Claude:
The _get_sorted_values() method in metadata.py uses Python's sorted() function without handling the case where some values might be None. When
  sorting server entries by name, any entry with a null name will cause the sort comparison to fail.

#### Suggested Fix

  Apply one of these solutions:

  1. Handle None values in the sort key:
   In get_all_servers()
  return self._get_sorted_values(lambda s: s["name"] or "")

  2. Add defensive handling in _get_sorted_values:
  def _get_sorted_values(self, sort_by):
      def safe_sort(item):
          key = sort_by(item)
          return key if key is not None else ""
      return sorted(self._data.values(), key=safe_sort)

  #### Additional Notes

  - This likely happens when server entries are manually edited or created with a null name
  - The set() method should validate that the name is not None before storing
  - Consider adding validation elsewhere to prevent null values in required fields

joshyam-k:
## Intent

Resolves #705 

## Type of Change

- [x] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change   

## Approach
- Prevent sorting function called by `rsconnect list` from being fed a null name. This will provide a fix for any users with currently "corrupted" `servers.json` files.
- ~~Make `--name` a required param for `rsconnect add`. To me there is no use case for `rsconnect add`-ing a server with no name as you wouldn't be able to reference it in other commands (like deploy).~~

## Validation

1. add a server with no name:

```
rsconnect add -s http://localhost:3939 -k <api-key>
```

2. Run `rsconnect list`

```
Nickname: "None"
    URL: http://localhost:3939
    API key is saved
```

## Checklist

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ??
 
+### Fixed
+
+- `rsconnect list` now properly functions when a stored server has no nickname.
+
 ## [1.28.0] - 2025-11-06
 
 ### Added


rsconnect/main.py
```diff
@@ -666,6 +666,9 @@ def add(
     set_verbosity(verbose)
     output_params(ctx, locals().items())
 
+    if not name:
```
joshyam-k: We can't directly make `--name` required in `@server_args` because that's used elsewhere. It would add some annoying redundancy to create a second version of it with `--name` set to required, so this seemed like the simplest solution

ChaitaC (commented):
I verified that the PR resolves the initial issue, and all the changes look good to me. 

However, since we’re making the name argument required, I’ll leave the final approval to someone more familiar with this part of the code. I can’t think of any scenarios where this change would cause issues.

Do we have any older documentation that explicitly mentioned `name` as optional? I did a quick check but didn’t find anything.

joshyam-k:
tests against latest version of Connect: https://github.com/posit-dev/connect/actions/runs/19311989599/job/55234313069

toph-allen (approved):
Do you think it's worth adding a server without a name to the [tests for list](https://github.com/posit-dev/rsconnect-python/blob/bf64351ac8c0884274bcbcb1c123c456bc44c25a/tests/test_metadata.py#L100-L110)? You would just need to add one without a nickname around line 25, it would be pretty easy.

I manually verified this locally, but it's better to also have a test probably.

I'll approve this now, as functionally it works — better if you add a short test that `list` works with a no-name server before you merge.

tdstein:
> `rsconnect deploy shiny -n None .` deploys to whichever server named None exists first in the `servers.json` file. Not only is deploying to "None" a really weird thing to do, but it's this behavior that made me want to make `name` required

Ok, just wanted to ensure that it didn't blow up if multiple entries have a name of `"None"`.

tdstein (approved):

Issue #649, opened 2025-03-21
CI failures due to workflow changes on Connect


tdstein:
These failures stem from this change in Connect: https://github.com/posit-dev/connect/commit/88ce4ff4d46ae1c4ba831e99b0136d86c0718cf3#diff-db146a204844f1dd26bd7bcafabfbd64d399001dd8bad4e7174611c628a8bfbe

It removed the Justfile commands that this CI uses (`just test-rsconnect-python-repo`).

The original intention was to ensure that changes to rsconnect-python would not break Connect's CI, which does the inverse operation. It pulls reconnect-python and uses it for its own test matrix. I think a better approach would be what we did with posit-sdk-py and use Docker to run against the preview version of Connect.

nealrichardson:
I did some archaeology (git blame) to try to understand what's happening in these integration test suites. Sources are #331, #384, and #526. `test-rsconnect` does appear to use the public preview docker image for testing something with vetiver and then another thing added later. `test-connect` has its own docker/compose setup and is the one that builds Connect from the checkout and runs the `rsconnect-python` tests contained in it. 

It would be interesting to try dropping the `test-connect` setup and run the test suite it runs using the `test-rsconnect` setup. That would be a step in the right direction perhaps, and would solve this issue literally (it would make all integration tests use the preview build).

PR #727, opened 2025-11-13
ci: remove test-connect jobs 

joshyam-k:
## Intent

Remove `test-connect` jobs from here and instead run tests against dev Connect by specifying the `rsconnect-python` branch in the workflow dispatch for `rsconnect-python-tests-at-night` job in Connect. (https://github.com/posit-dev/connect/actions/workflows/rsconnect-python.yml)

My remaining concern around this revolves around the tests against Connect being an important part of the testing for `rsconnect-python` and while it works fine to trigger them on the Connect side it is a bit of a clunky workflow and nothing is preventing someone from accidentally merging a PR without first running those.

~~TODO: should `test-rsconnect` in `main.yml` be removed as well?~~ (no)

Related to #649 

## Approach

`deploy_tests.yml` can be fully removed as it only included `test-connect` jobs.

## Checklist

- [ ] remove `secrets.CONNECT_PAT` post merge


.github/workflows/main.yml
```diff
@@ -89,11 +89,6 @@ jobs:
       id: create_dist
       env:
         PACKAGE_NAME: ${{ matrix.package_name }}
-    - uses: actions/upload-artifact@v4
```
joshyam-k: this only existed so that `test-connect` could download the artifacts.

nealrichardson:
> TODO: should `test-rsconnect` in `main.yml` be removed as well?

I think those are different (the second and I think third bullets on https://github.com/posit-dev/rsconnect-python/issues/649#issuecomment-3504025244). I don't think we should remove them (yet) but they also don't have the same problems that these do. They're using a nightly docker image of Connect (https://github.com/posit-dev/rsconnect-python/blob/main/docker-compose.yml#L4) but don't have the same source dependency that there is here.

.github/workflows/main.yml
```diff
@@ -172,83 +167,3 @@ jobs:
...
-              QUARTO_VERSION: "1.4.546"
-
-              # This allows us to start Connect separately in our own docker container
-              CONNECT_SERVER: "http://localhost:3939"
-              remote: "yes"
-            run: |
-              cd integration-testing
-              docker compose pull connect
-              docker compose up -d connect
-              just ../test/connect-rsconnect-python/test/rsconnect-python/test-rsconnect-python-repo
```
nealrichardson: Once this merges, we can tidy up this justfile in connect and at least remove this target, which nothing else calls. Probably worth doing an audit there of any other stuff, like in the `docker compose` setup, to see if there's more that we can prune.

.github/workflows/main.yml
```diff
@@ -172,83 +167,3 @@ jobs:
...
-          - uses: actions/checkout@v4
-            with:
-              repository: rstudio/connect
-              path: 'test/connect-rsconnect-python'
-              sparse-checkout: |
-                  test/rsconnect-python
-                  tools/dev
-                  examples
-              sparse-checkout-cone-mode: false
-              token: ${{ secrets.CONNECT_PAT }}
```
nealrichardson: Are you removing all uses of this secret? If so, let's delete it after this merges.

.github/workflows/main.yml
joshyam-k: Connect ticket for this follow up work: https://github.com/posit-dev/connect/issues/34626

PR #736, opened 2025-12-04
feat: add metadata to bundle uploads

nealrichardson:
## Intent

Allow adding metadata to bundle uploads, a (potentially) new API in Connect.

If you're in a git checkout, it will attempt to set the `source`, `source_repo`, `source_branch`, and `source_commit` fields that are already known to Connect. It is also possible to provide additional key-value pairs, or override the values of those fields, using the `--metadata` CLI flag. Or suppress metadata detection with `--no-metadata`. 

If you're on an older version of Connect that does not support this API, or specify the `--no-metadata` flag, the old bundle upload API will be used, and nothing is changed.

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach

I've lost the exact prompt now, but I basically gave Claude the API docs for the new API, described the behavior I wanted (see "Intent"), and let it go. It seemed to get it in one shot, except that rsconnect-python was not using the v1 bundle APIs yet. So I stashed that, did #731, and then reapplied the changes. 

## Automated Tests

There are mocked tests for what looks like all of the new behavior, but we don't have general integration tests in this repo that would test this behavior. The extended nightly integration tests in the connect repo should exercise the new APIs, if not confirm that they're sending the correct information.

I manually against a local Connect server with the new API and confirmed both that I could add extra fields with `--metadata` and that git commit info is picked up from the environment.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.orig
 *.pyc
 .DS_Store
 .coverage


rsconnect/api.py
```diff
@@ -380,6 +389,26 @@ class RSConnectClientDeployResult(TypedDict):
...
+    Git metadata support was added in Connect 2025.11.0.
+
+    :param server_version: The Connect server version string
+    :return: True if the server supports git metadata, False otherwise
+    """
+    if not server_version:
+        return False
+
+    try:
+        return compare_semvers(server_version, "2025.11.0") >= 0
```
nealrichardson:
```suggestion
        return compare_semvers(server_version, "2025.11.0") > 0
```

PR #737, opened 2025-12-04
fix: change User-Agent string product value to PascalCase

tdstein:
## Intent

Some WAFs strictly expect User-Agent strings that use PascalCase, causing the previous User-Agent string (rsconnect-python) to be rejected.

## Type of Change

- [x] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change   

## Checklist

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `rsconnect list` now properly functions when a stored server has no nickname.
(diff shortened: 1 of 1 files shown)


aronatkins (approved):
Do any of our other clients have a similar mistake? Should we use this opportunity to use "Posit"?

karawoo:
> Do any of our other clients have a similar mistake? 

I see rsconnect uses `rsconnect/x.y.z`: https://github.com/rstudio/rsconnect/blob/b234450f2995cffe89bb37a91203fe8a2a21ba4a/R/http.R#L644. Happy to update it.

> Should we use this opportunity to use "Posit"?

I think sticking with the name of the package itself makes more sense, personally

PR #740, opened 2025-12-06
refactor: more v0 API cleanup

nealrichardson:
## Intent

Replaces calls to `__api__/me` with `__api__/v1/user`. Internal renaming of some arguments to say "app" less and "content" more, since they're calling `v1/content` APIs now.

## Approach

Some Claude-assisted refactoring, some hand-crafted.

## Automated Tests

Should pass, happy to hold this until after #739 lands though for more coverage.


marcosnav (approved):
ezpz

PR #755, opened 2026-02-17
Support for quarto shiny app mode

### LINKED ISSUE #754: write-manifest quarto doesn't support Shiny Quarto documents
## Summary

`rsconnect write-manifest quarto` always creates a static document manifest, even for Quarto documents that use `server: shiny` in their YAML front matter. There's no way to generate a `quarto-shiny` manifest via the CLI.

## Details

The codebase has support for `AppModes.SHINY_QUARTO` internally, but the CLI doesn't expose this functionality:

- No `--shiny` flag on `write-manifest quarto`
- No auto-detection of `server: shiny` in the Quarto document's YAML front matter
- `write-manifest shiny` doesn't recognize `.qmd` files as valid entry points

## Current workarounds

1. Use `quarto publish connect` instead (which does handle this correctly)
2. Manually create a `manifest.json` with `"appmode": "quarto-shiny"`

## Expected behavior

Either:
- `write-manifest quarto` should detect `server: shiny` in the document and create a `quarto-shiny` manifest automatically, OR
- A `--shiny` flag should be added to explicitly request a Shiny Quarto manifest

nealrichardson:
## Intent

Closes #754

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach

Ask Claude to write the issue, ask Claude to do it, no not like that Claude, and Claude please review-and-fix your changes until the review is clean

## Automated Tests

Looks like it

## Directions for Reviewers

Other than passing CI, would be good to confirm that this change is enough to make it work in production. @edgararuiz can you check out this locally and see if it does the thing?

## Checklist

### DIFF
--- a/rsconnect/actions.py
+++ b/rsconnect/actions.py
@@ -294,23 +294,25 @@ def is_quarto_shiny(inspect: QuartoInspectResult) -> bool:
 
     # Check formats.<format>.metadata.server.type
     formats = inspect_any.get("formats", {})
-    for format_data in formats.values():
-        if isinstance(format_data, dict):
-            metadata = format_data.get("metadata", {})
-            if isinstance(metadata, dict):
-                server = metadata.get("server", {})
-                if isinstance(server, dict) and server.get("type") == "shiny":
-                    return True
+    if isinstance(formats, dict):
+        for format_data in formats.values():
+            if isinstance(format_data, dict):
+                metadata = format_data.get("metadata", {})
+                if isinstance(metadata, dict):
+                    server = metadata.get("server", {})
+                    if isinstance(server, dict) and server.get("type") == "shiny":
+                        return True
 
     # Check fileInformation.<path>.metadata.server
     file_info = inspect_any.get("fileInformation", {})
-    for file_data in file_info.values():
-        if isinstance(file_data, dict):
-            metadata = file_data.get("metadata", {})
-            if isinstance(metadata, dict):
-                server = metadata.get("server")
-                if server == "shiny":
-                    return True
+    if isinstance(file_info, dict):
+        for file_data in file_info.values():
+            if isinstance(file_data, dict):
+                metadata = file_data.get("metadata", {})
+                if isinstance(metadata, dict):
+                    server = metadata.get("server")
+                    if server == "shiny":
+                        return True
 
     return False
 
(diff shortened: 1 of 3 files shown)


karawoo (approved):
I tested this locally and was able to generate a manifest with `"appmode": "quarto-shiny"`

edgararuiz:
Used to install: `uv pip install git+https://github.com/posit-dev/rsconnect-python@claude-issue-754`
Used to write manifest: `uv tool run rsconnect write-manifest quarto ./dashboard-python/access-to-care.qmd  --overwrite`
The manifest still says: `quarto-static`. I merged the change to my repo here: https://github.com/sol-eng/access-to-care/tree/c8df41d997dc4dc56ec932c5c2517c9586c406de
Resulting manifest: https://github.com/sol-eng/access-to-care/blob/c8df41d997dc4dc56ec932c5c2517c9586c406de/dashboard-python/manifest.json

karawoo:
@edgararuiz are you sure that `uv tool run` is using the correct installed version? What happens if you do `rsconnect write-manifest quarto ./dashboard-python/access-to-care.qmd` directly, or alternatively `uv tool install` to install the dev rsconnect version (instead of `uv pip install`)?

edgararuiz:
@karawoo - Thanks! 
I ran `rsconnect write-manifest quarto ./dashboard-python/access-to-care.qmd --overwrite` and it changed it to `quarto-shiny`

So I tried to deploy to connect again, and it returned this error:

```
The Quarto document requires a server but does not use an executable engine. Consider including some executable code, specifying an engine, or removing the server configuration.
```

By the way, I've seen this error before.  It is same one I would also get when I manually changed `quarto-static` to `quarto-shiny` manually to see if I could override what `rsconnect` was doing.

karawoo:
Ah I see the issue. The rsconnect manifest generation works fine but Connect does not support python shiny quarto documents, only R.

nealrichardson:
I'm going to close this for now since Connect doesn't support this. We can reopen/redo when it does.

nealrichardson:
This could be achieved by calling `quarto render`, and then deploying the resulting app as `python-shiny`: https://quarto.org/docs/dashboards/interactivity/shiny-python/running.html#deployment

PR #742, opened 2025-12-08
ci: Update vetiver tests to use with-connect

nealrichardson:
WIP, trying to sort out whether there are global rsconnect-python things getting picked up that are causing local failure. (#646?)

## Intent

See https://github.com/posit-dev/rsconnect-python/issues/649#issuecomment-3504025244. Looking at the vetiver integration test, it should be able to run fine using `with-connect` and without the custom setup copied in from pins/vetiver.

## Type of Change

- [ ] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change   

## Checklist

- [ ] I have updated [CHANGELOG.md](../CHANGELOG.md) to cover notable changes.
- [ ] I have updated all related GitHub issues to reflect their current state.
- [ ] I have run the `rsconnect-python-tests-at-night` workflow in Connect against this feature branch.

PR #758, opened 2026-02-24
fix: input validation gap in `rsconnect add`

### LINKED ISSUE #757: Missing validation of required inputs in `rsconnect add`
```
% rsconnect add -n x
Checking None...                                 Traceback (most recent call last):
  File "~/.local/share/uv/tools/rsconnect-python/lib/python3.13/site-packages/rsconnect/actions.py", line 76, in cli_feedback
    yield
  File "~/.local/share/uv/tools/rsconnect-python/lib/python3.13/site-packages/rsconnect/main.py", line 540, in _test_server_and_api
    real_server, _ = test_server(api.RSConnectServer(server, api_key, insecure, ca_data))
                     ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "~/.local/share/uv/tools/rsconnect-python/lib/python3.13/site-packages/rsconnect/actions.py", line 151, in test_server
    for test in _to_server_check_list(url):
                ~~~~~~~~~~~~~~~~~~~~~^^^^^
  File "~/.local/share/uv/tools/rsconnect-python/lib/python3.13/site-packages/rsconnect/actions.py", line 124, in _to_server_check_list
    if "//" not in url:
       ^^^^^^^^^^^^^^^
TypeError: argument of type 'NoneType' is not iterable
[ERROR]
Internal error: argument of type 'NoneType' is not iterable
```

joshyam-k:
## Intent
Running `rsconnect add` with insufficient inputs (in this case only specifying the name) should result in a clear `RSConnectException`

fixes: #757

## Type of Change

- [x] Bug Fix            

## Approach

The add command supports three credential pathways, each requiring a different set of options:

1. Posit Connect — requires --server (and typically --api-key)
2. shinyapps.io / Cloud — requires --account, --token, and --secret
3. Snowflake (SPCS) — requires --server with a snowflakecomputing.app host (and optionally --snowflake-connection-name)

Adds an early check at the top of the add command. It raises an `RSConnectException` with a clear message if neither --server nor any of the shinyapps.io options are provided, before any downstream code can encounter None values

Note: click resolves env vars before calling the function, so this won't screw us up in cases where things are being specified as env vars instead of inputs

## Automated Tests
Added a unit test ` test_add_name_only_missing_server_and_credentials`

New behavior

```
uv run rsconnect add -n x
...
rsconnect.exception.RSConnectException: `rsconnect add` requires -s/--server (for Posit Connect) or -A/--account, -T/--token, and -S/--secret (for shinyapps.io).
```

## Checklist

### DIFF
--- a/rsconnect/main.py
+++ b/rsconnect/main.py
@@ -741,6 +741,12 @@ def add(
     set_verbosity(verbose)
     output_params(ctx, locals().items())
 
+    if not server and not any([token, secret, account]):
+        raise RSConnectException(
+            "`rsconnect add` requires -s/--server (for Posit Connect) or -A/--account, -T/--token, "
+            "and -S/--secret (for shinyapps.io)."
+        )
+
     validation.validate_connection_options(
         ctx=ctx,
         url=server,
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -715,6 +715,28 @@ class TestMain:
             if original_server_value:
                 os.environ["CONNECT_SERVER"] = original_server_value
 
+    def test_add_name_only_missing_server_and_credentials(self):
+        """Regression test: `rsconnect add -n x` should produce a validation error, not a TypeError."""
+        original_api_key_value = os.environ.pop("CONNECT_API_KEY", None)
+        original_server_value = os.environ.pop("CONNECT_SERVER", None)
+        try:
+            runner = CliRunner()
+            result = runner.invoke(
+                cli,
+                [
+                    "add",
+                    "--name",
+                    "some-name",
+                ],
+            )
+            assert result.exit_code == 1, result.output
+            assert "`rsconnect add` requires" in str(result.exception)
+        finally:
+            if original_api_key_value:
+                os.environ["CONNECT_API_KEY"] = original_api_key_value
+            if original_server_value:
+                os.environ["CONNECT_SERVER"] = original_server_value
+
     def test_add_shinyapps_missing_options(self):
         original_api_key_value = os.environ.pop("CONNECT_API_KEY", None)
         original_server_value = os.environ.pop("CONNECT_SERVER", None)
--- a/vetiver-testing/vetiver-requirements.txt
+++ b/vetiver-testing/vetiver-requirements.txt
@@ -1,6 +1,6 @@
 pandas
 numpy
-pydantic
+pydantic<2
 pytest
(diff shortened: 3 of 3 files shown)


rsconnect/main.py
```diff
@@ -741,6 +741,12 @@ def add(
     set_verbosity(verbose)
     output_params(ctx, locals().items())
 
+    if not server and not any([token, secret, account]):
```
nealrichardson, line 744: Does this belong inside `validate_connection_options` (below)?

tests/test_main.py
```diff
@@ -715,6 +715,28 @@ def test_add_shinyapps(self):
             if original_server_value:
                 os.environ["CONNECT_SERVER"] = original_server_value
 
+    def test_add_name_only_missing_server_and_credentials(self):
+        """Regression test: `rsconnect add -n x` should produce a validation error, not a TypeError."""
+        original_api_key_value = os.environ.pop("CONNECT_API_KEY", None)
```
nealrichardson, line 720: It makes me sad that you have to do this env management in every test manually.

rsconnect/main.py
joshyam-k: it probably could, but it would require a decent amount of additional complexity because `validate_connection_options` is used across commands. We can't just blindly reject `name` only inputs in it because commands like `rsconnect deploy -n my-server` are perfectly valid.

nealrichardson: Oh, that's sad. Ok.

nealrichardson:
That test failure is weird (and surely unrelated?). I was seeing it in #742 as well (which made slightly more sense), but it makes me wonder if it's really saying that _something_ is wrong, not that literally the model isn't right or whatever.

joshyam-k:
claude is recommending a package pin as a temporary fix. I'm going to try that before diving further into this

update: hmm that passed. maybe we just stumbled upon an issue with vetiver and pydantic>=2?

joshyam-k:
filed: https://github.com/rstudio/vetiver-python/issues/241

rsconnect/main.py
tdstein: Too much sadness in this pull request.

tdstein (approved):

PR #765, opened 2026-04-29
Release 1.29.0

dotNomad:
Prep the CHANGELOG for the `1.29.0` release.


marcosnav (approved):
Thank you!

One super minor comment, I think we tend to keep an empty `## [Unreleased]` section

dotNomad:
> One super minor comment, I think we tend to keep an empty `## [Unreleased]` section

Good catch I'll get that added back 👍

PR #767, opened 2026-05-19
feat: "quickstart" and "deploy pyproject" commands

### LINKED ISSUE #766: Quickstart command to create new projects
Experimental command to create skeleton projects that are ready for deploy to connect.

The goal is to automate and encourage the best practices for python projects and connect deployment by automatically configuring new projects that apply them.

amol-:
## Intent

* Implement `rsconnect deploy pyproject` to deploy via a pyproject file instead of manifest
* Implement `rsconnect quickstart apptype APP_NAME` to create a new content with a skeleton

Closes #766 

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach

- `rsconnect deploy pyproject` detects some connect specific options from `pyproject.toml` and use the standard pyproject options to generate the manifest of the deployed bundle. Allows a reproducible deploy across systems (compared to `rsconnect deploy automode` that reruns the detection of environment etc every time) without the constraint of `rsconnect deploy manifest` which is too specific (like listing the exact files)
- `rsconnect quickstart AppType` will create a new application of the specified app type with a preconfigured `pyproject.toml` and `uv` environment.
- `AppModes` has support for command line aliases, so that `deploy`,`write-manifest` and `quickstart` can share the same alias for the same types.

## Automated Tests

- Tests for the command themselves are there
- Tests for the skeletons exist, but I might go back and simplify them as the testing infrastructure currently goes as far as actually starting those projects and checking they serve requests.

## Directions for Reviewers

- `rsconnect/quickstart/quickstart.py` is the logic for the quickstart command itself, each supported template is implemented by a `TemplateSpec` that points to the template files.
- `quickstart` will also setup a local environment for the quickstart project inside `.venv` of the project when possible, so that the project is immediately runnable.
- `rsconnect/main.py:deploy_pyproject` is the entry point for `rsconnect deploy pyproject`, it primarily uses `read_tool_rsconnect` helper to read the project configuration from `pyproject.toml` itself.
- `uv` was added as a dependency because we already required it anyway for the `--requirement-file=uv.lock` support, as `quickstart` requires it too it made sense to promote it to an actual dependency.

## Checklist


rsconnect/quickstart/templates/api/__connect__.py.tmpl
```diff
@@ -0,0 +1,3 @@
+from .app import create_app
```
jonkeane, line 1: I presume that the stuff under `templates` is all similar boilerplate, so I'm not going to go into too much detail here, but LMK if there's something I should look at in more detail.

rsconnect/quickstart/quickstart.py
```diff
@@ -0,0 +1,472 @@
...
+    :param pathlib.Path cwd: override the working directory (testing hook);
+        defaults to :func:`pathlib.Path.cwd`.
+    """
+    cwd = (cwd or pathlib.Path.cwd()).resolve()
+
+    # Pre-flight checks. Each helper raises ``RSConnectException``
+    # with an actionable message; nothing on disk is mutated until every
+    # check has passed. Type validation now lives in Click's argument
+    # parser (see ``rsconnect/main.py``), so it has already passed before
+    # we get here.
```
jonkeane:
```suggestion
    # check has passed. Type validation lives in Click's argument
    # parser (see ``rsconnect/main.py``), so it has already passed before
    # we get here.
```

Super minor, but this looked like a bit of claude implementation cruft?

rsconnect/api.py
```diff
@@ -139,7 +139,7 @@ def handle_bad_response(self, response: HTTPResponse | T, is_httpresponse: bool
         if isinstance(response, HTTPResponse):
             if response.exception:
                 raise RSConnectException(
-                    "Exception trying to connect to %s - %s" % (self.url, response.exception), cause=response.exception
+                    "Could not connect to %s - %s" % (self.url, response.exception), cause=response.exception
```
jonkeane, line 147: Love this cleanup

rsconnect/main.py
```diff
@@ -2893,15 +3131,15 @@ def manifest_writer(
     return manifest_writer
 
 
-generate_write_manifest_python(AppModes.BOKEH_APP, alias="bokeh")
-generate_write_manifest_python(AppModes.DASH_APP, alias="dash")
-generate_write_manifest_python(AppModes.PYTHON_API, alias="api")
+generate_write_manifest_python(AppModes.BOKEH_APP)
+generate_write_manifest_python(AppModes.DASH_APP)
+generate_write_manifest_python(AppModes.PYTHON_API)
 generate_write_manifest_python(AppModes.PYTHON_API, alias="flask", desc="Flask API")
```
jonkeane, line 3384: is it intentional that this one line with `..., alias="...", desc="..."` is kept?

docs/CHANGELOG.md
```diff
@@ -7,6 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ## Unreleased
 
+### Added
+
+- `rsconnect quickstart` command for scaffolding a new Connect-ready project.
+  Supported types: `streamlit`, `shiny`, `fastapi`, `api`, `flask`, `notebook`,
+  `voila`, `quarto`, `quarto-shiny`. Creates a uv-managed virtualenv and prints
+  the local-run and deploy commands.
+- `rsconnect deploy pyproject` command for deploying a project described by
+  `pyproject.toml` with a `[tool.rsconnect]` table containing `app_mode` and
```
jonkeane, line 27: I wish this were `tool.connect` rather than `tool.rsconnect` but I imagine this was done since the CLI is still `rsconnect` (and we aren't migrating away from that yet), yeah?

jonkeane (commented):
I tried this locally, and I'm not sure if I'm messing this up or there's somethign wrong with the template, but this is what I ran the following:

```
rsconnect quickstart shiny myshiny
rsconnect deploy pyproject -n dogfood myshiny
```

And then the app [it deployed](https://dogfood.team.pct.posit.it/connect/#/apps/fe399bea-dca7-49df-b9d8-14ca5a5838b0) doesn't seem to function

```
2026/06/01 1:07:24 PM: LANG: C.UTF-8
2026/06/01 1:07:24 PM: Working directory: /opt/rstudio-connect/mnt/app
2026/06/01 1:07:24 PM: Bootstrapping environment using Python 3.13.3 (main, Apr 22 2025, 14:51:02) [GCC 11.4.0] at /opt/python/3.13.3/bin/python3.13
2026/06/01 1:07:24 PM: Running content with the Python virtual environment /opt/rstudio-connect/mnt/app/python/env (/opt/rstudio-connect/mnt/python-environments/_nocache/107745/345188/uv/3.13.3/afBgBmio6WnsFpxwFmQl1g_20260601180607_8fcfd4005de411f184a797abaa2ad926)
2026/06/01 1:07:24 PM: Environment ready
2026/06/01 1:07:27 PM: Running content using Python "3.13.3 (main, Apr 22 2025, 14:51:02) [GCC 11.4.0]" at "/opt/rstudio-connect/mnt/app/python/env/bin/python"
2026/06/01 1:07:27 PM: Loading code from "app.py"
2026/06/01 1:07:27 PM: Error while loading Python API: Failed to find an application object (('app', 'application', 'demo')) or factory function (('create_app', 'make_app')) in module 'app'. You can specify a different entrypoint in your deployment configuration.
```

I haven't dug too much to see what the issue is, but want to get these comments up before digging too deeply.

rsconnect/quickstart/templates/api/__connect__.py.tmpl
amol-: It's more or less all "hello world"  apps, so there shouldn't be much to review there.

rsconnect/main.py
amol-: It's intentional because internally the name is "api", but we also exposed it as "flask", so both "deploy api" and "deploy flask" are meant to do the same thing.

jonkeane: Ah ok ok maybe a comment would be helpful? Or maybe it just looked weird in the diff?

rsconnect/quickstart/quickstart.py
amol-: removed!

amol-:
> ```
> rsconnect quickstart shiny myshiny
> rsconnect deploy pyproject -n dogfood myshiny
> ```
> 
> 2026/06/01 1:07:27 PM: Error while loading Python API: Failed to find an application object (('app', 'application', 'demo')) or factory function (('create_app', 'make_app')) in module 'app'. You can specify a different entrypoint in your deployment configuration.
> ```

Fixed and did a run of all content types both locally and against a remove server.

rsconnect/main.py
amol-: The same is true for `generate_deploy_python`, I think adding a comment might pollute the code more than help, the fact the second time the `AppModes.PYTHON_API` is registered has an `alias` argument seems to  communicate already that it's registering an alias to the mode.

docs/CHANGELOG.md
amol-: yes, the best practice is to name the `tool` option exactly like the name of the tool that reads it.
So as far as it's named `rsconnect` the option would be `tool.rsconnect`.

But we can definitely alias it in the future without too much effort, so that we support both historical and new name if the tool gets renamed.

rsconnect/quickstart/templates/api/__connect__.py.tmpl
jonkeane: _nod_ I know this would be a lot more coordination costs, but I do wish that we had a slightly more evocative example than just "hello world". 

This is Shiny-only, but something from https://shiny.posit.co/py/templates/ (or maybe we could call [`shiny create`](https://shiny.posit.co/py/get-started/create-run.html) for shiny and not maintain our own?)? Even if we only provided rich examples for shiny and not other frameworks, that wouldn't be a bad thing (I would prefer one rich experience to 0 even if it means less consistency; especially since we as Posit build Shiny). This can be done as a follow on, but it would be really good to do.

jonkeane (approved):
Thanks for this, I tested it out again and worked like a charm. I was actually a little surprised that I didn't even need to re-create the example app, but it must have been an issue in the pyproject -> manifest trasnlation?

amol-:
> Thanks for this, I tested it out again and worked like a charm. I was actually a little surprised that I didn't even need to re-create the example app, but it must have been an issue in the pyproject -> manifest trasnlation?

It was an issue on how "deploy" command detected the entrypoint (the path were you launch the command mattered), and pyproject inherited the same bug, but with `pyproject` it's much more likely to happen because you point the command to a directory and run it from anywhere else.

amol-:
> nod I know this would be a lot more coordination costs, but I do wish that we had a slightly more evocative example than just "hello world".

Yes, definitely. I kept this PR focused on the infrastructure, so the examples are fairly basic, but we can evolve them as we want in dedicated PRs.

One note, I experimented with reusing the "create" commands of the frameworks, and I ended up ditching that approach as it required patching files generated by another tool without any guarantee that the hook points for the patches were still valid. Shipping our own skeleton was the solution that did guarantee no surprises to users.
Also if we ship our own skeleton you don't need to have Shiny installed to start a shiny project. Shiny is installed for you as a consequence of the fact that you had quickstarted a shiny project, not as a prerequisite to quickstart one.

PR #773, opened 2026-05-28
feat: Add `login` and `logout` subcommands for OAuth flows

### LINKED ISSUE #759: `rsconnect login` using OAuth flow and DCR
Connect 2026.02.0 adds an OAuth server: https://docs.posit.co/connect/admin/authentication/oauth-clients/

Can we use this to get a token? Refresh a token? Without requiring API keys

atheriel:
## Intent

Connect version 2026.02 and later supports dynamic OAuth client registration and the authorization code flow, which is a nice way to sidestep the need for API keys.

This commit wires up these flows to new `login` and `logout` subcommands. As a bonus, we store the resulting credentials in the system keyring -- when available -- for improved security.

As a secondary bonus: we also support the device code flow (via `--use-device-code`) if the Connect server advertises its support for it.

Indicators for the OAuth client and keyring use have also been added to the `rsconnect list` output.

Resolves #759.

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Automated Tests

Unit tests are included.

## Directions for Reviewers

You can test that it works against any recent Connect server:

```
$ uv run rsconnect login -s https://dogfood.team.pct.posit.it/
$ uv run rsconnect logout -s https://dogfood.team.pct.posit.it/
```

## Checklist

### DIFF
--- a/rsconnect/metadata.py
+++ b/rsconnect/metadata.py
@@ -259,6 +259,10 @@ class ServerDataDict(TypedDict):
     account_name: NotRequired[str]
     token: NotRequired[str]
     secret: NotRequired[str]
+    oauth_client_id: NotRequired[str]
+    oauth_access_token: NotRequired[str]
+    oauth_refresh_token: NotRequired[str]
+    oauth_token_expiry: NotRequired[float]
 
 
 class ServerData:
@@ -279,6 +283,10 @@ class ServerData:
         account_name: Optional[str] = None,
         token: Optional[str] = None,
         secret: Optional[str] = None,
+        oauth_client_id: Optional[str] = None,
+        oauth_access_token: Optional[str] = None,
+        oauth_refresh_token: Optional[str] = None,
+        oauth_token_expiry: Optional[float] = None,
     ):
         self.name = name
         self.url = url
@@ -290,6 +298,10 @@ class ServerData:
         self.account_name = account_name
         self.token = token
         self.secret = secret
+        self.oauth_client_id = oauth_client_id
+        self.oauth_access_token = oauth_access_token
+        self.oauth_refresh_token = oauth_refresh_token
+        self.oauth_token_expiry = oauth_token_expiry
 
 
 class ServerStore(DataStore[ServerDataDict]):
@@ -338,6 +350,10 @@ class ServerStore(DataStore[ServerDataDict]):
         account_name: Optional[str] = None,
         token: Optional[str] = None,
         secret: Optional[str] = None,
+        oauth_client_id: Optional[str] = None,
+        oauth_access_token: Optional[str] = None,
+        oauth_refresh_token: Optional[str] = None,
+        oauth_token_expiry: Optional[float] = None,
     ):
         """
         Add (or update) information about a Connect server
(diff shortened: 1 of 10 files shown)


rsconnect/metadata.py
```diff
@@ -351,6 +367,10 @@ def set(
         :param account_name: shinyapps.io account name.
         :param token: shinyapps.io token.
         :param secret: shinyapps.io secret.
+        :param oauth_client_id: OAuth client ID from DCR.
```
zackverham: I think I'm primed to know what DCR means because I know the related body of work - fresh readers might not know, I think it could be valuable to spell it out.

rsconnect/oauth.py
```diff
@@ -0,0 +1,502 @@
...
+) -> dict[str, Any]:
+    """Perform OAuth Device Code flow.
+
+    Displays a URL and user code for the user to enter in a browser,
+    then polls for token completion.
+    """
+    device_endpoint = str(metadata.get("device_authorization_endpoint", ""))
+    if not device_endpoint:
+        raise RSConnectException(
+            "Server does not support the device code flow. " "Try again without --use-device-code."
```
zackverham: do we also need to include error messaging to cover if the connect server needs to be upgraded?

rsconnect/oauth.py
```diff
@@ -0,0 +1,502 @@
...
+    HTTP server to receive the callback. Returns the token response dict.
+    """
+    code_verifier, code_challenge = generate_pkce_pair()
+    state = secrets.token_urlsafe(32)
+
+    result_queue: queue.Queue[Tuple[str, Optional[str], Optional[str]]] = queue.Queue()
+
+    callback_server = _HTTPServer(("127.0.0.1", 0), _CallbackHandler)
+    port = callback_server.server_address[1]
+    redirect_uri = f"http://127.0.0.1:{port}/callback"
```
zackverham, line 238: I feel like I've encountered at least one customer that was running in an environment where they had networking rules that were really restrictive and ended up blocking localhost. 

I think maybe if that were to happen here, they would get this error: 

> Could not open browser automatically. Please open this URL:\n\n  {auth_url}\n


Would it be useful to give any more information about why the browser couldn't be opened automatically?

tests/test_oauth.py
```diff
@@ -0,0 +1,510 @@
...
+
+
+class TestDeviceCodeFlow:
+    def test_device_code_not_supported(self):
+        metadata = {k: v for k, v in FAKE_METADATA.items() if k != "device_authorization_endpoint"}
+        with pytest.raises(RSConnectException, match="does not support the device code flow"):
+            login_with_device_code(FAKE_URL, "client-1", metadata)
+
+    @patch("rsconnect.oauth.time.sleep")
+    def test_poll_success(self, mock_sleep: MagicMock, mock_http_server: MagicMock):
```
zackverham: maybe I'm misreading something - but is mock_sleep used at all? This might be a dead parameter - I'm also not sure what the patch decorators are doing.

zackverham:
Question for my own understanding - I'm trying to follow what happens if they keyring isn't available. Is that fallback implemented in the `keyring` library? It looks like rsconnect-python is just calling keyring behavior and itself doesn't implement the fallback.

zackverham (approved):

atheriel:
> Question for my own understanding - I'm trying to follow what happens if they keyring isn't available.

If it isn't available then we store the tokens in the JSON file, like we do for API keys.

tests/test_oauth.py
atheriel: It's not used but apparently it's required by the patch decorator due to positional arguments. I'll mark it as such.

rsconnect/metadata.py
atheriel: Right. I'm just going to axe "from DCR" here entirely, it doesn't serve any real purpose.

rsconnect/oauth.py
atheriel: For sure. And that it can be intentionally disabled.

rsconnect/oauth.py
atheriel: Yep, makes sense. Done.

PR #777, opened 2026-06-01
feat: Allow setting a default server

### LINKED ISSUE #214: Ability to define a default server nickname
It would be nice if you could specify a default server in the `servers.json` so that you don't need to supply the `-n` argument to every command.  Today, the `-n` argument is always required when you have >1 server alias defined

atheriel:
## Intent

With more than one entry in the `servers.json` file, the `-s` or `-n` would previously have become mandatory. This commit allows a server to be marked as the default, in which case it is assumed to be the target.

Closes #214.

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach

Specifically, `rsconnect add` now has a `--set-default` flag to mark a server as the default. For backwards compatibility we do not do this automatically.

For `rsconnect login`, on the other hand, we don't need to be backwards compatible -- it hasn't been released yet -- and so that command will make the server the default automatically (unless `--no-set-default` is passed).

Some care has been taken to keep precendence of flags and environment variables intact and ensure that only one server is the default at a time.

## Automated Tests

Unit tests are included.

## Directions for Reviewers

Please advise.

## Checklist

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   write-manifest.
 - Perform case insensitive matching of the configured Snowflake connection authenticator.
 - New `login` and `logout` subcommands for authenticating to Connect via OAuth.
+- Servers can now be marked as the default with `rsconnect add --set-default`.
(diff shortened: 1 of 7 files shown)


edavidaja (approved):

edavidaja:
jackpot

PR #785, opened 2026-06-16
support for parsing renv.lock

### LINKED ISSUE #784: rsconnect-python support for generating R dependencies in manifest via renv.lock
When deploying content using `rpy2`, the project will usually depend on R libraries that need to be available on Connect at deploy time.
The `manifest.json` generated during deploy should handle this and install R dependencies when those are listed in the manifest, but `rsconnect-python` currently is unable to detect that an `renv.lock` exists and inject those in the manifest.

We need to have `renv.lock` detection and manifest injection similar to what the publisher does:

* https://github.com/posit-dev/publisher/blob/281c193d36d1fc086425d5fada4d6e9cc7732006/extensions/vscode/src/publish/dependencies.ts#L42-L73
* https://github.com/posit-dev/publisher/blob/main/extensions/vscode/src/publish/rPackageDescriptions.ts#L64-L103

amol-:
## Intent

Closes #784 

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach

Implement a `r_environment` that detects `renv.lock` and parses the dependencies.

## Checklist


rsconnect/environment_r.py
```diff
@@ -0,0 +1,282 @@
+"""Detects R dependencies from a project's renv.lock file.
+
+Given a directory that contains an renv.lock lockfile, this module parses it
+into the R version and package metadata needed for the deployment manifest.
+The parse is pure: it reads only renv.lock and never invokes R or inspects
+locally installed R packages. This mirrors how Posit Publisher resolves R
+dependencies for Python content that also uses R (e.g. rpy2 apps).
```
karawoo, line 7: This means that lockfiles outside of the project root will not be respected, correct? Lockfiles can legitimately be in other locations, either through setting `RENV_PATHS_LOCKFILE` or if the user is using renv profiles. 

In the rsconnect R package this is supported (https://github.com/rstudio/rsconnect/pull/1296). If we are choosing not to support it here, or not for now (which I think is a reasonable choice) then could we document that?

rsconnect/actions.py
karawoo, line 270: Perhaps knitr should be supported now? I believe it's the default for quarto docs with both python and R blocks.

karawoo (commented):
The code looks good to me and I was able to generate several correct manifests with it. A couple of minor questions below.

As far as I understand we don't expect this to actually work on Connect just yet, and indeed I was not able to deploy my content (quarto project using jupyter engine) to Connect. But the manifest itself looked fine, so I don't think it's an issue with anything here.

rsconnect/environment_r.py
amol-: I did mimic the code in publisher, but I realised now that publisher does invoke the `R` interpreter thus it would probably handle the profiles and the `RENV_PATHS_LOCKFILE` by virtue of that.

Invoking the `R` interpreter didn't seem a good idea in `rsconnect-python` because I don't want to force the need to install an R environment just to deploy a python project.

As a trade-off I support for `RENV_PATHS_LOCKFILE` as it was easy enough to support and documented the behavior. I think that in the majority of `rpy2` projects it will be perfectly fine to support the `renv.lock` exclusively in the project directory as it will be shipped with the project itself, but `RENV_PATHS_LOCKFILE` is a solid escape hatch for exceptions

https://github.com/posit-dev/rsconnect-python/pull/785/commits/d4717afd829328fc49c124e7c0db46294311c726

rsconnect/actions.py
amol-: Sounds reasonable, but as this was an existing behavior I'd delegate it to its own issue to avoid mixing too many concerns in one PR.

https://github.com/posit-dev/rsconnect-python/issues/786

amol-:
> As far as I understand we don't expect this to actually work on Connect just yet, and indeed I was not able to deploy my content (quarto project using jupyter engine) to Connect. But the manifest itself looked fine, so I don't think it's an issue with anything here.

Correct, that is undergoing work in connect side

rsconnect/environment_r.py
```diff
@@ -0,0 +1,297 @@
...
+        )
+
+
+def _renv_lockfile_path(directory: str) -> str:
+    # Mimics renv's renv_paths_lockfile() in R/paths.R: RENV_PATHS_LOCKFILE
+    # overrides the location and is used verbatim, except a trailing slash means
+    # "a directory" so renv.lock is appended. A relative override resolves against
+    # the current working directory (matching renv), not the project directory.
+    # With no override we fall back to renv's default of <project>/renv.lock.
+    override = os.environ.get("RENV_PATHS_LOCKFILE")
```
karawoo: This comment isn't correct about renv's behavior as of the most recent renv version: it resolves against the project directory, not the working directory.

https://github.com/rstudio/renv/blob/538c71944aab9b02d78ddaab626e70713ffcc1f4/R/paths.R#L66-L70

(introduced in https://github.com/rstudio/renv/commit/f0ab13530f36579e4557e4ca6320c7f638fa83c9)

Could we adjust the behavior to match and add a test for relative paths in `RENV_PATHS_LOCKFILE`?

amol-: https://github.com/posit-dev/rsconnect-python/pull/785/commits/02c49ad86dd18a15f4bea46e9206a2f7eb1395fe adjusts `RENV_PATHS_LOCKFILE` to be relative to the project directory.

amol-:
The failure on Python 3.8 is unrelated, it has to do with the release of mistune that is now incompatible with Python 3.8, the same failure happens on main.

karawoo (approved):

PR #789, opened 2026-06-22
fix(nodejs): add python env exclusion for nodejs manifest generation

### LINKED ISSUE #783: Exclude python virtual environemnts from manifest for Nodejs apps
I was deploying a nodejs app via Git-backed publishing and created the manifest.json with python-rsconnect. In doing so, my manifest file include all the files in my .venv. My .venv appeared to be properly .gitignored. 

I am not sure, but I think it makes sense to exclude .venv by default in this case?

## recording

https://github.com/user-attachments/assets/125c6ef4-1bcd-49cd-be72-4e0256ea5929

## vs code version

Version: 1.121.0 (Universal)
Commit: f6cfa2ea2403534de03f069bdf160d06451ed282
Date: 2026-05-19T11:51:53+02:00 (3 wks ago)
Electron: 39.8.8
ElectronBuildId: 13870025
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.5.0

## rsconnect version
1.29.0

joshyam-k:
Exculde python env files in nodejs manifest generation.

## Intent
resolves #783 

## Type of Change

- [x] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change   

## Approach

`make_nodejs_manifest` only excluded `manifest.json` and `node_modules`.  It needs to also exclude python env files. Every other content generator strips these via `list_environment_dirs(...)` so we just have to do that here as well

## Automated Tests

`test_manifest_excludes_python_virtual_environment` creates a venv-shaped .venv and asserts none of its files appear in the manifest. Confirmed it fails without the fix and passes with it.

## Checklist

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -50,6 +50,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `uv venv` and `uv sync` to populate the scaffolded project's virtualenv.
   `uv` installs as a self-contained wheel from PyPI alongside `rsconnect`.
 
+### Fixed
+
+- Python virtual environments (e.g. `.venv`) in a Node.js project directory are
+  now excluded from the generated `manifest.json`, matching the behavior of the
+  Python and Quarto content types.
+
 ## [1.29.0] - 2026-04-29
 
 - Added `rsconnect deploy nodejs` command for deploying Node.js applications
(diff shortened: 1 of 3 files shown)


dotNomad (approved):
Very clean and very easy approval ✅

PR #792, opened 2026-06-23
fix: Detect invalid API keys more carefully

atheriel:
## Intent

When credentials come from the command line (or environment variables) rather than the store, we attempt to validate them by hitting the `/v1/user` endpoint, and error out if this fails.

It turns out that this doesn't work for machine credentials, which aren't associated with a user.

Thankfully the difference between "invalid API key" and "not a user API key" can be detected based on the error code from the response, so this commit draws this exact distinction.

## Type of Change

- [x] Bug Fix            
- [ ] New Feature       
- [ ] Breaking Change   

## Automated Tests

Unit tests are included.

## Checklist

### DIFF
--- a/rsconnect/api.py
+++ b/rsconnect/api.py
@@ -1554,23 +1554,13 @@ class RSConnectExecutor:
     def verify_api_key(self, server: Optional[RSConnectServer] = None):
         """
         Verify that an API Key may be used to authenticate with the given Posit Connect server.
-        If the API key verifies, we return the username of the associated user.
         """
         if not server:
             server = self.remote_server
         if isinstance(server, ShinyappsServer):
             raise RSConnectException("Shinnyapps server does not use an API key.")
         with RSConnectClient(server) as client:
-            result = client.me()
-            if isinstance(result, HTTPResponse):
-                if (
-                    result.json_data
-                    and isinstance(result.json_data, dict)
-                    and "code" in result.json_data
-                    and result.json_data["code"] == 30
-                ):
-                    raise RSConnectException("The specified API key is not valid.")
-                raise RSConnectException("Could not verify the API key: %s %s" % (result.status, result.reason))
+            verify_api_key_response(client)
         return self
 
     @property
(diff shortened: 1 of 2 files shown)


rsconnect/api.py
```diff
@@ -2053,27 +2043,56 @@ def verify_server(connect_server: RSConnectServer):
...
+                cause=result.exception,
+            )
+        json_data = result.json_data if isinstance(result.json_data, dict) else {}
+        code = json_data.get("code")
+        # A service principal or machine identity authenticates successfully but has no
+        # associated user, so the v1/user endpoint rejects it with a 403 and error code
+        # 22. That code is unambiguous on this endpoint -- a genuinely invalid credential
+        # is rejected at the auth layer with code 30 instead -- so the credential is valid
+        # and we treat it as verified.
+        if result.status == 403 and code == 22:
```
nealrichardson, line 2070: Is it not possible to put this logic in the standard `handle_bad_response` handler?

rsconnect/api.py
```diff
@@ -2053,27 +2043,56 @@ def verify_server(connect_server: RSConnectServer):
...
+        # handle_bad_response.
+        if result.exception:
+            raise RSConnectException(
+                "Exception trying to connect to %s - %s" % (result.full_uri, result.exception),
+                cause=result.exception,
+            )
+        json_data = result.json_data if isinstance(result.json_data, dict) else {}
+        code = json_data.get("code")
+        # A service principal or machine identity authenticates successfully but has no
+        # associated user, so the v1/user endpoint rejects it with a 403 and error code
```
nealrichardson, line 2065: Is it necessary that we request the `v1/user` endpoint to validate the API key, or the OIDC token? I noticed that nothing else uses the `.me()` endpoint other than to check that an API key is valid.

nealrichardson (approved):
I confirmed that this works, I just have some design questions. I'm ok to merge this and make a followup to refactor if there's no other simplification we can make here.

rsconnect/api.py
atheriel: I also found this confusing. I believe the original point of it was to surface more user-friendly error messages, but I agree that I would personally prefer we just fail later on when attempting an actual operation. That felt like a more invasive behaviour change to make, though.

rsconnect/api.py
atheriel: Partially, yes. Fixed.

PR #795, opened 2026-06-24
feat: warn when a newer version is available

### LINKED ISSUE #779: CLI should warn when a newer version is available
## Problem

Users can end up running very outdated versions of `rsconnect-python` without realizing it. When the CLI hits removed API endpoints on newer versions of Connect, the error messages are confusing (e.g., `404 Not Found` on `/__api__/applications/{id}/upload`) and don't hint that upgrading the tool would fix the problem.

## Proposal

Add a version check that runs on CLI invocation (or at least on `deploy` commands) and prints a notice if a newer version is available on PyPI. Something like:

```
WARNING: You are using rsconnect version 0.1.3; however, version 1.29.0 is available.
You should consider upgrading via the 'pip install --upgrade rsconnect' command.
```

This is similar to how `pip` itself warns about outdated versions, and how `conda` shows "A newer version of conda exists."

### Considerations

- The check should be non-blocking and fail silently (don't break deploys if PyPI is unreachable)
- Could cache the check result for some period (e.g., 24 hours) to avoid slowing down every invocation
- Could be disableable via a flag or environment variable (e.g., `RSCONNECT_NO_UPDATE_CHECK=1`)
- Should only print to stderr so it doesn't interfere with scripted/CI usage

joshyam-k:
Print a warning message to stderr when a newer version of the cli is available.

## Intent
Fixes #779

## Type of Change

- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach
  - check PyPI for a newer version (only on deploy commands) and print the upgrade hint on every exit path, including failed/fast-failing deploys
  - cache the result on disk with a simple 24H TTL. 
  - the fetch runs in a background thread; only a cold cache waits on it (bounded by a 2s PyPI timeout) otherwise the failure can beat the fetch and the warning message is never printed.
  - fail silently: if the fetch errors or doesn't finish within the timeout, the deploy proceeds with no hint (and no error)
  - scoped to deploy only, so quick commands like `list`/`info` never pay the cold-cache wait.
  - can be disabled via `RSCONNECT_DISABLE_VERSION_CHECK=1`

## Automated Tests

Unit tests for each new helper in isolation and also a `TestCLIIntegration` test that tests the feature end-to-end through the cli.

## Checklist

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- `rsconnect deploy` commands now check PyPI once a day for a newer release of
+  rsconnect-python and print an upgrade hint to stderr when one is available.
+  The result is cached so most invocations make no network request. Set
+  `RSCONNECT_DISABLE_VERSION_CHECK=1` to turn the check off entirely.
 - The "no package-lock.json" error when deploying Node.js content now states
   that Connect installs Node.js dependencies with npm, helping publishers who
   build with yarn or pnpm understand why a `package-lock.json` is required.
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -432,6 +432,21 @@ is uploaded and deployed. If the deployment fails, the new environment variables
 will still take effect.
 
 
+### Update notifications
+
+When you run a `rsconnect deploy` command, rsconnect-python checks PyPI for a
+newer release and prints an upgrade hint to stderr if one is available. The
+result is cached for 24 hours, so most deployments make no extra network
+request, and the check never blocks or fails a deployment.
+
+To disable the check entirely, set the `RSCONNECT_DISABLE_VERSION_CHECK`
+environment variable to `1` (or `true`/`yes`):
+
+```bash
+export RSCONNECT_DISABLE_VERSION_CHECK=1
+```
+
+
 ### Updating a Deployment
 
 If you deploy a file again to the same server, `rsconnect` will update the previous
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,7 @@ dependencies = [
     "semver>=2.0.0,<4.0.0",
     "pyjwt>=2.4.0",
     "click>=8.0.0",
+    "packaging>=20.0",
     "toml>=0.10; python_version < '3.11'",
 ]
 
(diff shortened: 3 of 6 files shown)


marcosnav (commented):
This is a great addition. I have some usability feedback, I forced the output from `_fetch_latest_version` and while trying this out I noticed that the new version warning message only triggers on successful deployments.

For example, a project with missing `requirements.txt` shows this (no warning about a new version):
Image

It seems that any error that goes through the exception handler (which covers `RSConnectException` and general `Exception`) calls `sys.exit(1)` and the warning never prints.

So in practice, you'd only see the version warning after a successful deploy (or possibly a very late failure like a task timeout on the server side).

That seems like a significant gap, users that most likely could benefit from "hey, upgrade" are the ones hitting bugs that were already fixed.

Also, how complicated is to have this warning for all commands `info`, `version`, `list`...?

joshyam-k:
@marcosnav huh good catch I thought I fixed that. My guess is maybe on fast failures the error beats the fetch command and so we don't get the warning.

joshyam-k:
@marcosnav okay I figured out the issue and it stems from an issue that we have to resolve in what our approach is here.

There are 3 things we want with this feature:

1. Warn on every exit path
2. Warn even with no cache
3. Do not block on the check

The issue is that 2 and 3 are actually at odds with each other because if there is no cache (i.e first time running it in over 24h) then the check **must** be blocking in order to fire before the command finishes (in the case of a fast failure i.e missing requirements.txt)

Maybe the best we can do here to satisfy all criteria is to make the check blocking but only when the cache is empty. Otherwise there's just no way around avoiding the case that you caught above.

marcosnav:
> then the check **must** be blocking in order to fire before the command finishes (in the case of a fast failure i.e missing requirements.txt)

@joshyam-k  Just to confirm, by **blocking** you mean a synchronous wait on the version check right? If that is the case I think it is ok to wait. Maybe just to avoid hanging issues, we could have a timeout of about 3s to the PyPI check and discard the check if it takes longer than that so the command continues. WDYT?

joshyam-k:
@marcosnav yes correct. And if we go that route I'm even more inclined to keep this scoped to just deploy commands. We don't want multi second latency on things like `rsconnect list` just because we are blocking on checking the version. In deploy we usually can just hide behind what is already a multi-second process.

marcosnav:
@joshyam-k Yes, that sounds right. BTW, there is already a `_PYPI_TIMEOUT_SECONDS = 2` so ignore my timeout comment, it is already there.

joshyam-k:
@marcosnav ok ready for another look!

marcosnav (approved):
This is great! :shipit: 

Tried it out hardcoding the check version to 99.0.0

**I see the warning on deployment failures**

Image

**Also on deployment success**

Image


And, as expected, the message does not show when:
- The version cache is present, (in my case at `~/Library/Application\ Support/rsconnect-python/version_check.json`)
- or using the env var `RSCONNECT_DISABLE_VERSION_CHECK=true`

PR #802, opened 2026-06-26
feat: exchange OIDC identity tokens for an API key in `rsconnect login`

### LINKED ISSUE #800: rsconnect login with OIDC token
Wrap the trusted publisher auth flow like `rsconnect login <server> --token <token>` or something.

nealrichardson:
## Summary

Closes #800.

Adds OIDC identity-token exchange to the `rsconnect login` command, so trusted publishers (e.g. GitHub Actions) can obtain a Connect API key without interactive OAuth login.

### How it works
1. `rsconnect login <server> --identity-token <oidc-token>` (or `--identity-token-file <path>`, or the `CONNECT_IDENTITY_TOKEN` / `CONNECT_IDENTITY_TOKEN_FILE` env vars). Use `--identity-token -` to read from stdin; prefer the file/env forms to keep the secret out of process arguments and CI logs.
2. Discovers the server's OAuth metadata to confirm token-exchange support and resolve the correct token endpoint (preserving any path prefix and query).
3. Performs an RFC 8693 token exchange (`grant_type=token-exchange`, `subject_token_type=id_token`, `requested_token_type=access_token`). Connect verifies the token against a configured trusted publisher and mints a short-lived API key.
4. Validates the key (`test_server` + `test_api_key`) and stores it as the server credential, honoring `--name`, `--insecure`, `--cacert`, and `--no-set-default`. Subsequent `deploy`/`list`/etc. then just work.

Failure responses are mapped to actionable errors: unsupported token exchange, connection/TLS errors, and `invalid_grant` (ambiguous match, verification failure, no match).

## Testing
- New tests in `tests/test_oauth.py` cover metadata discovery, the exchange success/error paths, endpoint preservation, connection-error handling, and the CLI flag / file / stdin / empty-token behavior.
- `black` and `flake8` clean on changed files; `pyright` reports only pre-existing errors.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   write-manifest.
 - Perform case insensitive matching of the configured Snowflake connection authenticator.
 - New `login` and `logout` subcommands for authenticating to Connect via OAuth.
+- `rsconnect login --token <oidc-token>` exchanges an OIDC token (such as a
+  GitHub Actions OIDC token) for a short-lived Connect API key via Connect's
+  trusted-publishing token exchange, and saves it as the server credential.
+  Use `--token -` to read the token from stdin.
 - Servers can now be marked as the default with `rsconnect add --set-default`.
   When neither `-n/--name` nor `-s/--server` is provided, the default server is
   used automatically. `rsconnect login` sets the server as default unless
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,10 +14,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   write-manifest.
 - Perform case insensitive matching of the configured Snowflake connection authenticator.
 - New `login` and `logout` subcommands for authenticating to Connect via OAuth.
-- `rsconnect login --token <oidc-token>` exchanges an OIDC token (such as a
-  GitHub Actions OIDC token) for a short-lived Connect API key via Connect's
-  trusted-publishing token exchange, and saves it as the server credential.
-  Use `--token -` to read the token from stdin.
+  `login` opens a browser interactively (or uses `--use-device-code` for
+  headless environments). Alternatively, `--identity-token` (or
+  `--identity-token-file`, and the `CONNECT_IDENTITY_TOKEN` /
+  `CONNECT_IDENTITY_TOKEN_FILE` environment variables) exchanges an OIDC
+  identity token, such as a GitHub Actions OIDC token, for a short-lived
+  Connect API key. In all cases the resulting credential is saved for the
+  server.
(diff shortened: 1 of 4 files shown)


rsconnect/oauth.py
```diff
@@ -452,6 +452,110 @@ def refresh_access_token(
...
+    """Translate a failed token-exchange response into an actionable exception."""
+    error = str(data.get("error", "")) if data else ""
+    description = str(data.get("error_description", "")) if data else ""
+
+    if status == 404:
+        return RSConnectException(
+            "This Connect server has no OIDC token-exchange endpoint (/oauth/v1/token). "
+            "It is likely too old to support trusted publishing; upgrade Connect, or "
+            "authenticate with an API key instead."
+        )
```
atheriel: For the device code and regular authorization code flow, we check whether the grant type is supported using Connect's discovery endpoint. Could we do the same here as a preflight check rather than failing on 404s?

rsconnect/main.py
```diff
@@ -1053,6 +1101,14 @@ def remove(
...
     help="Path to a trusted CA certificate file for TLS.",
 )
+@click.option(
+    "--token",
+    default=None,
+    help=(
+        "An OIDC token to exchange for a Connect API key via trusted-publishing token exchange "
+        "(RFC 8693), instead of interactive OAuth login. Use '-' to read the token from stdin."
+    ),
+)
```
atheriel, line 1140: I have a strong preference for `--identity-token` as the flag to disambiguate from other kinds of "token". Ideally we'd also support a `CONNECT_IDENTITY_TOKEN` environment variable, or, better, `CONNECT_IDENTITY_TOKEN_FILE`. Since passing the plaintext token as a flag might be weird in CI/CD logs.

I also think we should avoid verbiage specific to trusted publishing here, because in theory we could also support identity federation using this flag.

docs/CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   write-manifest.
 - Perform case insensitive matching of the configured Snowflake connection authenticator.
 - New `login` and `logout` subcommands for authenticating to Connect via OAuth.
+- `rsconnect login --token <oidc-token>` exchanges an OIDC token (such as a
+  GitHub Actions OIDC token) for a short-lived Connect API key via Connect's
+  trusted-publishing token exchange, and saves it as the server credential.
+  Use `--token -` to read the token from stdin.
```
atheriel: Since `rsconnect login` hasn't been released yet, I haven't added NEWS items for follow-up features. But I'm fine either way.

atheriel (commented):
This makes complete sense to do.

docs/CHANGELOG.md
nealrichardson: Yeah we can fold all of the `rsconnect login` features into a single bullet.

atheriel (approved):
😎

PR #805, opened 2026-06-29
feat: verify deployment on a draft, activate on success

### LINKED ISSUE #768: bug: verify_deployment doesn't verify a draft
https://github.com/posit-dev/rsconnect-python/blob/f6cd9554669670dd00c021e5e4436ee8c5b02ac7/rsconnect/api.py#L1234

If using the `--draft` flag to deploy, we should verify that the draft version is working. If the point of the draft is to confirm a fix for a broken app, but we check the active version, we'll get a false failure.

### LINKED ISSUE #769: feat: verify deployment on a draft, activate on success
A better fix for #768 too. 

Verify-before-activate workflow:

  1. Deploy as draft — `POST /v1/content/:guid/deploy` with `activate: false`. Returns a bundle ID.
  2. Hit the preview URL to verify — draft bundles are accessible at `/content/:guid/_bundle:id`. The process spins up cold
  (`MinProcesses=0`) on that request, so a successful HTTP response confirms it actually runs.
  3. Activate — `POST /v1/content/:guid/deploy` again with the same `bundleId` and `activate: true`. This is what the "Activate Draft" button in the UI does. Only do this if the verify step succeeds. If it does not, we do not activate the bundle. That way, the active bundle should never get a broken build.

nealrichardson:
Closes #769. Also fixes #768.

## What

Makes verify-before-activate the default for `rsconnect deploy`, on Connect servers that support it:

1. **Deploy as a draft** — the new bundle is deployed with `activate: false`.
2. **Verify the draft** — `access_content` hits the draft bundle's preview URL (`/content/<guid>/_bundle<id>/`), which cold-starts the process, so a non-5xx response confirms the content actually runs.
3. **Activate on success** — a second `POST .../deploy` with `activate: true` (the "Activate Draft" button's behavior). If verification fails, the bundle is left as a draft and the previously-active bundle keeps serving, so a broken build never becomes the active version.

This also fixes #768: `--draft` now verifies the *draft* bundle rather than the currently-active content.

### Flag behavior (Connect >= 2025.06.0)

| flags | deploy | verify | activate |
|-------|--------|--------|----------|
| (none) | draft | draft preview URL | yes, on success |
| `--draft` | draft | draft preview URL | no |
| `--no-verify` | active | — | immediate |

### Version gating

Connect added the `activate` field on the deploy endpoint in **2025.06.0**; older servers reject the unknown field. So the draft-first flow is gated on server version:

- **Connect >= 2025.06.0:** the flow above.
- **Older Connect / shinyapps.io:** deploy and activate in one step (no `activate` field sent) and verify the active content — i.e. the pre-PR behavior. `--draft` against these servers is unsupported (it cannot be expressed without the `activate` field).

## Changes

- `api.py`:
  - `access_content` takes an optional `bundle_id` and accesses the draft preview URL when given one.
  - `verify_deployment` targets the draft bundle when one was deployed (not yet active), otherwise the active content.
  - new `activate_deployment` method (re-issues the deploy with `activate: true`).
  - `bundle_id` threaded into the deploy result.
  - `server_supports_draft_deploy(version)` helper + `RSConnectExecutor.supports_verify_before_activate` (cached) and `should_deploy_as_draft(...)` to centralize the gating decision.
  - fixes a pre-existing `//content` double-slash in the access path.
- `main.py`: deploy call sites use `should_deploy_as_draft` / `supports_verify_before_activate`; `--draft` / `--no-verify` help text updated.
- `tests/test_main.py`:
  - unit tests for the draft -> verify -> activate sequence and for the old-server fallback (no `activate` field, single deploy, verify active).
  - two live-Connect integration tests (run via `posit-dev/with-connect`, gated to Connect >= 2025.06.0) that fail without these changes:
    - `test_default_deploy_does_not_activate_broken_bundle` (#769) — a broken redeploy fails and leaves the previously-active working bundle serving.
    - `test_draft_deploy_verifies_the_draft_not_the_active_bundle` (#768) — deploying a broken draft fails verification; previously `--draft` verified the still-good active content and reported success.

### DIFF
--- a/rsconnect/api.py
+++ b/rsconnect/api.py
@@ -602,14 +602,13 @@ class RSConnectClient(HTTPServer):
 
     def access_content(self, content_guid: str, bundle_id: Optional[str] = None) -> None:
         method = "GET"
-        base = dirname(self._url.path)  # remove __api__
-        if bundle_id is not None:
-            # Access a specific (e.g. draft, not-yet-activated) bundle's preview URL.
-            # Connect spins the process up cold to serve this, so a successful response
-            # confirms the bundle actually runs without touching the active bundle.
-            path = f"{base}/content/{content_guid}/_bundle{bundle_id}/"
-        else:
-            path = f"{base}/content/{content_guid}/"
+        base = dirname(self._url.path).rstrip("/")  # strip "__api__" and any trailing slash
+        # Access a specific (e.g. draft, not-yet-activated) bundle's preview URL when a
+        # bundle id is given. Connect spins the process up cold to serve this, so a
+        # successful response confirms the bundle actually runs without touching the
+        # active bundle.
+        suffix = f"_bundle{bundle_id}/" if bundle_id is not None else ""
+        path = f"{base}/content/{content_guid}/{suffix}"
         response = self._do_request(method, path, None, None, 3, {}, False)
 
         if self.is_failed_response(response):
--- a/rsconnect/api.py
+++ b/rsconnect/api.py
@@ -390,6 +390,7 @@ class RSConnectClientDeployResult(TypedDict):
     app_url: str
     dashboard_url: str
     draft_url: str | None
+    bundle_id: str | None
     title: str | None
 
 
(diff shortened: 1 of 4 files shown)


rsconnect/api.py
```diff
@@ -1472,14 +1504,74 @@ def save_deployed_info(self):
 
         return self
 
+    @property
+    def supports_verify_before_activate(self) -> bool:
+        """Whether the target server supports deploying a bundle as a draft and
+        activating it separately. shinyapps.io / Posit Cloud and pre-2025.06.0 Connect
+        do not, so for those we deploy and activate in one step and verify the active
+        content instead."""
```
marcosnav, line 1512: One caveat, Connect configured with `Server.HideVersion` will hide the version even for authenticated users

marcosnav (approved):
This looks good, with one caveat for `Server.HideVersion`. Not sure how often customers configure Connect with it, but it'll block the drafts feature in those cases even for recent versions of Connect. Something to keep in mind for docs maybe.

rsconnect/api.py
nealrichardson: I'm not sure there is a good way around this because if you try to deploy a draft and the version of Connect doesn't support it, it errors badly, and that means you have to upload the bundle again, which could be expensive. 

I had an idea for how to work around this though, WDYT? #807

marcosnav: Yeah, I think setting an env var is a good workaround, it is an easy thing to do to unblock folks

Issue #774, opened 2026-05-29
fix pyright errors


karawoo:
pyright was added to `make lint` in #567, but it was immediately suppressed with a leading `-` in the Makefile so it doesn't fail the build. Currently there are 110 pyright errors. I believe these are all things that we at one point intended to enforce but are not currently doing so.

This issue is to do cleanup of those 110 errors, and then turn on real enforcement by removing the `-` 

https://github.com/posit-dev/rsconnect-python/blob/19f54159428daf20eb07b9c77b29823166538de1/Makefile#L59-L64

PR #808, opened 2026-06-30
Modernize tooling: uv build/test, just, ruff (drop Docker for Python envs & Make)

edavidaja:
## Summary

Modernizes the build and test toolchain onto an astral-native stack, removing Docker where it was only providing Python environments, and removing `make`. The Connect-**server** Docker setups (`docker-compose.yml`, `vetiver-testing/`, `integration-testing/`) are intentionally kept — `uv` cannot run the Connect product.

- **Build backend:** `setuptools` + `setuptools_scm` → `uv_build` (flat-layout `module-name`/`module-root`).
- **Versioning:** git-tag-derived → static `[project].version` managed with `uv version`. `main` carries a `.dev` version (`1.29.1.dev0`) so dev installs stay pre-release (keeps the update-nag off) and never collide with a published release. Runtime `VERSION` now resolves via `importlib.metadata` (both published dist names). A CI guard asserts the tag matches the version on tag builds.
- **Dev tooling:** moved to PEP 735 `[dependency-groups]` so `ruff`/`pyright`/`twine`/`pytest`/mkdocs are **not** advertised in the published wheel metadata. Genuine runtime features (`keyring`, `snowflake`, `mcp`) remain extras.
- **Lint/format:** `black` + `flake8` → `ruff` (format + check). `pyright` kept, **advisory** (matches today's behavior; see #774).
- **Multi-version testing:** per-version Docker images → `uv run --python <X>`.
- **Task runner:** `Makefile` → `Justfile`.
- **CI:** `main.yml` + `preview-docs.yml` run on `astral-sh/setup-uv` + `just`; integration jobs keep the Docker Connect server but install/run via `uv`. Added a `uv lock --locked` drift check and a wheel-contents assertion (py.typed + quickstart templates present, tests excluded).

**Deleted:** `Dockerfile`, `docs/Dockerfile`, `Makefile`, `setup.py`, `scripts/build-image`, `rsconnect/version.py`.

## Testing / review notes
- `just lint`, `just test <ver>`, and `just dist` (wheel + sdist, `twine check` PASSED) verified locally; built wheel confirmed to contain `rsconnect/py.typed` and the quickstart templates and to exclude `tests/`.
- Pre-existing, out-of-scope: 10 test failures present at the base commit (env-specific `orjson`/Cargo build for the `shiny` quickstart + `test_main_content` TEMP_DIR paths) and 131 pre-existing pyright strict errors (#774, why pyright is advisory).
- Fixed during review: the `test-dev-connect` job now passes `--vetiver` so vetiver tests actually run (the conftest skip-hook was previously inert and is now correctly applied).

## Draft — open items before marking ready
- [ ] Full CI run (multi-version matrix, integration jobs against live Connect) is green.
- [ ] Confirm the `uv_build>=0.9.0,<0.10.0` pin does not warn under the CI uv version.


nealrichardson:
This is great. Can you also remove the .vscode settings file that insists on installing black?

.vscode/extensions.json
```diff
@@ -1,6 +1,6 @@
 {
   "recommendations": [
     "ms-python.vscode-pylance",
```
nealrichardson: I think we should drop pylance too

Justfile
```diff
@@ -0,0 +1,81 @@
...
+
+# Remove build/test artifacts
+clean:
+    rm -rf .coverage .pytest_cache build dist htmlcov rsconnect_python.egg-info rsconnect.egg-info site
+
+# Remove local rsconnect store directories
+clean-stores:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    find . -name "rsconnect-python" -o -name "rsconnect_python-*" -o -name "rsconnect-*" | xargs rm -rf
```
karawoo:
```suggestion
    find . -type d \( -name "rsconnect-python" -o -name "rsconnect_python-*" \) -exec rm -rf {} +
```

I think this is safer. in the event there is any whitespace in the filepaths, xargs will split on it and might end up `rm -rf`ing the wrong thing. also `rsconnect-*` feels like a bit too wide of a net to me.

karawoo (approved):

PR #809, opened 2026-06-30
Drop vetiver test baggage; mock system-caches CLI test

edavidaja:
## What & why

Removes rsconnect-python's vetiver-specific test baggage and converts its own `system caches` integration test from a live-Connect test into an offline `httpretty`-mocked unit test. rsconnect no longer maintains a Connect-in-Docker harness for testing.

Pairs with rstudio/vetiver-python#243, which moves vetiver's own Connect integration tests fully into the vetiver repo (on `with-connect`).

## Key insight

`rsconnect system caches list/delete` is a thin wrapper over `GET`/`DELETE v1/system/caches/runtime`. Connect's **permission enforcement** on that endpoint (admin allowed; publisher/viewer/anon → 403) is already covered upstream in `connect`'s API suite. The old test paid for a live Connect + a second non-admin user just to re-test Connect's behavior through the CLI. We instead test only rsconnect's CLI layer — offline, with `httpretty` (the repo's standard mocking) — and rely on upstream for enforcement. Net result: **no live Connect / no `with-connect`** needed here, and the test now runs on every PR instead of nightly.

## Changes

- **`tests/test_main_system_caches.py`** — rewritten as offline `httpretty` tests: `list` happy path (exit 0 + JSON round-trip), `delete` happy path (handles the real task-polling flow), required-flag validation (exit 2), and 403 → exit 1 + permission message. `allow_net_connect=False`.
- **`conftest.py` / `pyproject.toml`** — removed the `--vetiver` option/marker plumbing and the `vetiver-testing` ruff exclude; deleted `tests/test_vetiver_pins.py`.
- **`.github/workflows/main.yml`** — deleted the `test-dev-connect` job (nothing live to run anymore).
- Deleted `vetiver-testing/`, root `docker-compose.yml`, and the `dev`/`dev-stop` `Justfile` recipes.
- **`rsconnect/actions.py`** — relabeled the `deploy_python_fastapi`/`deploy_app` block comments from "deprecated … will be removed" to "supported compatibility entry point used by vetiver-python" (comments only; the shim is intentionally **kept** — vetiver still calls it).

## Notes

- Entirely offline — no license/Docker needed to run these tests.
- Full offline suite: 728 passed. The 4 `test_quickstart.py[shiny]` failures are pre-existing and unrelated to this change.
- Based on `uv-tooling-modernization` (its file edits target that branch's `Justfile`/`main.yml`/`conftest.py`), not `main`.
- Deliberately did NOT swap `deploy_app`'s `validate_*` calls to the `bundle.py` versions: that changes deploy-path behavior this repo's offline suite can't test (only vetiver's live test exercises it). Deferred to a separately-tested change.


tests/test_main_system_caches.py
```diff
@@ -1,159 +1,163 @@
...
+        runner = CliRunner()
         args = ["system", "caches", "delete", "--language", "Python", "--version", "1.2.3", "--image-name", "Local"]
-        apply_common_args(args, server=CONNECT_SERVER, key=api_key)
-
+        apply_common_args(args, server=CONNECT_SERVER, key=API_KEY)
         result = runner.invoke(cli, args)
-        self.assertEqual(result.exit_code, 1)
 
-        self.assertRegex(result.output, "You don't have permission to perform this operation.")
+        self.assertEqual(result.exit_code, 0, result.output)
```
karawoo, line 113: maybe worth checking that we're sending the right query string

```suggestion
        self.assertEqual(result.exit_code, 0, result.output)
        delete_request = next(r for r in httpretty.latest_requests() if r.method == "DELETE")
        body = json.loads(delete_request.body)
        self.assertEqual(body["language"], "Python")
        self.assertEqual(body["version"], "1.2.3")
        self.assertEqual(body["image_name"], "Local")
```

karawoo (approved):

PR #813, opened 2026-07-02
Remove orphaned integration folders

edavidaja:
## What

Removes three top-level folders with defunct tests, and adds a small guard test for the bootstrap JWT constants.

- **Delete `integration-testing/`** — a `docker-compose` harness that built bats/cypress suites which were removed long ago. CI now runs the pytest suite via the `posit-dev/with-connect` action (#739); nothing invokes this compose setup.
- **Delete `integration/jwt_testbed.py`** — a manual, non-CI script that POSTed to a live server's bootstrap endpoint to check authorization responses. That is server-side behavior outside this CLI's scope. The client-side JWT signing this tool *is* responsible for remains covered by `tests/test_json_web_token.py`.
- **Drop the now-dead `"integration"` entry** from ruff's `extend-exclude` in `pyproject.toml`.
- **Add `TestBootstrapContract`** to `tests/test_json_web_token.py`, pinning `DEFAULT_ISSUER` / `DEFAULT_AUDIENCE` / `BOOTSTRAP_SCOPE` to their literal values so a change to these wire-contract constants can't slip through unnoticed.

## Why

Neither folder is referenced by CI, the Justfile, docs, or scripts. They're dead scaffolding; keeping them invites confusion about what actually runs.

## Verification

- `just lint` — clean.
- Full pytest suite runs with no collection errors referencing the removed folders. (Pre-existing `test_quickstart.py` failures are an unrelated local `orjson`/maturin build issue.)


nealrichardson (approved):
Love it

PR #815, opened 2026-07-06
Fix git metadata detection for deploy with bare manifest filename

### LINKED ISSUE #770: bug: deploy with manifest doesn't pull in git metadata
See #736. Not sure why but I observed that `rsconnect deploy manifest` doesn't add the git metadata, maybe if the manifest exists we short-circuit early and don't assemble the metadata?

nealrichardson:
## Intent

Resolves #770.

`rsconnect deploy manifest manifest.json` (and other deploy commands given a bare relative filename) failed to attach git metadata to the bundle upload, even inside a git checkout.

## Approach

The deploy commands derive the directory to scan for git metadata with `dirname(file)`. For a bare filename like `manifest.json`, `os.path.dirname("manifest.json")` returns `""`. That empty string was passed straight to `subprocess.run(..., cwd="")`, which raises `FileNotFoundError`; `_run_git_command` catches it and returns `None`, so `is_git_repo("")` returned `False` and metadata detection was silently skipped — the "short-circuit" described in the issue.

The fix normalizes an empty `cwd` to `"."` at the single choke point (`_run_git_command`), so it covers every deploy command that derives `base_dir` via `dirname(...)`, not just `deploy manifest`.

## Type of Change

- [x] Bug Fix
- [ ] New Feature
- [ ] Breaking Change

## Automated Tests

Added two tests in `tests/test_git_metadata.py` covering the empty-directory-uses-cwd behavior (`is_git_repo("")` and `detect_git_metadata("")` from within a git repo). All tests in the file pass; ruff format/check clean.

## Checklist

- [x] I have updated CHANGELOG.md to cover notable changes.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Fixed a bug where `rsconnect deploy manifest manifest.json` (and other deploy
+  commands given a bare filename) failed to detect git metadata. The directory
+  passed to git detection was empty in that case, which caused detection to be
+  silently skipped; it now falls back to the current directory.
 - The `CONNECT_SERVER_VERSION` environment variable can now be set to tell
   rsconnect-python which Connect version to assume for feature-availability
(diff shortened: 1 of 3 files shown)


lucasrod16 (approved):
LGTM

PR #818, opened 2026-07-13
Add `rsconnect server set-default` and a `server` command group

### LINKED ISSUE #812: enable setting existing saved server as default
currently you can only do it via add.

edavidaja:
## What

Adds `rsconnect server set-default -n <name> | -s <url>` to mark an already-saved server as the default without re-adding it, and introduces a `rsconnect server` command group as the home for server management.

## The `server` command group

A single Click `Command` object can belong to multiple groups, so the group **reuses the existing** `add`/`list`/`remove`/`details`/`bootstrap` command objects — no duplication; the top-level commands are unchanged and keep working. `set-default` is new and registered only under the group.

- `rsconnect server add|list|remove|details|bootstrap` — aliases of the existing top-level commands
- `rsconnect server set-default` — new; the only entry point for this command

`login`/`logout` are intentionally **not** in the group — they stay top-level (common publisher commands). `bootstrap` (rare server-provisioning) is grouped.

## Behavior notes

- **`set-default` is local only** — no server contact/re-validation; use `rsconnect details -n <name>` to re-check.
- `-s` matches the stored, normalized URL exactly (like `remove -s`); prefer `-n`. Reports the resolved nickname.
- Exactly one of `-n`/`-s` required.

## Tests & docs

Coverage in `tests/test_main.py` for set-default (name/URL, errors, both/neither, idempotency, shinyapps.io, no top-level `set-default`), group help, and the alias wiring. New `docs/commands/set-default.md` + CHANGELOG entry. Suite green, lint clean.

### DIFF
--- a/rsconnect/main.py
+++ b/rsconnect/main.py
@@ -995,9 +995,6 @@ def set_default(
 def server_group():
     """
     Manage saved Posit Connect and shinyapps.io server nicknames.
-
-    These subcommands are aliases for the top-level add, list, remove, and
-    details commands, grouped under a single namespace.
     """
 
 
@@ -1006,6 +1003,7 @@ server_group.add_command(list_servers)
 server_group.add_command(remove)
 server_group.add_command(details)
 server_group.add_command(set_default)
+server_group.add_command(bootstrap)
 
 
 def _resolve_identity_token(identity_token: Optional[str], identity_token_file: Optional[str]) -> Optional[str]:
--- a/rsconnect/main.py
+++ b/rsconnect/main.py
@@ -994,7 +994,7 @@ def set_default(
 @cli.group("server", no_args_is_help=True)
 def server_group():
     """
-    Manage saved Posit Connect and shinyapps.io server nicknames.
+    Manage connections with Posit Connect and shinyapps.io servers.
     """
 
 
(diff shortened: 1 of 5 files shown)


rsconnect/main.py
```diff
@@ -947,7 +947,65 @@ def remove(
...
+    click.echo('Server "%s" is now the default.' % name)
+
+
+@cli.group("server", no_args_is_help=True)
+def server_group():
+    """
+    Manage saved Posit Connect and shinyapps.io server nicknames.
+
+    These subcommands are aliases for the top-level add, list, remove, and
+    details commands, grouped under a single namespace.
```
karawoo: I feel like this implies that `set-default` also exists as a top-level command. There's nothing in the docstring that suggests `set-default` is different from the rest of the server commands.

karawoo (approved):
nice, I like the addition of the `server` group

rsconnect/main.py
```diff
@@ -947,7 +947,63 @@ def remove(
...
+
+        server_store.set_default(name)
+
+    click.echo('Server "%s" is now the default.' % name)
+
+
+@cli.group("server", no_args_is_help=True)
+def server_group():
+    """
+    Manage saved Posit Connect and shinyapps.io server nicknames.
```
edavidaja:
```suggestion
    Manage connections with Posit Connect and shinyapps.io servers.
```

PR #821, opened 2026-07-16
improve release documentation

amol-:
## Intent

Make release documentation as repeatable as possible and ideally copy-pastable

## Type of Change

- [x] Documentation

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -88,6 +88,13 @@ jobs:
           echo "::error::tag '$tag' does not match pyproject version '$ver'"
           exit 1
         fi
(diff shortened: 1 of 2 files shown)


edavidaja (approved):
Could we add a couple of just targets for the release workflow? That might make it easier to do the right thing.

amol-:
> Could we add a couple of just targets for the release workflow? That might make it easier to do the right thing.

Agree, that would be the right thing to do.
But before turning those into automation it made sense to confirm this flow works as expected by having someone else than me execute it. If we all agree this flow is the right one, then we can easily move the snippets into just scripts

Issue #829, opened 2026-07-23
Support reading/writing Publisher `.posit/` deployment records (in addition to `rsconnect-python/*.json`)


mconflitti-pbc:
## Summary

`rsconnect-python` and the Posit Publisher (VS Code extension) currently track deployments in two mutually exclusive, incompatible formats. A directory deployed with one tool is invisible to the other, so the two toolchains can't collaborate on the same content. I'd like `rsconnect-python` to be able to read (and eventually write) the Publisher's `.posit/` records so all of our deployment tooling speaks the same language.

## Current behavior

Deploying the same app two ways produces two independent record sets that neither tool reconciles:

**Publisher** (`.posit/publish/`) — TOML, with a public versioned schema:
- `.posit/publish/<config>.toml` — configuration (`posit-publishing-schema-v3.json`)
- `.posit/publish/deployments/<deployment>.toml` — deployment record (`posit-publishing-record-schema-v3.json`), including `server_url`, content `id`/GUID, `bundle_id`, `dashboard_url`, `direct_url`, deployed files, requirements, etc.

**rsconnect-python** (`rsconnect-python/<name>.json`) — JSON keyed by server URL, with `app_id`, `app_guid`, `app_url`, `app_mode`, `app_store_version`.

Because rsconnect-python only knows about its own `rsconnect-python/*.json` store, a re-deploy via the CLI won't find or update the Publisher's record (and creates a *separate* piece of content rather than updating the existing one).

## Requested behavior

- On deploy/redeploy, have `rsconnect-python` **detect an existing `.posit/` directory** and use those records to resolve the target server + existing content (GUID/id) for update-in-place.
- Treat this as an **addition, not a replacement** of the existing `rsconnect-python/*.json` behavior initially, so we can begin transitioning without breaking existing workflows.
- The Publisher schema is public and versioned, which makes it a good shared source of truth:
  - Config: `https://cdn.posit.co/publisher/schemas/posit-publishing-schema-v3.json`
  - Record: `https://cdn.posit.co/publisher/schemas/posit-publishing-record-schema-v3.json`

## Why

The Publisher approach is considered the better/preferred way of recording deployments, and aligning `rsconnect-python` on it means the CLI, the VS Code extension, and other tooling share one deployment source of truth instead of drifting into incompatible records for the same content.

## Related issues / prior art

| # | Repo · Issue | State | Why it's relevant |
|---|---|---|---|
[...]

PR #830, opened 2026-07-27
Posit Publisher .posit/publish: TOML interop, redeploy, and config-driven bundling

mconflitti-pbc:
## Summary

Adds interoperability with **Posit Publisher's `.posit/publish` project files**, so content can be published with either rsconnect-python or the Publisher VS Code extension. Addresses #829.

rsconnect-python now reads and writes Publisher's TOML **configuration** (`.posit/publish/<name>.toml`) and **deployment record** (`.posit/publish/deployments/<name>.toml`) alongside the existing legacy `rsconnect-python/*.json` store, and gains a `redeploy` command that redeploys straight from a `.posit` project.

Bundling becomes **config-driven only when a configuration curates a subset**: if a `.posit` config's `files` list names real content patterns, exactly those files are bundled (Publisher's `.gitignore`-syntax include/exclude semantics — a matching pattern includes, `!` excludes). Otherwise — no config, or a config declaring `files = ["*"]` — file selection is **unchanged from today**: the whole-tree walk with only the built-in ignore list. A config's `integration_requests` are propagated into the generated `manifest.json` exactly as Publisher emits them, so integrations authored in Publisher are honored on deploy.

### 1. Deploy as usual — now also writes `.posit`
```console
$ rsconnect deploy shiny ./my-app -n prod
# ... deploy runs as before, and now also creates:
#   my-app/.posit/publish/my-app-4F2A.toml                     # configuration (type, entrypoint, files, python)
#   my-app/.posit/publish/deployments/deployment-9C1B.toml     # record: server_url, content id/GUID, bundle_id, dashboard/direct/logs URLs
```

### 2. Zero-arg redeploy from an existing `.posit` project
```console
$ cd my-app
$ rsconnect redeploy
# reads .posit, recovers the target server + content GUID from the record,
# and redeploys in place — no framework, entrypoint, server, or app-id needed
```

### 3. Redeploy a project by path
```console
$ rsconnect redeploy ./my-app
```

### 4. Interop: redeploy a project that was published from the Publisher VS Code extension
```console
$ git clone git@github.com:me/my-app.git && cd my-app   # repo already contains .posit/publish/ authored by Publisher
$ rsconnect login https://connect.example.com           # one-time: authenticate the CLI to that server (OAuth)
$ rsconnect redeploy                                     # matches the record's server_url to that saved login and redeploys, reusing the GUID
```

### 5. Credentials are matched by the record's server URL
```console
# If you've already `rsconnect add`/`login`-ed the server the record points at,
# redeploy finds it automatically — no --server / --api-key required:
$ rsconnect redeploy
# Ambiguous (two saved servers for the same URL)? It asks you to choose:
#   Error: Multiple saved servers match https://connect.example.com (prod-a, prod-b); pick one with --name.
$ rsconnect redeploy --name prod-a
```

### 6. Disambiguate when a project has multiple configs or servers
```console
$ rsconnect redeploy --config-name my-app                # multiple .posit configs in the project
$ rsconnect redeploy --server https://connect.example.com # config deployed to more than one server
```

### 7. First-ever deploy from a Publisher config (config exists, no record yet)
```console
$ rsconnect redeploy --server prod        # or --name prod; a destination is required the first time
# On success a deployment record is written, so subsequent `rsconnect redeploy` needs no args.
```

### 8. `write-manifest` now also emits a `.posit` config
```console
$ rsconnect write-manifest shiny ./my-app
# writes my-app/manifest.json AND my-app/.posit/publish/my-app-7Q3D.toml
```

### 9. Redeploy pre-`.posit` content (manifest + legacy JSON fallback)
```console
$ rsconnect redeploy ./legacy-app
# legacy-app has manifest.json + rsconnect-python/*.json but no .posit:
# deploys the manifest bundle to the recorded server/GUID, then writes .posit for next time
```

### 10. Config-driven bundling — only the config's files are bundled
```console
# .posit/publish/my-app.toml declares:  files = ["/app.py", "/requirements.txt", "data/"]
$ rsconnect redeploy ./my-app
# bundles exactly app.py, requirements.txt, and everything under data/ —
# nothing else in the tree. Without a curated files list, bundling is unchanged:
# the whole tree except the built-in ignore list.
```

A config rsconnect-python writes itself records `files = ["*"]`, *not* a snapshot of
the files that happened to deploy — so a project that never had `.posit` keeps
bundling identically on every subsequent deploy, and newly added modules or freshly
rendered output are still picked up.

## What's included

- **New `rsconnect/publisher/` package** porting Publisher's format (`posit-dev/publisher`):
  - `schema.py` — schema URLs, the `app_mode ↔ content-type` map (from `bundler/appMode.ts`), `.posit` path helpers.
  - `serialize.py` — `tomli_w` writer matching Publisher (quoted `$schema` first, multiline arrays, tables last); reader via stdlib `tomllib` / `toml` backport.
  - `config.py` / `record.py` — dataclasses with read‑merge‑write that preserves fields rsconnect doesn't own (incl. first-class `connect_cloud`, and an existing `[r]` table); recovers `files`/`requirements`/entrypoint from the built bundle manifest, resolving a bare Python module reference (e.g. `app` for `app.py`) back to the literal file path Publisher's schema expects, so long as the manifest confirms that file exists.
  - `store.py` — dual-write on deploy + `resolve_publisher_deploy_target` for reads; **Publisher-compatible random file naming** (`<title>-<CODE>` / `deployment-<CODE>`, 4-char base‑32) with find-or-reuse so redeploys update in place, disambiguated by content GUID when more than one deployment record shares a server; normalizes `server_url` the same way Publisher's own writers do (lowercase scheme+host, no trailing slash, no `/__api__` suffix, no default port, no duplicate slashes), so a project deployed first with rsconnect-python doesn't fail a later Publisher redeploy on a byte-for-byte URL mismatch.
- **Dual-write on Connect/SPCS deploys** via `RSConnectExecutor.save_deployed_info` (best-effort; never fails the deploy).
- **`rsconnect redeploy [PATH]`** with the behaviors shown above, dispatching to every content type rsconnect-python can build a bundle for: Shiny, FastAPI, Flask/API, Streamlit, Dash, Bokeh, Gradio, Panel, HTML, Node.js, Jupyter notebooks (incl. Voila), and Quarto.
- **`write-manifest`** now also emits a `.posit` config.
- **Config-driven file selection** — new `rsconnect/publisher/files.py` (a `pathspec`/gitwildmatch engine reproducing Publisher's `bundler/collect.ts`: `select_config_files` allowlist + `STANDARD_EXCLUSIONS`). Wired through `bundle.create_file_list` via a `restrict_to_files` context set by the executor, so **all** deploy commands honor it. `store.resolve_bundle_files` returns the config's selection (force-including the entrypoint), or `None` — impose no restriction, leaving today's whole-tree walk untouched — when no config applies **or** when the config declares no real restriction (`files` absent, empty, or `["*"]`, discounting the `.posit` paths Publisher adds so they ship). Configs rsconnect mints record `files = ["*"]`; the concrete deployed set lives on the deployment record, where it can't be mistaken for curation.
- **`integration_requests` → `manifest.json`** — `bundle.overlay_manifest` merges config-authored manifest fields rsconnect can't derive (currently `integration_requests`) in `Manifest.__init__`; `store.config_manifest_overlay`/`resolve_manifest_overlay` map them to Publisher's manifest shape.
- Shared `_plan_deploy_bundle` helper extracted from `deploy pyproject` (reused by `redeploy`, including its widened content-type coverage above).
- Adds `tomli-w` and `pathspec` dependencies.

## Scope notes

- **R content types are out of scope, and not a phased rollout** — rsconnect-python has no R bundle builder at all (R deploys go through the R `rsconnect` package), so `.posit` interop can't originate R content either, and `redeploy` never dispatches to r-shiny/r-plumber/rmd/rmd-shiny. The one R-related change here: rewriting a config now preserves an existing `[r]` table (matching how `python`/`quarto` are preserved), so rsconnect-python never destroys Publisher-authored R metadata it doesn't understand.
- **Connect Cloud** (`connect.posit.cloud`) `.posit` files are read/preserved for interop, but deploying to Connect Cloud is **not** supported by this tool (no client) — only Posit Connect and Snowflake/SPCS write `.posit`. shinyapps.io / Posit Cloud stay legacy-JSON-only (as Publisher also writes no `.posit` for them).
- **No change to file selection without a hand-curated `files` list.** Two revisions of this PR got this wrong and were fixed:
  - Config-less deploys briefly honored `.gitignore`. Reverted — `.gitignore` is the wrong signal for bundling: a Quarto project's rendered HTML is routinely gitignored precisely because it shouldn't be committed, yet it's exactly what needs to deploy. Narrowing the default file set is a larger behavior change than `.posit` interop and belongs in its own PR.
  - A config minted by deploy #1 recorded the concrete deployed file set, which deploy #2 then read back as curation — silently dropping modules added in between. Now it records `files = ["*"]`, and an unrestricted list imposes no restriction.
  Verified by capturing the tarball member list for a first deploy of a plain project on this branch and on `main`: byte-identical. A second deploy adds only the two `.posit` files.
- **Out of scope (follow-up #831):** a config's Connect-side settings that Publisher applies via the Connect API rather than the manifest — `description`, `connect.runtime.*`, `connect.access.*`, `connect.kubernetes.*`, and `environment`/`secrets` — are preserved on disk but never read for behavior; `redeploy` only ever carries forward `entrypoint`, `requirements`, `title`, and the content id, and doesn't issue the Connect API calls (PATCH content, env vars) that would apply the rest — whether or not they were just edited in the TOML. This is a **no-op, not a clobber**: the live Connect deployment is left exactly as it was, never reset to a default. `redeploy` now prints a warning when a resolved configuration sets any of these, so an edit isn't silently swallowed with no feedback. Full application is tracked in #831.
- The `uv.lock` diff is large but only adds `tomli-w`/`pathspec` (a reserialization; nothing removed or version-bumped).

## Testing

- `tests/test_publisher.py` — type-map coverage (incl. TensorFlow → `unknown`), round-trips, idempotent redeploy, URL normalization, a Publisher-authored interop fixture, `connect_cloud` round-trip, naming methodology, config/record pinning, that a written config's `files` reference the record actually written, `bundle_url` round-tripping (including surviving a redeploy that doesn't set it), the settings-warning helper (`unhonored_redeploy_settings`), and that two apps in one project deployed to the same server get distinct, non-colliding records.
- `tests/test_redeploy.py` — command registration, identity reuse, config-name selection, legacy fallback, credential matching by URL (incl. ambiguous → error), and tarball-level assertions that `redeploy` bundles exactly the config's files, matches `deploy`'s bundle, stays stable across repeated redeploys, and — for a project with no `.posit` — that a second deploy still bundles a later-added module and gitignored rendered output.
- `tests/test_publisher_files.py` — config allowlist matching (rooted vs. basename, `!` exclusion, dir includes, `**`, `STANDARD_EXCLUSIONS`, venv/renv skipping), the `create_file_list` restriction + builder-exclude interaction, `resolve_bundle_files` returning `None` without a config (and a gitignored build artifact still bundling) and for an unrestricted config (absent / `["*"]` / `["*"]`+`.posit` paths), restriction for a genuinely curated config, entrypoint force-include, and `integration_requests` propagation into the manifest (incl. an end-to-end executor test).
- `tests/test_deploy_pyproject.py` — updated so the "unsupported app_mode" coverage still exercises a genuinely unhandled mode (R Shiny) now that Dash/Bokeh/Gradio/Panel/HTML/Node.js are supported, using a synthetic aliased-name case to keep covering that an unsupported alias fails quoting the configured string, not its resolved target.
- Full suite green (843 passed, 12 skipped); `ruff format`/`ruff check` clean.

## Try it

```console
uvx --from "git+https://github.com/posit-dev/rsconnect-python.git@feat/publisher-toml-support" rsconnect redeploy --help
```


mconflitti-pbc:
Follow-up work is tracked in #831: applying a `.posit/publish` config's **Connect-side settings** (`description`, `connect.runtime.*`, `connect.access.*`, `connect.kubernetes.*`, `environment`, `secrets`) that Connect applies via API rather than through `manifest.json`. This PR already propagates the manifest-side fields (including `integration_requests`); #831 covers a post-deploy hook plus an ad-hoc `rsconnect sync` command to push config changes to a chosen (or config/record-default) server.

nealrichardson:
> Behavior change: deploys without a .posit config now honor .gitignore when selecting files (previously only the built-in ignore list was applied). Called out in the CHANGELOG.

I'm not sure that's what we want, the reason IIRC that there is a separate tracking of included/excluded files is that for rendered content, you might gitignore the build output (the html that a quarto project produces, for example) because you don't want to check it in, but might want to deploy exactly that. That's at least why .gitignore wasn't just used before. 

It's possible that that's not the right tradeoff, and that there's a better way to solve for that. But I wouldn't just casually do that here, I think that's a more significant behavior change.

mconflitti-pbc:
> > Behavior change: deploys without a .posit config now honor .gitignore when selecting files (previously only the built-in ignore list was applied). Called out in the CHANGELOG.
> 
> I'm not sure that's what we want, the reason IIRC that there is a separate tracking of included/excluded files is that for rendered content, you might gitignore the build output (the html that a quarto project produces, for example) because you don't want to check it in, but might want to deploy exactly that. That's at least why .gitignore wasn't just used before.
> 
> It's possible that that's not the right tradeoff, and that there's a better way to solve for that. But I wouldn't just casually do that here, I think that's a more significant behavior change.

Yeah that is fair! Will revert that

amol-:
I'm a bit confused about the interactions with `deploy manifest`, `deploy pyproject` and `redeploy`,
they more or less seem to target reproducible deploys and it's not immediately clear when a user should use one or the other.

nealrichardson:
Why do we need `redeploy` at all? Why isn't `deploy` sufficient?

mconflitti-pbc:
> Why do we need `redeploy` at all? Why isn't `deploy` sufficient?

to avoid breaking changes/behavior with existing deploy commands. Could try to trigger redeploy behavior with empty deploy command but figured we could avoid that.

mconflitti-pbc:
> I'm a bit confused about the interactions with `deploy manifest`, `deploy pyproject` and `redeploy`, they more or less seem to target reproducible deploys and it's not immediately clear when a user should use one or the other.

That is fair. Was punting on UX improvements of existing commands if I can help it. If you have ideas, I am open to them!

amol-:
> Why do we need `redeploy` at all? Why isn't `deploy` sufficient?

I think because when users say _"I want to deploy"_, 
rsconnect answers: _"great! And how do you want to deloy?"_
1. via the automatic configuration (IE: `rsconnect deploy fastapi`)
2. via the configuration provided in the `manifest.json` (`rsconnect deploy manifest`)
3. via the configuration provided in the `pyproject.toml` (`rsconnect deploy pyproject`)

and now there would be

4. via the configuration provided in `.posit` directory (`redeploy`/`deploy publisher`?)

that's why I was wondering how we can ease reasoning about this for users.

Especially seems it looks like that this PR would always introduce the `.posit` directory, and thus from that point onward you would have two different commands you could use to redeploy

mconflitti-pbc:
@nealrichardson i reverted the gitignore change.

@amol- re: the new command

adding a parallel path allows this to be added and used by our agentic deployment skill alongside projects that have been deployed using publisher or vice versa. the goal of this pr was not to change the fundamental structure of the deploy command. 

definitelty open to folding that behavior into the deploy command in a follow up.

there are other things to address as well regarding how we want to rsconnect-python to handle settings which are tracked in #831

docs/CHANGELOG.md
```diff
@@ -7,14 +7,56 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-- `rsconnect deploy` subcommands now accept `--quiet`, which suppresses the
-  step-by-step progress lines and the streamed server build log, printing only
-  the deployed content URL to stdout so it can be captured with
-  `URL=$(rsconnect deploy ... --quiet)`. Errors still go to stderr, and on a
-  failed deploy the server task log is emitted to stderr so failures remain
-  diagnosable. `--quiet` cannot be combined with `-v/--verbose`, and for
-  shinyapps.io deploys it also skips opening a browser.
+### Added
+
+- Python 3.14 support. The test suite now runs on Python 3.14 in CI.
```
nealrichardson, line 12: (I know this isn't from your PR but it's in the diff) but is it true that we "added" support for Python 3.14? Or did we just add CI? If the latter, then we should delete this line.

docs/CHANGELOG.md
```diff
@@ -7,14 +7,56 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  failed deploy, the server task log also goes to stderr. `--quiet` cannot
+  combine with `-v`/`--verbose`. For shinyapps.io deploys, `--quiet` also
+  skips opening a browser.
+- Interoperability with Posit Publisher `.posit/publish` project files. On
+  deploys to Posit Connect or Snowflake (SPCS), rsconnect-python writes a
+  Publisher configuration (`.posit/publish/<name>.toml`) and a deployment
+  record (`.posit/publish/deployments/<name>.toml`) next to the existing
+  `rsconnect-python/` metadata, and reads and preserves configurations and
+  records that Publisher wrote. One project can publish with either tool.
+- `rsconnect redeploy [PATH]` command. It redeploys content from an existing
```
nealrichardson: To my earlier question: one thing that sounds odd to me about adding a `redeploy` command is that we already can re-deploy. rsconnect-python records somewhere where it has previously deployed, so `rsconnect deploy fastapi .` for example will use the same content guid as the last time. With this change, does it mean that `rsconnect deploy fastapi .` and `rsconnect redeploy .` will both re-deploy but go through distinct code paths?

Why instead not have just `rsconnect deploy .`, and (in the case here where there is no pseudo-appmode given) it first looks for the TOML file, then falls back to looking for the old rsconnect-python record, etc.? Since the appmode is immutable on Connect, you don't need to specify it when you're re-deploying content, it must be whatever it already is on Connect so you can query it from the API. (That's what connect-actions does.) So even without a publisher TOML you can reconstruct enough to deploy with just the content GUID (from the rsconnect-python record, or as an argument).

docs/CHANGELOG.md
```diff
@@ -7,14 +7,56 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  from the deployment record, so you do not need to give a framework,
+  entrypoint, or server. `PATH` defaults to the current directory. It also
+  works from a legacy `manifest.json` and `rsconnect-python/` deployment
+  record, and writes `.posit` files for future use. `redeploy` supports
+  every content type rsconnect-python can bundle except R, since
+  rsconnect-python has no R bundler. It carries forward only the
+  entrypoint, requirements, title, and content ID from a configuration; if
+  the configuration sets `description`, `environment`, `secrets`, or
+  `connect.*`, it prints a warning and deploys anyway, since applying those
+  settings needs a Connect API call `redeploy` does not yet make.
```
nealrichardson: "yet" implies that it will, and I'm not sure it should.

docs/CHANGELOG.md
```diff
@@ -7,14 +7,56 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `.posit/publish` output from `rsconnect write-manifest` commands. Each
+  command now writes a configuration next to the generated `manifest.json`.
+- Read support for Connect Cloud (`connect.posit.cloud`) `.posit` files.
+  rsconnect-python reads and preserves these files for interoperability, but
+  does not support deploying to Connect Cloud.
+- File curation from a `.posit/publish` configuration's `files` list. When
```
nealrichardson: These bullets sound more like documentation about how it works, not separate changelog entries.

rsconnect/publisher/config.py
```diff
@@ -0,0 +1,175 @@
+"""The ``.posit/publish/<name>.toml`` configuration file (the "what").
```
nealrichardson, line 1: what "the 'what'"?

dotNomad (commented):
Looking at how this is setup I think it could cause some confusion.

My read is that `deploy` will not use `.posit/publish` files to determine what or how to deploy, but will write the `.posit/publish` files.

The `redeploy` command will use and write the `.posit/publish` files.

But there is a pretty large difference in how re-deployment and `--new` work. `rsconnect` will, when targeting the same server, will forget about previous deployments - only storing one content item per-server. That will cause the `.posit/publish` files to have a single configuration and two deployments - that is totally fine, Publisher handles that and it is fully expected however `rsconnect` doesn't play nicely there.

`rsconnect deploy` will use the new deployment and `rsconnect redeploy` will error because it doesn't know which deployment of the two to use. If you remove one of the deployment records you could have `rsconnect deploy` going to the new content item and `rsconnect redeploy` going to the old which is very confusing.

---

I think it would be best for us to sync about this and come up with a minimal path forward to get the behavior, but not try to re-design `rsconnect` too much.

My preference would be to use `.posit/publish` configuration and deployment files, but mimic the `rsconnect` behavior of only having one deployment per server to start. It means that you cannot easily have two content items on a server like you can with Publisher, but it avoids some real confusing cases like the one above and reduces the need, initially at least, of the `redeploy` command.

amol-:
> 
> `rsconnect deploy` will use the new deployment and `rsconnect redeploy` will error because it doesn't know which deployment of the two to use. If you remove one of the deployment records you could have `rsconnect deploy` going to the new content item and `rsconnect redeploy` going to the old which is very confusing.
> 

What if we made a new command for real? `"reconnect publish"`. That relies on the new `.posit/publish` files for both first and subsequent deploys, has a predictable behavior fully separate from deploy and avoids polluting the deploy command behavior itself which causes confusion 

A parallel deploy path we can describe as the one for people using both reconnect and the publisher to deploy the same content.

nealrichardson:
> What if we made a new command for real?

If we were to do that, we're basically making a v2 deploy CLI, and I would advocate for doing that in `posit-cli`. Most of the code could still live here, if that were easier, but if we're going to teach a new command for deployment, then we should be definitely be moving off of calling it `rsconnect`.

I'm not trying to put extra requirements on here to stop forward progress, but if that's the direction we're headed, let's just do it.

mconflitti-pbc:
> If we were to do that, we're basically making a v2 deploy CLI, and I would advocate for doing that in `posit-cli`. Most of the code could still live here, if that were easier, but if we're going to teach a new command for deployment, then we should be definitely be moving off of calling it `rsconnect`.
> 
> I'm not trying to put extra requirements on here to stop forward progress, but if that's the direction we're headed, let's just do it.

so are you saying we implement the command here fully but only document it in posit-cli? Or have some sort of private execution path living here that only posit-cli exposes as usable?

Decoupling the deployment logic and/or replicating a lot of it into posit-cli as a lift and shift would be much bigger scope than what I was hoping for.

mconflitti-pbc:
A hybrid approach seems like a good fit:

- `rsconnect-python` owns the reusable `.posit/publish` parsing, bundling, record selection/writing, and deployment service, but does not add `rsconnect redeploy` or change normal `rsconnect deploy` behavior.
- `posit-cli` owns the new UX as a thin wrapper around that service: `posit connect publish`.

```console
# First publish: config exists, destination is explicit
posit connect publish . --server https://connect.example.com

# Subsequent publish: destination comes from the deployment record
posit connect publish .

# Disambiguate multiple configs or deployments
posit connect publish . --config sales-dashboard
posit connect publish . --deployment connect-production
```

This keeps the deployment engine and Publisher interoperability code here, while making the new v2-style command part of `posit-cli` rather than expanding the `rsconnect` CLI.

nealrichardson:
> A hybrid approach seems like a good fit

Yeah that's more what I was suggesting. I don't know exactly where everything would go and how exactly it would work, but more like that, where the net-new CLI command is in `posit-cli` (and we could take the opportunity there to clean up anything else in the UX of `rsconnect deploy` that we wanted to change).

mconflitti-pbc:
> > A hybrid approach seems like a good fit
> 
> Yeah that's more what I was suggesting. I don't know exactly where everything would go and how exactly it would work, but more like that, where the net-new CLI command is in `posit-cli` (and we could take the opportunity there to clean up anything else in the UX of `rsconnect deploy` that we wanted to change).

I will wire it up and see how it feels!

PR #836, opened 2026-08-04
fix: preserve title_is_default in deploy manifest/bundle redeploys

### LINKED ISSUE #835: `deploy manifest`/`deploy bundle` always resolve a default title, triggering an unwanted PATCH of the content title (breaks trusted publishing)
`deploy manifest` and `deploy bundle` pre-resolve the title before constructing `RSConnectExecutor`:

- [`main.py:1901`](https://github.com/posit-dev/rsconnect-python/blob/97cd7900f20d5f925db3103808000ca37f877773/rsconnect/main.py#L1901): `title = title or default_title_from_manifest(file)`
- [`main.py:1997`](https://github.com/posit-dev/rsconnect-python/blob/97cd7900f20d5f925db3103808000ca37f877773/rsconnect/main.py#L1997): `title = title or default_title_from_bundle(file)`

Because the executor computes [`title_is_default = not title`](https://github.com/posit-dev/rsconnect-python/blob/97cd7900f20d5f925db3103808000ca37f877773/rsconnect/api.py#L1296), these commands always hand `deploy()` a non-empty title with `title_is_default=False`. `RSConnectClient.deploy` then hits [`api.py:1145`](https://github.com/posit-dev/rsconnect-python/blob/97cd7900f20d5f925db3103808000ca37f877773/rsconnect/api.py#L1145):

```python
if app["title"] != app_title and not title_is_default:
    result = self.content_update(app_guid, {"title": app_title})  # PATCH v1/content/{guid}
```

So every redeploy to an existing `--app-id` where the derived default differs from the content's real title issues a `PATCH /v1/content/{guid}`. The derived default is usually junk: for a manifest with no `metadata.entrypoint` (e.g. anything from R's `rsconnect::writeManifest()`), `_default_title_from_manifest` falls back to `dirname(manifest_file)` — deploying `manifest.json` by relative path yields the cwd's basename (in our case, the title `"inst"`).

**Consequences:**

1. **Trusted publishing is broken for manifest deploys.** Connect machine identities are only permitted on the content-get/bundle/deploy/task endpoints; `PATCH /v1/content/:guid` is rejected with `You don't have permission to perform this operation`, failing the whole deploy at the "Deploying bundle" step. Since R content can only be deployed via manifest, every R + trusted-publishing deploy (e.g. through posit-dev/connect-actions) fails.
2. **With an API key, content is silently renamed** to the derived default (our app would have been renamed to "inst").

**Repro:** create content on Connect with a title, then `rsconnect deploy manifest --app-id <guid> manifest.json` — observe the PATCH (403 under a trusted-publishing key; silent rename under an API key).

**Suggested fix:** don't pre-resolve the title in the command functions. Pass the user's `--title` (possibly `None`) into the executor so `title_is_default` stays accurate, and use the manifest/bundle-derived default only where a title is actually needed (content creation). Other deploy subcommands already behave this way, which is why source deploys work under trusted publishing.

tdstein:
## Summary

`deploy manifest` and `deploy bundle` pre-resolve the content title before constructing `RSConnectExecutor`:

- `rsconnect/main.py` (`deploy_manifest`): `title = title or default_title_from_manifest(file)`
- `rsconnect/main.py` (deploy bundle command): `title = title or default_title_from_bundle(file)`

`RSConnectExecutor.__init__` computes `self.title_is_default = not title`, so these two commands always produced `title_is_default=False`, even when the user never passed `--title`. `RSConnectClient.deploy` then does:

```python
if app["title"] != app_title and not title_is_default:
    result = self.content_update(app_guid, {"title": app_title})  # PATCH v1/content/{guid}
```

Consequences:
- Every redeploy to an existing `--app-id` where the derived default title differs from the server's title issued `PATCH /v1/content/{guid}`.
- Under Connect trusted publishing (machine identities), that endpoint is categorically forbidden, so the whole deploy failed with a 403.
- Under an API key, the content got silently renamed to a junk derived title (e.g. `"inst"` for R manifests with no `metadata.entrypoint`, since `_default_title_from_manifest` falls back to `dirname(manifest_file)` — the cwd basename when deploying `"manifest.json"` by relative path).

Other deploy subcommands (e.g. Python source deploys) were unaffected because they pass `title=None` through to the executor, so `title_is_default=True` and the PATCH is correctly skipped.

## Fix

`RSConnectExecutor.__init__` now accepts an explicit `title_is_default: Optional[bool]` parameter. When provided, it overrides the `not title` inference. `deploy_manifest` and `deploy_bundle` compute `title_is_default = not title` *before* filling in the manifest/bundle-derived default, and pass it through explicitly:

- When the user passes `--title`, behavior is unchanged (title used, `title_is_default=False`).
- When they don't, `title_is_default=True`, so no PATCH is issued on redeploy to an existing `--app-id`.
- New-content creation (no `--app-id`) is unaffected: `RSConnectClient.deploy` forces `title_is_default=False` whenever `app_id is None`, so the manifest/bundle-derived default title is still applied via `content_update` for brand new content, exactly as before.

## Tests

Added to `tests/test_main.py`:
- `test_deploy_manifest_redeploy_preserves_title` — redeploying existing content via `deploy manifest --app-id` without `--title` issues no PATCH, even when the manifest-derived default title differs from the server's title.
- `test_deploy_manifest_redeploy_with_explicit_title_still_updates` — with `--title` explicitly passed, the PATCH still happens when the title differs.
- `test_deploy_bundle_redeploy_preserves_title` — same regression, for `deploy bundle --app-id`.
- `test_deploy_manifest_new_content_uses_manifest_derived_title` — new-content creation (no `--app-id`) still gets its title defaulted from the manifest (`"app5"`, derived from the entrypoint) rather than a generic fallback, and that default is still applied via PATCH after `content_create`.

Verified the first three tests fail on `main` (reproducing the bug) and pass with this fix. Also ran the full `tests/test_main.py` and `tests/test_api.py` suites (134 passed, 12 skipped — the skipped ones require a live Connect server or R), plus `ruff format`/`ruff check` on the touched files.

Added a changelog entry to `docs/CHANGELOG.md` under `## Unreleased`, per this repo's release process (`CONTRIBUTING.md`).

Fixes #835

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Fixed a bug where `rsconnect deploy manifest` and `rsconnect deploy bundle`,
+  when redeploying to an existing `--app-id` without passing `--title`, always
+  issued a `PATCH /v1/content/{guid}` to rename the content to its
+  manifest/bundle-derived default title. Under trusted publishing this request
+  is forbidden outright, failing the whole deploy with a 403; under an API key
+  it silently renamed the content. These commands now only update the title
+  when the user explicitly passes `--title`, matching the other deploy
+  subcommands; new content still defaults its title from the manifest/bundle as
+  before.
 - Added support for Python 3.14. The test suite now runs on Python 3.14 in CI.
 - `rsconnect deploy` subcommands now accept `--quiet`, which suppresses the
   step-by-step progress lines and the streamed server build log, printing only
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,15 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-- Fixed a bug where `rsconnect deploy manifest` and `rsconnect deploy bundle`,
-  when redeploying to an existing `--app-id` without passing `--title`, always
-  issued a `PATCH /v1/content/{guid}` to rename the content to its
-  manifest/bundle-derived default title. Under trusted publishing this request
-  is forbidden outright, failing the whole deploy with a 403; under an API key
-  it silently renamed the content. These commands now only update the title
+- `rsconnect deploy manifest` and `rsconnect deploy bundle`,
+  when redeploying to an existing `--app-id`, only update the title
   when the user explicitly passes `--title`, matching the other deploy
-  subcommands; new content still defaults its title from the manifest/bundle as
-  before.
+  subcommands.
 - Added support for Python 3.14. The test suite now runs on Python 3.14 in CI.
 - `rsconnect deploy` subcommands now accept `--quiet`, which suppresses the
   step-by-step progress lines and the streamed server build log, printing only
--- a/rsconnect/api.py
+++ b/rsconnect/api.py
@@ -1294,11 +1294,11 @@ class RSConnectExecutor:
         self.app_store: AppStore = AppStore(fake_module_file_from_directory(self.path))
         self.app_store_version: int | None = None
         self.api_key_is_required: bool | None = None
-        # Callers that pre-resolve a default title (e.g. `deploy manifest` / `deploy
-        # bundle`, which need the manifest/bundle-derived default rather than the
-        # generic `_default_title(self.path)` fallback above) can pass
-        # `title_is_default` explicitly so this still reflects whether the user
-        # actually supplied `--title`, rather than always being False.
+        # `deploy manifest` / `deploy bundle` pre-resolve `title` to a
+        # manifest/bundle-derived default rather than the generic
+        # `_default_title(self.path)` fallback above, so they pass
+        # `title_is_default` explicitly to report whether the user really
+        # supplied `--title`.
         self.title_is_default: bool = not title if title_is_default is None else title_is_default
         self.deployment_name: str | None = None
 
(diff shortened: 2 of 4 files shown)


nealrichardson:
I feel like supporting `--title` at all when redeploying (when `--app-id` [sic] is provided) is a mistake, but given that it is there, making it skip correctly when no title is provided sounds like a good idea. I'm still a little confused though why this was hitting you in the app you were deploying.

tdstein:
> I'm still a little confused though why this was hitting you in the app you were deploying.

I hit this because the derived title (`inst`) differed from the deployed content (`bio-whatever`), therefore it tries to run the `PATCH v1/content/{guid}` call. The app guid is provided to the connect-actions workflow, so it shouldn't need to update the title.

docs/CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+- Fixed a bug where `rsconnect deploy manifest` and `rsconnect deploy bundle`,
+  when redeploying to an existing `--app-id` without passing `--title`, always
+  issued a `PATCH /v1/content/{guid}` to rename the content to its
+  manifest/bundle-derived default title. Under trusted publishing this request
+  is forbidden outright, failing the whole deploy with a 403; under an API key
+  it silently renamed the content. These commands now only update the title
+  when the user explicitly passes `--title`, matching the other deploy
+  subcommands; new content still defaults its title from the manifest/bundle as
+  before.
```
nealrichardson:
```suggestion
- `rsconnect deploy manifest` and `rsconnect deploy bundle`,
  when redeploying to an existing `--app-id`, only update the title
  when the user explicitly passes `--title`, matching the other deploy
  subcommands.
```

nealrichardson (approved):
I fixed up the tests

PR #838, opened 2026-08-14
feat: add `content migrate-to-connect-cloud`

### LINKED ISSUE #826: support migrating deployment configs from shinyapps.io to connect cloud

samperman:
## Intent
Point a directory's local deployment record at an existing Posit Connect
Cloud content item, so the next deploy from that directory updates that
item instead of creating a second one. Connect Cloud cannot look content
up by name, so the local record is the only way back to a content item;
without one, deploying content that already exists there duplicates it.

Nothing is copied and no bundle is uploaded. The record that was
migrated from is removed, leaving the directory with one deployment
target rather than two; `--from-server` chooses which record to migrate
when the store covers several servers, and `-o/--overwrite` replaces an
existing Connect Cloud record.

Records are keyed by account, and a deploy refuses content owned by
another account, so a record written under the wrong account would be
silently useless. Migration therefore requires the content's owning
account to be the one being targeted and names it when it is not. With
no local record at all the record is reconstructed from the content,
with app mode `unknown`, which `validate_app_mode` already tolerates.

Fixes #826
    
## Type of Change
- [ ] Bug Fix            
- [x] New Feature       
- [ ] Breaking Change   

## Approach
This follows the patterns established by https://github.com/rstudio/rsconnect/pull/1353 and https://github.com/rstudio/rsconnect/pull/1358

## Automated Tests
Unit tests exist, integration tests are scheduled but will happen after this merges.

## Directions for Reviewers

These steps can be used to verify it works locally:
1. Setup credentials
    * `uv run rsconnect add -n shinyapps -A <shinyapps-account> -T <token> -S  <secret>`
    * `uv run rsconnect add -n cloud --connect-cloud -A <cc-account>` # opens a browser
2. deploy an app to shinyapps.io
    * `uv run rsconnect deploy shiny ./my-app -n shinyapps`
3. Migrate the content in https://connect.posit.cloud
    * remember the content id
4. Migrate the deployment record
    * `uv run rsconnect content migrate-to-connect-cloud ./my-app -n cloud  --content-id <content-id>`
5. Verify it, deploy it again
    * `uv run rsconnect info ./my-app`
    * `uv run rsconnect deploy shiny ./my-app -n cloud`

This tool will also create a new deployment record if its missing. This test can be done without shinyapps.io or content migration in Connect Cloud.

1. Setup credentials as above
2. Deploy content to Connect Cloud
    * `uv run rsconnect deploy shiny ./my-app -n cloud`
3. Delete the deployment record
    * `rm -rf ./my-app/rsconnect-python`
4. Migrate the deployment record, same as step 4 above
5. Verify as in step 5 above
    
## Checklist


docs/CHANGELOG.md
```diff
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  deploy from that directory updates that item instead of creating a second one.
+  Use it after migrating shinyapps.io content to Connect Cloud:
+  `rsconnect content migrate-to-connect-cloud ./my-app -n cloud
+  --content-id <id>`. Nothing is copied and no bundle is uploaded — the content
+  must already exist in Connect Cloud, and only local files change. The record
+  it was migrated from is removed, so the directory is left with one deployment
+  target rather than two; pass `--from-server` to choose which record to migrate
+  when there are several.
+
+## Unreleased
```
vrsarah: Looks like this heading is duplicated

rsconnect/api.py
```diff
@@ -2256,6 +2272,134 @@ def write_deployed_info(self):
             self.app_mode,
         )
 
+    def migration_source_record(self, from_server: Optional[str] = None) -> Optional[AppMetadata]:
+        """The deployment record being migrated away from, if there is one.
+
+        Only records for other servers are candidates: a Connect Cloud record is
+        what this migration produces, so treating one as a source would delete the
+        result. With no `from_server` a lone record is taken, and several are
+        reported rather than picked between.
```
vrsarah, line 2299: Is this safe when the lone record is Connect's deployment record?

samperman: i cleaned that up. now we preserve the Connect record _and_ create a Connect Cloud record

rsconnect/api.py
```diff
@@ -2236,6 +2257,158 @@ def write_deployed_info(self):
...
+        """
+        candidates = [
+            record
+            for record in self.app_store.get_all()
+            if not connect_cloud.is_connect_cloud_url(record.get("server_url", "").split("#")[0])
+        ]
+
+        if from_server:
+            target = resolve_server_alias(from_server)
+            matches = [record for record in candidates if record.get("server_url") == target]
```
karawoo: do we need to accommodate possible trailing slashes here?

rsconnect/api.py
```diff
@@ -2236,6 +2257,158 @@ def write_deployed_info(self):
...
+        reported rather than picked between.
+        """
+        candidates = [
+            record
+            for record in self.app_store.get_all()
+            if not connect_cloud.is_connect_cloud_url(record.get("server_url", "").split("#")[0])
+        ]
+
+        if from_server:
+            target = resolve_server_alias(from_server)
```
karawoo: IIUC this will not work if someone tries to migrate a record from a Connect instance by passing `--from-server <server name>`, you'd have to pass the URL. But for migrating from shinyapps.io you _can_ provide the server name. I know shinyapps.io is the main migration use case but it still would be nice to be consistent if it's not too hard.

karawoo (approved):
couple of small suggestions below but otherwise lgtm

vrsarah (approved):

rsconnect/api.py
samperman: good catch! i will fix this

rsconnect/api.py
samperman: good call. i'll make that change

PR #837, opened 2026-08-13
feat: add support for deploying to Posit Connect Cloud

samperman:
## Intent

Add Posit Connect Cloud as a deployment target, alongside Posit Connect and shinyapps.io, mirroring the R rsconnect package's support.

Fixes #817

## Type of Change

- [ ] Bug Fix
- [x] New Feature
- [ ] Breaking Change

## Approach

- **Target selection**: `--connect-cloud` (or `-s connect.posit.cloud`) with `-A <account>`, or a saved nickname via `-n`. The production, staging, and development environments are selected with `CONNECT_CLOUD_ENVIRONMENT` and pinned to the saved server's URL.
- **Authentication** (`oauth.py`, `connect_cloud.py`): interactive OAuth device-code login by default; service-account client ID/secret (client-credentials grant) via `--client-id`/`--client-secret` or `CONNECT_CLOUD_CLIENT_ID`/`CONNECT_CLOUD_CLIENT_SECRET` for CI. Tokens refresh automatically on 401 and the rotated refresh token is written back to the credential store.
- **Credential registration**: `rsconnect add` verifies the account exists and grants the `content:create` permission before storing.
- **Deployment** (`api.py`): follows the Connect Cloud revision model — create or update content, upload the bundle to a presigned URL, publish, poll the revision, and print the publish log from the logs service on failure. Deployments are recorded locally *before* publishing because Connect Cloud cannot look content up by name.
- **Credential redaction** (`http_support.py`): verbose logs and error messages now redact authorization/signing headers, OAuth tokens and client secrets in request/response bodies, presigned-URL query credentials, and `-E` environment-variable values.

Two behavior changes outside the Connect Cloud path (see CHANGELOG for details):

- `deploy manifest`/`deploy bundle` no longer overwrite an existing content title when `-t` is not given, matching the other deploy commands.
- Deploying with a nickname no longer conflicts with `SHINYAPPS_*` environment variables that happen to be set, and `deploy pyproject` no longer uses the nickname as a title override.

## Automated Tests

- `tests/test_connect_cloud.py` (new, ~2800 lines): device-code and client-credentials auth flows, token refresh and write-back, `rsconnect add` verification, the full deploy/publish/poll flow, error paths and publish-log output, and environment selection — all HTTP mocked with `httpretty`.
- `tests/test_http_support.py`: credential redaction coverage.
- Updates to `test_metadata.py`, `test_oauth.py`, `test_validation.py`, and others for the credential store and validation changes.

## Directions for Reviewers

[...]

PR #840, opened 2026-08-18
add support for deploying to Posit Connect Cloud

### LINKED ISSUE #817: Add support for publishing to Posit Connect Cloud

samperman:
## Intent

Add support for deploying to Posit Connect Cloud: OAuth device-code and service-account (client-credentials) authentication, credential management, content create/update/publish with log streaming, and deployment records — for the content types Connect Cloud supports.

Resolves #817. Supersedes #837 and #839, merging both into one reviewable unit; see #837 for the earlier review history and manual test-plan results (staging and production).

## Type of Change

- [ ] Bug Fix
- [x] New Feature
- [ ] Breaking Change

## Approach

Beyond the base feature, this incorporates the target-resolution and OAuth feedback from https://github.com/posit-dev/rsconnect-python/pull/837#issuecomment-5294462925, one commit per item:

- **Actionable refresh failures** — typed `invalid_grant`/`invalid_client` errors; an expired session clears its dead stored tokens and the error names the exact `rsconnect add` command to re-authenticate (preserving a non-production `-s` URL and the saved entry's account); transient failures still surface the original 401.
- **Nickname-based credential selection** — entries are credentials, accounts are publish targets: one saved credential is used regardless of `-A`, several require `-n`, and `-A` always selects the account to publish to (a Cloud login can publish to every account its user has rights on). Deployment records stay keyed by URL + account id. For non-Cloud targets, `-n` with `-A` still fails, but after nickname resolution and with a new message.
- **Keyring-backed credential storage** — Cloud tokens and stored client secrets live in the system keyring keyed `<url>#<nickname>`, with `servers.json` as the documented fallback when no usable keyring exists (CI, headless). The existing Connect keyring key format is frozen, with a test guarding it.
- **Shared 401 → refresh → retry machinery** — `BearerTokenHTTPServer` owns the retry-once skeleton (including request-body rewind) for both clients; token minting stays per-target (Connect keeps OAuth discovery + client re-registration, Cloud keeps client-credentials vs refresh-token and the typed errors above).

Two follow-ups from the same review land separately after this merges: redeploy target inference from a directory's deployment record (changes target-resolution precedence for all server types, so it warrants its own review) and the shinyapps.io deployment migration command (#838).

Companion: posit-dev/connect#42351 updates two nightly bats assertions that pinned the old `-n`/`-A` validation message. Each repo's CI installs the other's `main`, so the original changes were red in isolation; the assertions are now version-agnostic and that PR merges independently, in either order.

## Automated Tests

Full suite: 1081 passed / 12 skipped; CI green across Python 3.8–3.14 on ubuntu/macos/windows plus the Connect integration suites. Each behavior above has dedicated unit tests (refresh error paths, credential selection rules, keyring fallback/migration incl. the `NoKeyringError` CI case, retry-once with body rewind), and every commit was individually machine-reviewed.

## Directions for Reviewers

To validate manually:

```bash
uvx --from "git+https://github.com/posit-dev/rsconnect-python.git@connect-cloud-auth-refactor" rsconnect add --connect-cloud -A <account> -n cc
# device-code login; then from an app directory:
uvx --from "git+https://github.com/posit-dev/rsconnect-python.git@connect-cloud-auth-refactor" rsconnect deploy shiny .
```

Worth poking at: `-n cc -A <other-account>` (publish-target selection), a second `add` under another nickname then a bare deploy (must demand `-n`), `rsconnect list` (reports keyring vs file storage), and `rsconnect remove` (cleans up keyring entries). Set `PYTHON_KEYRING_BACKEND=keyring.backends.fail.Keyring` to exercise the `servers.json` fallback.

## Checklist

- [x] I have run the `rsconnect-python-tests-at-night` workflow in Connect against this feature branch. (Two bats message assertions failed as expected; addressed by posit-dev/connect#42351.)

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -19,17 +19,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `--connect-cloud` (or `-s connect.posit.cloud`) with `-A <account>` (or the
   `CONNECT_CLOUD_ACCOUNT` environment variable; a `SHINYAPPS_ACCOUNT` variable
   exported for shinyapps.io is ignored here), or `-n <nickname>` for a saved
-  account. Supported content types: Shiny (Python
-  and R), Streamlit, Dash, Bokeh, Jupyter notebooks, Quarto, R Markdown, and
-  static content.
+  credential — which publishes to the account it was saved with, or to another
+  account of the same login when `-A` is given as well. Supported content types:
+  Shiny (Python and R), Streamlit, Dash, Bokeh, Jupyter notebooks, Quarto,
+  R Markdown, and static content.
 - `rsconnect add` now reports invalid option combinations and unreadable
   certificate files as plain error messages. Previously these surfaced as raw
   Python tracebacks.
 - Deploying with a nickname (`-n`) no longer fails when `SHINYAPPS_ACCOUNT`,
   `SHINYAPPS_TOKEN`, or `SHINYAPPS_SECRET` happen to be set in the environment.
   Those values are ignored for nickname deploys — the saved credential is what
-  the nickname means. Shinyapps options typed on the command line still
-  conflict with `-n`.
+  the nickname means. A typed `-T/--token` or `-S/--secret` still conflicts with
+  `-n`.
 - Credentials are now redacted from verbose (`-v`) log output and from error
   messages that quote a request URL. This covers authorization and signing
   headers, OAuth tokens and client secrets in request and response bodies, the
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,43 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Posit Connect Cloud is now a supported deployment target, alongside Posit
+  Connect and shinyapps.io, mirroring the R rsconnect package's support.
+  Register an account with `rsconnect add -n <nickname> --connect-cloud
+  -A <account>`, which verifies the account exists and that you can publish to
+  it. Authentication is an interactive browser login by default; for CI and
+  other non-interactive use, pass a service account credential with
+  `--client-id`/`--client-secret` (or the `CONNECT_CLOUD_CLIENT_ID` and
+  `CONNECT_CLOUD_CLIENT_SECRET` environment variables). Tokens are refreshed
+  automatically. Deploy with any `rsconnect deploy` subcommand by passing
+  `--connect-cloud` (or `-s connect.posit.cloud`) with `-A <account>` (or the
+  `CONNECT_CLOUD_ACCOUNT` environment variable; a `SHINYAPPS_ACCOUNT` variable
+  exported for shinyapps.io is ignored here), or `-n <nickname>` for a saved
+  account. Supported content types: Shiny (Python
+  and R), Streamlit, Dash, Bokeh, Jupyter notebooks, Quarto, R Markdown, and
+  static content.
+- `rsconnect add` now reports invalid option combinations and unreadable
+  certificate files as plain error messages. Previously these surfaced as raw
+  Python tracebacks.
+- Deploying with a nickname (`-n`) no longer fails when `SHINYAPPS_ACCOUNT`,
+  `SHINYAPPS_TOKEN`, or `SHINYAPPS_SECRET` happen to be set in the environment.
+  Those values are ignored for nickname deploys — the saved credential is what
+  the nickname means. Shinyapps options typed on the command line still
+  conflict with `-n`.
+- Credentials are now redacted from verbose (`-v`) log output and from error
+  messages that quote a request URL. This covers authorization and signing
+  headers, OAuth tokens and client secrets in request and response bodies, the
+  query-string credentials of presigned bundle-upload URLs, and the values of
+  environment variables passed with `-E` (names are still logged).
+- `rsconnect deploy manifest` and `rsconnect deploy bundle` no longer overwrite
+  the title of existing content when `-t/--title` is not given. Previously a
+  redeploy reset the server-side title to one derived from the manifest or
+  bundle; now, as with the other deploy commands, a redeploy without `-t`
+  leaves the existing title unchanged.
+- `rsconnect deploy pyproject` no longer uses the server nickname (`-n`) as a
+  title override. The title now comes from `-t/--title` or the pyproject
+  metadata, so redeploying with a nickname no longer renames existing content
+  after the nickname.
 - `rsconnect deploy` subcommands now accept `--quiet`, which suppresses the
   step-by-step progress lines and the streamed server build log, printing only
   the deployed content URL to stdout so it can be captured with
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -8,21 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 - Posit Connect Cloud is now a supported deployment target, alongside Posit
-  Connect and shinyapps.io, mirroring the R rsconnect package's support.
-  Register an account with `rsconnect add -n <nickname> --connect-cloud
-  -A <account>`, which verifies the account exists and that you can publish to
-  it. Authentication is an interactive browser login by default; for CI and
-  other non-interactive use, pass a service account credential with
-  `--client-id`/`--client-secret` (or the `CONNECT_CLOUD_CLIENT_ID` and
-  `CONNECT_CLOUD_CLIENT_SECRET` environment variables). Tokens are refreshed
-  automatically. Deploy with any `rsconnect deploy` subcommand by passing
-  `--connect-cloud` (or `-s connect.posit.cloud`) with `-A <account>` (or the
-  `CONNECT_CLOUD_ACCOUNT` environment variable; a `SHINYAPPS_ACCOUNT` variable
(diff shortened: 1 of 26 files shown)


rsconnect/api.py
```diff
@@ -2450,6 +3354,252 @@ def do_deploy(self, bundle_id: str, app_id: str):
...
+        content_type = AppModes.get_connect_cloud_content_type(app_mode)
+        if content_type is None:
+            raise RSConnectException(
+                "Posit Connect Cloud does not support %s content." % app_mode.desc(),
+            )
+
+        # None (no -E given) means "leave the server's secrets alone"; it reaches
+        # update_content as None and is omitted from the PATCH. A non-empty -E set
+        # replaces the whole collection, matching the R client.
+        secrets = [{"name": name, "value": value} for name, value in env_vars.items()] if env_vars else None
```
vrsarah, line 3445: If I'm reading this right, a redeploy with a secret (-E NEW_SECRET=xyz) will wipe any existing secrets deployed before the redeploy. Might be worth documenting that in the -E help text

docs/CHANGELOG.md
```diff
@@ -7,6 +7,44 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Posit Connect Cloud is now a supported deployment target, alongside Posit
+  Connect and shinyapps.io, mirroring the R rsconnect package's support.
```
karawoo:
```suggestion
- Posit Connect Cloud is now a supported deployment target, alongside Posit
  Connect and shinyapps.io.
```

docs/CHANGELOG.md
```diff
@@ -7,6 +7,44 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  `--client-id`/`--client-secret` (or the `CONNECT_CLOUD_CLIENT_ID` and
+  `CONNECT_CLOUD_CLIENT_SECRET` environment variables). Tokens are refreshed
+  automatically. Deploy with any `rsconnect deploy` subcommand by passing
+  `--connect-cloud` (or `-s connect.posit.cloud`) with `-A <account>` (or the
+  `CONNECT_CLOUD_ACCOUNT` environment variable; a `SHINYAPPS_ACCOUNT` variable
+  exported for shinyapps.io is ignored here), or `-n <nickname>` for a saved
+  credential — which publishes to the account it was saved with, or to another
+  account of the same login when `-A` is given as well. Supported content types:
+  Shiny (Python and R), Streamlit, Dash, Bokeh, Jupyter notebooks, Quarto,
+  R Markdown, and static content.
```
karawoo: I think some of this detail is better left to the documentation rather than the changelog

rsconnect/api.py
```diff
@@ -1525,14 +1743,44 @@ def validate_server(self):
...
 
         return self
 
+    def validate_connect_cloud_server(self):
+        if not isinstance(self.remote_server, ConnectCloudServer):
+            raise RSConnectException("remote_server must be a Posit Connect Cloud server.")
+        if not isinstance(self.client, ConnectCloudClient):
+            raise RSConnectException("client must be a ConnectCloudClient.")
+        if self.visibility is not None:
+            # Connect Cloud has no equivalent setting.
```
vrsarah: Connect Cloud has visibility concept (content.access). Are we intentionally not covering resource access here?

rsconnect/api.py
karawoo: I think this should include `response.status is None` like we do at 405

```suggestion
                if response.status is None or response.status < 200 or response.status > 299:
```

rsconnect/api.py
samperman: oops! good catch. i will add that in

rsconnect/api.py
```diff
@@ -2382,6 +2823,469 @@ def get_applications_like_name(self, name: str) -> list[str]:
...
+        store.set(
+            server.server_name,
+            entry_url,
+            connect_cloud_account_name=entry.get("connect_cloud_account_name") or server.account_name,
+            connect_cloud_account_id=entry.get("connect_cloud_account_id"),
+            connect_cloud_client_id=entry.get("connect_cloud_client_id"),
+            connect_cloud_client_secret=None if secret_in_keyring else file_client_secret,
+            connect_cloud_access_token=None if in_keyring else access_token,
+            connect_cloud_refresh_token=None if in_keyring else refresh_token,
+        )
```
karawoo, line 3060: This rewrites servers.json on every refresh even if the contents haven't changed (i.e. if the keyring is in use). What do you think about rewriting only if the file has actually changed?

rsconnect/api.py
```diff
@@ -2382,6 +2823,469 @@ def get_applications_like_name(self, name: str) -> list[str]:
...
+        new_bundle: bool = True,
+        title: Optional[str] = None,
+    ) -> ConnectCloudContent:
+        """Update content, optionally minting a fresh revision to upload into.
+
+        `content_type` and `primary_file` are always sent alongside `app_mode`:
+        the API only recomputes `app_mode` when one of them is present in the
+        override set, and the stored content type would otherwise survive a
+        redeploy that changes what kind of content this is (--app-id pointing at
+        content of another type).
```
karawoo: I'm not sure I follow this

rsconnect/certificates.py
```diff
@@ -20,20 +22,33 @@ def read_certificate_file(location: str) -> str | bytes:
...
-    if suffix in TEXT_ENCODED_FILETYPES:
-        with open(path, "r") as tFile:
-            return tFile.read()
+    # Readability is judged before the suffix so a bad path reports as the real
+    # problem: the file type of a file that does not exist is beside the point.
+    # Both are operational errors: the CLI no longer checks existence at parse
+    # time (the certificate only applies once the target is known), so this is
+    # where a bad path surfaces. is_file() sits inside the handler because it
+    # raises OSError itself when the path's metadata cannot be read, e.g.
+    # through a permission-denied directory.
```
karawoo: I had a pretty hard time following this comment, I think the code itself is clearer

rsconnect/api.py
```diff
@@ -1525,14 +1743,44 @@ def validate_server(self):
...
 
+    def validate_connect_cloud_server(self):
+        if not isinstance(self.remote_server, ConnectCloudServer):
+            raise RSConnectException("remote_server must be a Posit Connect Cloud server.")
+        if not isinstance(self.client, ConnectCloudClient):
+            raise RSConnectException("client must be a ConnectCloudClient.")
+        if self.visibility is not None:
+            # Connect Cloud has no equivalent setting.
+            raise RSConnectException(
+                "-V/--visibility is not supported by Posit Connect Cloud. "
```
karawoo: Are there other flags that are supported for connect but should be disabled for connect cloud? thinking about `--image`, `--disable-env-management`/--disable-env-management-py`/`--disable-env-management-r`/`--disable-env-management-node`, `--node`, --hide-all-input` / `--hide-tagged-input` -- I'm not sure which ones connect cloud supports.

rsconnect/api.py
karawoo, line 2270: I think we need to adjust this, "Deploying as a draft requires Posit Connect 2025.06.0 or later" doesn't make sense if you're trying to deploy to Connect Cloud

karawoo (commented):
I tested this locally and was able to deploy to connect cloud successfully. Added a few comments below but they're pretty minor. For next time, if we could get independent changes in separate PRs that would help a lot with reviewing. I think the credential redaction and some of the refactoring could have been their own PRs and made the main feature easier to review.

rsconnect/api.py
samperman: It already is? too subtle?

> A non-empty -E set replaces the whole collection

rsconnect/api.py
samperman: good idea

tests/test_connect_cloud.py
```diff
@@ -0,0 +1,3556 @@
...
+    dropped."""
+
+    OPTIONS = {
+        "image": "-I/--image",
+        "disable_env_management": "--disable-env-management",
+        "env_management_py": "--disable-env-management-py",
+        "env_management_r": "--disable-env-management-r",
+        "draft": "--draft",
+        "metadata": "--metadata",
+    }
```
karawoo: I think the node options should be here too. these tests are probably a bit brittle since they rely on these hard coded options which can get out of sync with the real ones.

karawoo (approved):

tests/test_connect_cloud.py
samperman: ooh. good catch. i'll clean that up

rsconnect/connect_cloud.py
```diff
@@ -0,0 +1,315 @@
...
+        """
+        return "%s/%s/content/%s" % (self.ui, account_name, content_id)
+
+
+_ENVIRONMENTS: dict[str, ConnectCloudUrls] = {
+    "production": ConnectCloudUrls(
+        api="https://api.connect.posit.cloud/v1",
+        ui="https://connect.posit.cloud",
+        auth="https://login.posit.cloud",
+        logs="https://logs.connect.posit.cloud",
```
vrsarah: I think we are missing `/v1` for these logs URLs, [according to what the UI refers to](https://github.com/posit-hosted/vivid-ui/blob/3bd9caa1ac482d8c19dc468ea90f813e8f4afee9/src/config/prod.json#L17)
```suggestion
        logs="https://logs.connect.posit.cloud/v1",
```

vrsarah (approved):

rsconnect/connect_cloud.py
samperman: its added at the call site... but its inconsistent so i'll clean it up

PR #841, opened 2026-08-19
address PR review feedback on the Connect Cloud auth refactor

samperman:
Reject Connect-only deploy options for Connect Cloud.

Reject --draft so the deploy step no longer cites a Connect version for a target that has no draft step at any version.

Skip rewriting the store file when a save would not change it. DataStore._set always calls save(), so a Connect Cloud token refresh backed by the system keyring rewrote servers.json on every refresh with contents identical to what was already there,

Guard the response status against None before comparing it, since a failed connection produces a response carrying only the exception.

Trim the changelog entry, and rewrite two comments review found hard to follow.

## Intent
PR feedback, will be merged into the main PR when ready: #840

### DIFF
--- a/rsconnect/validation.py
+++ b/rsconnect/validation.py
@@ -60,6 +60,38 @@ def _get_present_options(
     return result
 
 
+# Deploy options that configure Posit Connect features Connect Cloud does not
+# have.
+_CONNECT_ONLY_DEPLOY_OPTIONS: dict[str, str] = {
+    "image": "-I/--image",
+    "disable_env_management": "--disable-env-management",
+    "env_management_py": "--disable-env-management-py",
+    "env_management_r": "--disable-env-management-r",
+    "draft": "--draft",
+    "metadata": "--metadata",
+}
+
+
+def _typed_connect_only_deploy_options(ctx: Optional[click.Context]) -> list[str]:
+    """The Connect-only deploy options this command line passed.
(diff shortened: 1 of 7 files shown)


rsconnect/validation.py
```diff
@@ -60,6 +60,38 @@ def _get_present_options(
...
+# Deploy options that configure Posit Connect features Connect Cloud does not
+# have.
+_CONNECT_ONLY_DEPLOY_OPTIONS: dict[str, str] = {
+    "image": "-I/--image",
+    "disable_env_management": "--disable-env-management",
+    "env_management_py": "--disable-env-management-py",
+    "env_management_r": "--disable-env-management-r",
+    "draft": "--draft",
+    "metadata": "--metadata",
+}
```
karawoo, line 74: Should `--disable-env-management-node` and `--node` be included here too? What about `--hide-all-input` / `--hide-tagged-input`?

karawoo (approved):
One question about the supported deploy options but otherwise lgtm

rsconnect/validation.py
samperman: good catch on the node flags. i've added those to the list. As for the two `hide` flags, i think i'm going to leave those as is for now and instead investigate how to properly support them on the Connect Cloud side. I've created an issue there to track it.

PR #849, opened 2026-09-08
fix: send the Python version to Connect Cloud

### LINKED ISSUE #848: Connect Cloud deploys always use default python version
## Summary

Deploys to Posit Connect Cloud always build with Connect Cloud's fallback Python (3.9), regardless of what the project or the manifest asks for. rsconnect-python detects the Python version requirement correctly and writes it into `manifest.json`, but never sends it to the Connect Cloud API, and Connect Cloud does not read it out of the bundle's manifest.

Reported on the forum: https://forum.posit.co/t/set-python-version-on-connect-cloud-when-deploying-using-rsconnect-via-command-line/217824

## Reproduction

With a `.python-version` containing `3.12` (or `pyproject.toml` with `requires-python = ">=3.12"`) and a dependency that needs >= 3.10:

```
rsconnect add --connect-cloud --account <account> --name cloud
rsconnect write-manifest shiny --overwrite .
rsconnect deploy manifest --name cloud manifest.json
```

The generated manifest is correct:

```json
"python": {"version": "3.12.4", ...},
"environment": {"python": {"requires": "~=3.12.0"}}
```

The deploy fails during dependency resolution:

```
Error resolving dependencies: Because the current Python version (3.9.17) does not
satisfy Python>=3.10 and plotnine==0.15.3 depends on Python>=3.10...
```

## Cause

The Connect Cloud API accepts a Python version on the revision:

- `POST /v1/contents` — `next_revision.python_version`
- `PATCH /v1/contents/{id}` — `revision_overrides.python_version`

It is a MAJOR.MINOR string from a fixed set (currently `3.8` through `3.14`).

`ConnectCloudClient.create_content` and `ConnectCloudClient.update_content` in `rsconnect/api.py` omit the field. With it unset, Connect Cloud falls back to a hardcoded default of 3.9 — it does not inspect `environment.python.requires` in the uploaded `manifest.json`. (The R `rsconnect` package omits the field too, so it has the same behavior.)

Note that Connect Cloud's documented default of 3.11 applies to the version dropdown in the GitHub deploy flow, not to bundle uploads.

## Workaround

Set the Python version on the content's settings page in Connect Cloud and redeploy.

samperman:
## Intent

Resolves #848.

Deploys to Posit Connect Cloud always built with Connect Cloud's fallback of Python 3.9, whatever the project asked for. rsconnect detects the Python version requirement correctly and writes it into `manifest.json`, but never sent it to the Connect Cloud API — and Connect Cloud does not read it out of the bundle. Content needing 3.10+ failed during dependency resolution.

Reported on the forum: https://forum.posit.co/t/set-python-version-on-connect-cloud-when-deploying-using-rsconnect-via-command-line/217824

## Type of Change

- [x] Bug Fix
- [ ] New Feature
- [ ] Breaking Change

## Approach

Connect Cloud takes the interpreter from the revision's `python_version` field — `next_revision.python_version` on `POST /v1/contents`, `revision_overrides.python_version` on `PATCH /v1/contents/{id}`. It is MAJOR.MINOR only, from a fixed set (`RequestPythonVersion` in https://api.connect.posit.cloud/openapi.json, currently 3.9 through 3.14). Posit Publisher already sends it; neither rsconnect client did.

`resolve_python_version()` derives the value from the manifest the bundle already carries — `environment.python.requires` and `python.version`. Reading the built bundle rather than the environment means `deploy manifest` is covered too, which does no environment inspection of its own, and no per-command plumbing was needed.

Which version wins:

- The interpreter the bundle was built against, when Connect Cloud offers that MAJOR.MINOR and the constraint allows it, so a deploy reproduces the development environment.
- Otherwise the lowest offered line satisfying the constraint, so `>=3.10` stays on 3.10 as Connect Cloud adds newer ones rather than drifting upward.
- A constraint no offered line satisfies fails before the bundle is uploaded, instead of surfacing as a server-side dependency resolution error.

Two trade-offs worth flagging:

**Patch-level constraints can only be honored as far as their minor line.** The API cannot express a patch, so `==3.11.14` deploys onto whatever 3.11.x Connect Cloud runs. Rejecting such constraints would refuse content Connect Cloud can run, and there is no endpoint exposing the patch behind each line. This matches Connect's own default of MAJOR.MINOR matching.

**Judging a constraint against a minor line needs care.** Testing `3.11` alone treats it as `3.11.0`, which would wrongly reject `>=3.11.3,<3.12`. `_line_satisfies()` instead tests the patches the constraint names, one above each, and the bottom of the line — every edge of the region a PEP 440 constraint admits falls on one of those. This avoids an arbitrary patch cap, so `==3.11.99` and `==3.11.100` behave identically.

Both payloads omit the field when there is no version to send, so R content deploys unchanged and a redeploy does not clear a version picked in the Connect Cloud UI. `python_version` sits at the end of the optional parameters on `create_content`, `update_content`, and `prepare_deploy` so it cannot capture a positional argument.

`r_version` and `quarto_version` are accepted on the same payloads and are still never sent — likely the same bug for R and Quarto content, left for a separate change since confirming it needs a real deploy.

## Automated Tests

25 new tests in `tests/test_connect_cloud.py`:

- `TestResolvePythonVersion` — constraint shapes (`~=`, `>=`, `==X.Y.*`, exact patch, `!=`), local-interpreter preference, versions outside the offered range, unparsable constraints, and the high-patch cases.
- `TestConnectCloudPythonVersionFromBundle` — reads the version out of real gzipped bundles, including a manifest with no Python section.
- Client payload tests — the field present and absent on both `create_content` and `update_content`.
- `prepare_deploy` passthrough on the create and update paths.

`_line_satisfies()` was additionally checked against exhaustive enumeration over patches 0–400, across 9,996 line/constraint pairs generated from 1,666 constraints, with no disagreements. That check is not committed; it is a one-off verification of the boundary-point argument.

Full suite: 1240 passed, 12 skipped. `ruff` clean. Pyright unchanged from baseline.

## Directions for Reviewers

The Python version rsconnect runs under is the one it inspects, so install the branch into the *project's* venv rather than running it from this repo:

```bash
cd ~/some-shiny-app          # with .python-version containing 3.12
VIRTUAL_ENV=$PWD/.venv uv pip install -e /path/to/rsconnect-python
.venv/bin/rsconnect deploy shiny -n <cloud-target> -vv .
```

`-vv` logs request bodies. Look for `"python_version": "3.12"` in `next_revision` on a first deploy, or `revision_overrides` on a redeploy, and confirm the build log shows 3.12 rather than 3.9.17.

Worth exercising:

- A redeploy, which takes the PATCH path with a different payload.
- `.python-version` set to `3.8`, which should fail before the upload with "Posit Connect Cloud does not offer".
- R content, where the field should be absent rather than `null`, so a version set in the UI survives.
- `deploy manifest`, which is the path from the forum report.

## Checklist

- [x] I have run the `rsconnect-python-tests-at-night` workflow in Connect against this feature branch.

### DIFF
--- a/rsconnect/api.py
+++ b/rsconnect/api.py
@@ -2194,8 +2194,9 @@ for shinyapps.io. See command help for further details."
     def python_version_for_connect_cloud(self) -> Optional[str]:
         """The MAJOR.MINOR Python version to ask Connect Cloud for, or None.
 
-        None means the request omits the field, leaving the version Connect Cloud
-        already has on the content. Content with no Python at all lands here too.
+        None means the request omits the field: a redeploy keeps the version already
+        set on the content, and a first deploy takes Connect Cloud's default. Content
+        with no Python at all lands here too.
         """
         manifest = self.bundle_manifest()
         environment: dict[str, Any] = manifest.get("environment") or {}
--- a/rsconnect/api.py
+++ b/rsconnect/api.py
@@ -2165,16 +2165,15 @@ for shinyapps.io. See command help for further details."
             )
             upload_result = S3Server(upload_url).handle_bad_response(upload_result, is_httpresponse=True)
 
-    def primary_file_for_connect_cloud(self) -> str:
-        """The entrypoint to report to Connect Cloud, read from the built bundle.
+    def bundle_manifest(self) -> dict[str, Any]:
+        """The manifest of the built bundle.
 
-        Connect Cloud needs a `primary_file` and uses it to decide, for example,
-        whether Shiny content is R or Python. Every deploy path writes the
-        entrypoint into the bundle's manifest, so reading it back from there
-        avoids having to thread it separately through each command.
+        Every deploy path writes what Connect Cloud needs into the bundle's
+        manifest, so reading it back from there avoids having to thread the same
+        values separately through each command.
         """
         if self.bundle is None:
-            raise RSConnectException("A bundle must be created before determining the primary file.")
+            raise RSConnectException("A bundle must be created before reading its manifest.")
 
         position = self.bundle.tell()
         try:
@@ -2186,12 +2185,33 @@ for shinyapps.io. See command help for further details."
                 extracted = tar.extractfile(member) if member is not None else None
                 if extracted is None:
                     raise RSConnectException("The bundle does not contain a manifest.json.")
-                manifest = json.loads(extracted.read().decode("utf-8"))
+                return cast("dict[str, Any]", json.loads(extracted.read().decode("utf-8")))
         except (tarfile.TarError, KeyError, ValueError) as exc:
             raise RSConnectException("Could not read the bundle manifest: %s" % exc) from exc
         finally:
             self.bundle.seek(position)
 
+    def python_version_for_connect_cloud(self) -> Optional[str]:
+        """The MAJOR.MINOR Python version to ask Connect Cloud for, or None.
+
+        None means the request omits the field, leaving the version Connect Cloud
+        already has on the content. Content with no Python at all lands here too.
+        """
+        manifest = self.bundle_manifest()
+        environment: dict[str, Any] = manifest.get("environment") or {}
+        requirement: dict[str, Any] = environment.get("python") or {}
+        interpreter: dict[str, Any] = manifest.get("python") or {}
+        requires: Optional[str] = requirement.get("requires")
+        local_version: Optional[str] = interpreter.get("version")
+        return connect_cloud.resolve_python_version(requires, local_version)
+
+    def primary_file_for_connect_cloud(self) -> str:
+        """The entrypoint to report to Connect Cloud, read from the built bundle.
+
+        Connect Cloud needs a `primary_file` and uses it to decide, for example,
+        whether Shiny content is R or Python.
+        """
+        manifest = self.bundle_manifest()
         metadata = manifest.get("metadata") or {}
         primary_file = metadata.get("entrypoint") or metadata.get("primary_rmd") or metadata.get("primary_html")
         files = manifest.get("files") or {}
@@ -2302,6 +2322,7 @@ for shinyapps.io. See command help for further details."
                     title=self.title,
                     app_mode=self.app_mode,
                     primary_file=self.primary_file_for_connect_cloud(),
+                    python_version=self.python_version_for_connect_cloud(),
                     env_vars=self.env_vars,
                     update_title=not self.title_is_default,
                     app_id_is_explicit=self.app_id_is_explicit,
@@ -3293,6 +3314,7 @@ class ConnectCloudAccountSearchResults(TypedDict):
 class ConnectCloudRevision(TypedDict):
     id: str
     content_id: NotRequired[str]
+    python_version: NotRequired[Optional[str]]
     status: NotRequired[str]
     source_bundle_upload_url: NotRequired[str]
     publish_result: NotRequired[Optional[str]]
@@ -3648,21 +3670,29 @@ class ConnectCloudClient(BearerTokenHTTPServer):
         primary_file: str,
         secrets: Optional[list[dict[str, str]]] = None,
         access: Optional[str] = None,
+        python_version: Optional[str] = None,
     ) -> ConnectCloudContent:
         """Create content to upload a bundle into.
 
         `access` is the content's visibility ("public" or "private"). Omitted from
         the request when None so the server picks its own default.
+
+        `python_version` is MAJOR.MINOR. Omitted when None, which is how content
+        with no Python is created; Connect Cloud then falls back to its own
+        default rather than reading the version out of the bundle's manifest.
         """
+        next_revision: dict[str, Any] = {
+            "source_type": "bundle",
+            "content_type": content_type,
+            "app_mode": app_mode,
+            "primary_file": primary_file,
+        }
+        if python_version is not None:
+            next_revision["python_version"] = python_version
         body: dict[str, Any] = {
             "account_id": account_id,
             "title": title,
-            "next_revision": {
-                "source_type": "bundle",
-                "content_type": content_type,
-                "app_mode": app_mode,
-                "primary_file": primary_file,
-            },
+            "next_revision": next_revision,
             "secrets": secrets or [],
         }
         if access is not None:
@@ -3680,6 +3710,7 @@ class ConnectCloudClient(BearerTokenHTTPServer):
         new_bundle: bool = True,
         title: Optional[str] = None,
         access: Optional[str] = None,
+        python_version: Optional[str] = None,
     ) -> ConnectCloudContent:
         """Update content, optionally minting a fresh revision to upload into.
 
@@ -3693,14 +3724,19 @@ class ConnectCloudClient(BearerTokenHTTPServer):
         deploy without -E/-t must leave the existing values alone. `access` (the
         content's visibility) is omitted the same way, so a redeploy without
         -V keeps whatever visibility the content already has.
+
+        `python_version` is omitted when None for the same reason: a redeploy of
+        content whose bundle carries no Python version keeps the version already
+        set on the content, including one chosen in the Connect Cloud UI.
         """
-        body: dict[str, Any] = {
-            "revision_overrides": {
-                "primary_file": primary_file,
-                "app_mode": app_mode,
-                "content_type": content_type,
-            },
+        revision_overrides: dict[str, Any] = {
+            "primary_file": primary_file,
+            "app_mode": app_mode,
+            "content_type": content_type,
         }
+        if python_version is not None:
+            revision_overrides["python_version"] = python_version
+        body: dict[str, Any] = {"revision_overrides": revision_overrides}
         if secrets is not None:
             body["secrets"] = secrets
         if title is not None:
@@ -3895,6 +3931,7 @@ class ConnectCloudService:
         update_title: bool = False,
         app_id_is_explicit: bool = False,
         visibility: Optional[str] = None,
+        python_version: Optional[str] = None,
     ) -> ConnectCloudDeployResult:
         """Create or fetch the content item and get a revision to upload into.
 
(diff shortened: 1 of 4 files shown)


rsconnect/connect_cloud.py
```diff
@@ -313,3 +317,104 @@ def credentials_from_keyring(url: str, nickname: str) -> tuple[Optional[str], Op
...
+        return None
+
+    for candidate in PYTHON_VERSIONS:
+        if _line_satisfies(candidate, specifier):
+            return candidate
+
+    raise RSConnectException(
+        "This content requires Python %s, which Posit Connect Cloud does not offer. "
+        "Supported versions are %s. Change the requirement in .python-version, "
+        "pyproject.toml, or setup.cfg." % (requires, ", ".join(PYTHON_VERSIONS))
```
vrsarah: or, the user might need to upgrade their rsconnect-python version (if we add a new python version to the PYTHON_VERSIONS list)

samperman: good point. i'm going to get rid of the list of versions

rsconnect/connect_cloud.py
```diff
@@ -313,3 +317,160 @@ def credentials_from_keyring(url: str, nickname: str) -> tuple[Optional[str], Op
...
+    version the constraint admits is used, which keeps ">=3.10" on 3.10 as Connect
+    Cloud adds newer ones.
+
+    Whether Connect Cloud offers the result is left to Connect Cloud, so a version
+    it adds after this release still deploys. The one version rule kept here is the
+    floor: below it the request would be rejected outright, and omitting the field
+    to take the platform default is more useful than failing.
+
+    Connect Cloud picks the patch itself, so a constraint that names one ("==3.11.14")
+    can only be honored as far as its minor line.
```
karawoo, line 476: Is there no way to respect the patch version? If not, can we include some sort of warning or info logging if the `.python-version` specifies a patch version so users know it won't be respected?

rsconnect/connect_cloud.py
```diff
@@ -313,3 +317,160 @@ def credentials_from_keyring(url: str, nickname: str) -> tuple[Optional[str], Op
...
+
+    The interpreter the content was built against wins when the constraint allows
+    it, so a deploy reproduces the development environment. Otherwise the lowest
+    version the constraint admits is used, which keeps ">=3.10" on 3.10 as Connect
+    Cloud adds newer ones.
+
+    Whether Connect Cloud offers the result is left to Connect Cloud, so a version
+    it adds after this release still deploys. The one version rule kept here is the
+    floor: below it the request would be rejected outright, and omitting the field
+    to take the platform default is more useful than failing.
```
karawoo: This seems like context for decisions within the PR more than external facing documentation

karawoo (commented):
I think there are some cases here where someone could get a Python version they've explicitly forbidden which will be confusing. If they've set `.python-version` to `>3.11.5`, we request 3.11, and if Connect Cloud only has 3.11.2 they'll be running on a version `<3.11.5`. Another scenario is if the version constraint isn't satisfiable by Connect Cloud, e.g. `>=3.8,<3.9`. 3.8 is the only version that satisfies this but it's lower than Connect Cloud's floor so it resolves to `None` and then Connect Cloud will pick a higher version that's expliclty not allowed according to the constraint.

rsconnect/connect_cloud.py
samperman: agreed... will clean this up

rsconnect/connect_cloud.py
samperman: no, Connect Cloud doesn't have all the patch versions available. we can provide the detailed patch but connect cloud will basically ignore it and choose the latest patch available. We could simplify this and just send the full patch version, then let connect cloud decide what to do. that might be less complicated

samperman: turns out this doesn't work (Connect Cloud doesn't accept it). i'll add some logging

vrsarah (approved):

rsconnect/connect_cloud.py
```diff
@@ -313,3 +317,194 @@ def credentials_from_keyring(url: str, nickname: str) -> tuple[Optional[str], Op
...
+    # every version it offers and resolves to 3.9. A requirement with no lower bound,
+    # or one naming only a major version, starts at Connect Cloud's lowest.
+    floor = _effective_floor(specifier)
+    start = max(floor, MINIMUM_PYTHON_VERSION) if floor is not None else MINIMUM_PYTHON_VERSION
+    minor = _lowest_admitted_line(start, specifier)
+    if minor is None:
+        logger.warning(
+            "No Python version Posit Connect Cloud offers satisfies the requirement %s. "
+            "It will choose a version, and the content will run on one the requirement "
+            "does not allow." % requires
```
karawoo:
```suggestion
            "rsconnect will not send a version, so the content will run on a Python "
            "version the requirement does not allow." % requires
```
As written I don't think this is accurate in the case of redeploys, where Connect Cloud will keep an existing version rather than choosing a version. Also the grammatical reference of "It" was ambiguous.

rsconnect/connect_cloud.py
karawoo: I see we're now logging `Requesting Python <major>.<minor>` but I don't think that is enough to convey that the patch version is being ignored.

rsconnect/connect_cloud.py
```diff
@@ -313,3 +317,194 @@ def credentials_from_keyring(url: str, nickname: str) -> tuple[Optional[str], Op
...
+
+
+def resolve_python_version(requires: Optional[str], local_version: Optional[str]) -> Optional[str]:
+    """Pick the Connect Cloud Python version for a deploy.
+
+    `requires` is the PEP 440 constraint from the manifest's
+    ``environment.python.requires``; `local_version` is the interpreter the
+    bundle was built against, from ``python.version``. Returns None when there is
+    nothing usable to send, which leaves the field off the request so Connect
+    Cloud keeps whatever the content already has.
```
karawoo: This assumes a redeploy, can we also note what happens on first deploy?

rsconnect/connect_cloud.py
```diff
@@ -313,3 +317,194 @@ def credentials_from_keyring(url: str, nickname: str) -> tuple[Optional[str], Op
...
+    if local is not None and (specifier is None or local in specifier):
+        minor = _minor_version(local)
+        if minor is not None and Version(minor) >= MINIMUM_PYTHON_VERSION:
+            return _log_requested(minor)
+        # rsconnect runs on Pythons older than Connect Cloud offers. Without a
+        # requirement there is nothing else to go on, so let the platform choose;
+        # with one, the search below may still find a version it does offer.
+        if specifier is None:
+            logger.warning(
+                "Posit Connect Cloud does not offer Python %s; it will choose a version for this content." % minor
```
karawoo: This warning message seems specific to first deploys; redeploys behave differently right (reuse the current python version for the content)?

rsconnect/connect_cloud.py
```diff
@@ -313,3 +317,194 @@ def credentials_from_keyring(url: str, nickname: str) -> tuple[Optional[str], Op
...
+    # or one naming only a major version, starts at Connect Cloud's lowest.
+    floor = _effective_floor(specifier)
+    start = max(floor, MINIMUM_PYTHON_VERSION) if floor is not None else MINIMUM_PYTHON_VERSION
+    minor = _lowest_admitted_line(start, specifier)
+    if minor is None:
+        logger.warning(
+            "No Python version Posit Connect Cloud offers satisfies the requirement %s. "
+            "It will choose a version, and the content will run on one the requirement "
+            "does not allow." % requires
+        )
```
karawoo: This also seems inaccurate in the case of redeploys.

rsconnect/api.py
```diff
@@ -2186,12 +2185,33 @@ def primary_file_for_connect_cloud(self) -> str:
...
         except (tarfile.TarError, KeyError, ValueError) as exc:
             raise RSConnectException("Could not read the bundle manifest: %s" % exc) from exc
         finally:
             self.bundle.seek(position)
 
+    def python_version_for_connect_cloud(self) -> Optional[str]:
+        """The MAJOR.MINOR Python version to ask Connect Cloud for, or None.
+
+        None means the request omits the field, leaving the version Connect Cloud
+        already has on the content. Content with no Python at all lands here too.
```
karawoo: What if this is a first deploy?

rsconnect/connect_cloud.py
```diff
@@ -313,3 +317,194 @@ def credentials_from_keyring(url: str, nickname: str) -> tuple[Optional[str], Op
...
+            logger.warning(
+                "Ignoring the manifest's Python version requirement, which is not a valid "
+                "PEP 440 constraint: %s" % requires
+            )
+
+    local = _parsed(local_version) if local_version else None
+    if local is not None and (specifier is None or local in specifier):
+        minor = _minor_version(local)
+        if minor is not None and Version(minor) >= MINIMUM_PYTHON_VERSION:
+            return _log_requested(minor)
```
karawoo, line 492: What happens if the local version we send is higher than the highest one Connect Cloud has?

samperman: the build will fail with an error message that the requested python version is unsupported. I thought this was better than maintaining a list of supported versions in the client and requiring a patch when new versions are added. Do you think we should maintain a list here?

karawoo: No I don't think we should maintain a list, I think it makes sense to fail if the requested version isn't supported, but it seems odd that the behavior is different depending on if the user's python version is too low or too high.

rsconnect/connect_cloud.py
samperman: i updated all of these... the log about patch version being ignored doesn't actually say what patch version is being ignored, since at the time that log happens, the info isn't readily available, so it require a little more surgery than i thought was needed. let me know if you think that is still not strong enough. (but if thats the only thing left, maybe we could do that in a future PR)

rsconnect/api.py
samperman: updated

rsconnect/connect_cloud.py
samperman: true... i considering dropping the min too 🤷 there is no chance we will ever support python 3.8 or 3.7 (or anything lower) so its possible to short circuit in that case

karawoo (approved):

